PR #2, opened 2010-10-06
rebased fixes for 1.4.x codeline

lindner:
Hi, Finally got around to rebasing my 1.4.x branch

Note that this compiles under gcc 4.4 without problems and without the patch that was submitted for that fix.

Would appreciate it if you could take a second look at these patches.


dormando:
I pulled the gitignore change.. thought I had already. In the middle of refactoring a bunch of stuff which we can then re-profile and add likely/unlikely behind new macros.

Updates to the connection code will happen soon after some code cleanup... will give a chance to reorder the structures, but more importantly idle connections won't hold onto buffers anymore. saves a lot more memory.

closing this out.

PR #30, opened 2012-12-03
Minor build changes

jbergstroem:
...namely

- fix deprecation for am_init_automake
- call automake with foreign so it doesn't generate unwanted files 
- use canonical_host over system/target which should allow for better cross-compilation


jbergstroem:
Closing based on age. Let me know if you want me to resubmit.

dormando:
actually I would appreciate a re-submission based on the 'next' branch? (or master if 1.4.25 is out by the time you do this) I did fix one thing in an unrelated change in that area. this was closed and I missed it initially. I won't ignore it for 3 years this time, I promise.

PR #36, opened 2013-01-24
Record and report on time spent in listen_disabled

zwischenzugs:
Hi. 

We had some issues on production where we were in a listen_disabled state and it would have been operationally useful to know how long the server was in this state across the various events.

I've written a patch which reports on the time the server is in the listen_disabled state in microseconds.

Thanks for your time!

zwischenzugs


ianmiell:
Is this worth pursuing?

dormando:
rebased, fixed up slightly, and pushed to 'next' tree.

PR #41, opened 2013-06-16
Mark automake-1.13 as supported

yuvipanda:
Minor fix to make `autogen.sh` work with automake-1.13 (tested on OS X Lion)


dormando:
fixed in the 'next' branch, properly enough for now. will have to revisit if bug reports of incompatible versions come in, since I know of none.

PR #44, opened 2013-08-01
reserve binary statuses and commands for locking.

ymmt2005:
I will post a mail to memcached ML to describe details.

This patch reserves binary protocol statuses and commands used for server-side locking.

Server-side locking is implemented in [yrmcds](http://cybozu.github.io/yrmcds/), a memcached-compatible KVS.
The location of the locking specification is here:
https://github.com/cybozu/yrmcds/blob/master/docs/locking.md

### DIFF
--- a/protocol_binary.h
+++ b/protocol_binary.h
@@ -67,6 +67,8 @@ extern "C"
         PROTOCOL_BINARY_RESPONSE_EINVAL = 0x04,
         PROTOCOL_BINARY_RESPONSE_NOT_STORED = 0x05,
         PROTOCOL_BINARY_RESPONSE_DELTA_BADVAL = 0x06,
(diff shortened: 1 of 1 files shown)


dormando:
need to review this to ensure it's still useful to do...

ymmt2005:
@dormando 
We have been using yrmcds with this server-side locking for years.

Specifically, the server-side locking is used to implement locking PHP sessions
in our products.  The lock is very robust even when PHP process suddenly dies
as yrmcds automatically unlock the session when it detects the locked client
disconnects / dies.

dormando:
thanks for the prompt reply to this ancient PR. With the number of open PR's and issues waning hopefully I can review this for inclusion or not.

dormando:
closing this out as we've deprecated the binary protocol. this could be added to the meta protocol as flags.

PR #54, opened 2014-01-15
Man page and usage text improvements

mlichvar:
These commits update the memcached man page, fix its usage message and add a new man page for memcached-tool.


dormando:
pushed to the 'next' branch, thanks.

PR #62, opened 2014-03-07
hacked in some basic idle timeout functionality

jsipprell:
**Idle timeout support:**

The following PR provides basic idle timeout support for the 1.4 tree. I realize this isn't the "active development" branch, but it is the one I was required to work with so this is what I'm sending back upstream.

Timeouts are configured globally, in seconds, via the `-T` command line argument. Default timeout is 0 -- *no timeout*. Each connection's timer is, of course, distinct. Each connection's idle timer is reset to the global initial value during *any* I/O and cleared when a connection is finalized. Idle timeouts are not meaningful with regard to UDP sockets and thus are not applied to them. Timeouts are handled synchronously, via libevent, and will perform a clean disconnect if/when they fire on a valid active connection.

Idle timeout support is enabled via ``./configure --enable-idle-timouts``, but one must still supply `-T` on the command line to activate at run-time.

A new stat is provided to track idle disconnections -``idle_disconnected_connections``. This counter can be viewed just like the rest of the general statistics.

Use of this feature is not recommended in extremely high connection count environments as the overhead could begin to take its toll when there are many thousands of concurrent connections.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -553,6 +553,13 @@ AM_CONDITIONAL([BUILD_CACHE], [test "x$build_cache" = "xyes"])
 AC_ARG_ENABLE(docs,
   [AS_HELP_STRING([--disable-docs],[Disable documentation generation])])
 
+AC_ARG_ENABLE(idle-timeouts,
+  [AS_HELP_STRING([--enable-idle-timeouts],[Enable per-connection idle timeouts (not for high-volume use!)])])
+
+AS_IF([test "x$enable_idle_timeouts" = "xyes"],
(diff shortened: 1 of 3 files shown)


dormando:
sad we never finished it. maybe I can drag it over the line... needs tests and validation and etc.

dormando:
going to close this out as we're going with the other version. This was highly educational as for what the needs are though. Thanks for your time on this (a whopping two years ago...)

I've also done some benchmarking with adding libevent timers to an event loop and there was a nontrivial amount of loss somehow. :/

PR #63, opened 2014-04-08
Enabling several -f(actors) separated by comma, changes to slab size logic

slawomir-pryczek:
Purpose of the change: 
Optimize memory usage for small items while keeping slab numbers low

Changes:
- No longer slabs with different sizes and same perslab will be created, eg. before we could have:
[ 3 slabs, item sizes = 350kb / 400kb / 480kb ] , with slab size=1M, they all end up with perslab = 2
After the change just one slab will be created. This will allow more variance in upper slab size distribution, so we can optimize more evenly using even small slab number (200), because slabs won't be wasted on same perslab items.
- You can now have several -f (factors) that'll be used evenly to calculate slab sizes, there's maximum of 10 factors which is defined as: FACTOR_MAX_COUNT 10
- Some more logic that'll prevent having incorrect slab sizes instead of throwing an error, eg. if there's 8 byte alignment, having factor of 1 will increase next slab size by 8 bytes
- Updated statistics to display all factors used


dormando:
closing this out. never got cleaned up... would be nice to allow explicit definitions of slab classes though.

PR #64, opened 2014-04-16
LRU Crawler

dormando:
LRU crawler.

Background thread that (safely?) crawls stuff on demand.

Needs to be torture tested by mc-crusher before I merge. Just opening a pull-request against myself for shiggles.

EDIT: been torture testing it.. the impact (especially at the default 100 microsecond sleep) is surprisingly low. Isn't crashing while working through millions of items with 15 million gets/sec going on at the same time. Something might still come up during a better prolonged torture test but.. so far it seems okay.

### DIFF
--- a/items.c
+++ b/items.c
@@ -789,7 +789,8 @@ static void *item_crawler_thread(void *arg) {
             }
             pthread_mutex_lock(&cache_lock);
             search = crawler_crawl_q((item *)&crawlers[i]);
-            if (search == NULL) {
+            if (search == NULL ||
+                (crawlers[i].remaining && --crawlers[i].remaining < 1)) {
                 if (settings.verbose > 2)
                     fprintf(stderr, "Nothing left to crawl for %d\n", i);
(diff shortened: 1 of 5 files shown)


items.c
```diff
@@ -612,3 +636,306 @@ void do_item_flush_expired(void) {
...
+    }
+    pthread_mutex_lock(&cache_lock);
+
+    if (strcmp(slabs, "all") == 0) {
+        for (sid = 0; sid < LARGEST_ID; sid++) {
+            tocrawl[sid] = 1;
+        }
+    } else {
+        for (char *p = strtok_r(slabs, ",", &b);
+             p != NULL;
```
rmccullagh: what is the point of this expression?

dormando: that's the middle of the for loop definition. if p isn't NULL, keep iterating.

rmccullagh: wow. the strtok_r completely threw me off. I thought the for loop was terminated. :)

PR #70, opened 2014-06-03
Use unlink instead of `rm -f` in start-memcached script

CameronNemo:
This is necessary for an Upstart job, because the call out to rm -f causes Perl to fork off a process, and Upstart think that new process is the main one (not the exec'd memcached process later on).


dormando:
pushed to next, thanks.

PR #71, opened 2014-06-03
Add Upstart jobs

CameronNemo:
This depends on PR #70.

The jobs are a little wordier than I prefer, but that was necessary to get the multiple instances (and it is a lot better than the init script, IMO).


dormando:
pushed to next. thanks. (I haven't actually tested these: but they can't possibly regress :P)

PR #72, opened 2014-06-08
fixed libevent version check: add the missing 1.0.x version check

NsLib:
The origin code only check the 1.1.x & 1.2.x version, however libevent has 1.0.x release.


dormando:
indeed there were a few of these in 2004 and 2005. I really hope you didn't discover this from installing onto a production machine with a libevent that old. merged into 'next.

PR #77, opened 2014-07-14
add vim swap file to .gitignore and t/whitespace.t's exempt list

menghan:
to avoid failing tests


dormando:
pushed to 'next', thanks

PR #80, opened 2014-09-04
idle connection kicker using 'stats conns' and event notifier frameworks

elfchief:
Still needs some cleanup and a stats counter, but submitting now for review. Could you please let me know if I'm missing anything obvious and that I'm following the suggested path for making this happen?

Possible problems/issues/observations:

1. it wasn't real obvious how many connections to do per run through the thread, or how long to sleep. I ended up picking an arbitrary number (100) and sleeping for enough time that it takes roughly 1 second to cycle through all the possible connections. This may be a bit heavyweight (we could probably get away with once per 10 seconds or once per minute)
2. That number is just a #define right now. I don't really see much value in making it tunable at runtime, though.
3. There might be a more efficient way to doing this by figuring out which connection would time out next, and then scanning the entire connection list to find a new one when that connection expires/updates, but that requires doing a full connection scan and probably a mutex'd variable that's checked on every command, plus is way more complex, so probably isn't worth it.
4. Only works on idle (rather than stuck-in-transmission) connections. Not sure how hard it would be to make it kill 'active' connections too.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -78,6 +78,7 @@ static enum try_read_result try_read_network(conn *c);
 static enum try_read_result try_read_udp(conn *c);
 
 static void conn_set_state(conn *c, enum conn_states state);
+static int start_conn_timeout_thread();
 
 /* stats */
 static void stats_init(void);
@@ -248,6 +249,7 @@ static void settings_init(void) {
     settings.hot_lru_pct = 32;
     settings.warm_lru_pct = 32;
     settings.expirezero_does_not_evict = false;
+    settings.idle_timeout = 0; /* disabled */
     settings.hashpower_init = 0;
     settings.slab_reassign = false;
     settings.slab_automove = 0;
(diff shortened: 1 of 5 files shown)


dormando:
This is more or less the right approach. Feels weird putting it into memcached.c though - but I guess the rest of the conn structures are in there already.

There're some whitespace changes that make the code changes harder to see; please submit those as a separate PR. There's also a TODO or two in here? And the docs files need updating.

If you want to get fancy, you can adjust the scan time based on the given idle timeout and maybe a histogram. On a scan, populate some buckets for how many connections have been idle for so long. If the idle_timeout is 60, and 0 connections were idle when you scanned, there's no point to waking up until 60s later.

If 10,000 connections are open, idle_timeout is 60, and the last scan had 100 conns idle for at most 50s, you can wake up in 10s and scan again. Probably no point in optimizing it further, but you could in a few ways.

dormando:
Oh, and use the per-thread stats for the stats counter. Very few things should be under the global stats lock.

dormando:
Any interest in finishing this thing? :P I think this might be closer than the other one (#62). I'd like to get this into a release after 1.4.25. I'll end up cleaning it up myself but it'd be nice if someone else did too.

elfchief:
I will look into finishing this up in the next week or two!

dormando:
perfect! Remember to rebase to master/next instead of merging it in. Let me know if you need a more thorough code review to go on as well. My old notes are still accurate as well.

dormando:
:'(

elfchief:
Yes, I am a horrible person. :( Life has gotten in the way much more than I'd really like. I do still intend to finish this up, unless you get to it first.

(I could also use more time in the day. I really don't understand why nobody has invented a time machine yet.)

dormando:
:)

elfchief:
Okay, I updated this (finally).

Changes from last time:

- Stats updates are now thread-local
- Find the longest idle we're not in the process of booting, and sleep for long enough that we'll kick that one on-time when we wake next
- Protocol docs updated
- Added "stats settings" line for config setting

Still to do:

- Split that whitespace change (I am apparently saving that for last)

dormando:
:))) Noting this again just in case: https://github.com/memcached/memcached/pull/62 - anything from their approach useful/etc. Otherwise I'll close it out in favor of this one.

elfchief:
Okay, there, split that whitespace change. ;)

I just flipped through #62, it's a valid approach, but I'm not sure it's a _better_ approach, partially because it seems harder to follow what's going on. I think this one (mine) will also make it easier to add a timeout for "data pending for this client but no progress made in a long time", which I want to do in a future PR.

(Which I'm open to opinions on how best to add. I think adding most of the logic would be straightforward, but I'm not sure how to handle shutting down the socket correctly.)

And to be fair, #62 is probably more CPU-efficient, though I'd not expect the difference to be particularly noticeable.

memcached.c
```diff
@@ -306,6 +308,89 @@ static int add_msghdr(conn *c)
...
+#define TIMEOUT_MSG_SIZE (1 + sizeof(int))
+static void *conn_timeout_thread(void *arg) {
+    int i;
+    conn *c;
+    char buf[TIMEOUT_MSG_SIZE];
+    rel_time_t oldest_last_cmd;
+    int sleep_time;
+    useconds_t timeslice = 1000000 / (max_fds / CONNS_PER_SLICE);
+
+
```
dormando: extraneous newlines

memcached.c
```diff
@@ -5213,6 +5324,7 @@ int main (int argc, char **argv) {
         HOT_LRU_PCT,
         WARM_LRU_PCT,
         NOEXP_NOEVICT,
+	IDLE_TIMEOUT,
```
dormando: tab, should be spaces

memcached.c
```diff
@@ -5607,7 +5723,7 @@ int main (int argc, char **argv) {
                 hash_type = MURMUR3_HASH;
                 start_lru_crawler = true;
                 start_lru_maintainer = true;
-                break;
+		break;
```
dormando: whitespace again

memcached.c
```diff
@@ -5788,6 +5904,10 @@ int main (int argc, char **argv) {
         exit(EXIT_FAILURE);
     }
 
+    if(settings.idle_timeout && start_conn_timeout_thread() == -1) {
```
dormando: needs space between if and (

elfchief:
Huh. I did a pass that caught most of that, must have gotten lost when I was rebasing stuff. Will fix.

memcached.c
```diff
@@ -471,6 +576,7 @@ conn *conn_new(const int sfd, enum conn_states init_state,
     c->msgcurr = 0;
     c->msgused = 0;
     c->authenticated = false;
+    c->last_cmd_time = current_time;
```
dormando: a comment would be nice here: initializing for the idle kicker.

memcached.c
elfchief: These are hand-crafted *artisan* newlines! :)

dormando:
left some pretty minor style comments. read over the other PR again, and I think this might end up being less resource intensive. Managing events in libevent adds more overhead than you'd expect: I've benchmarked just adding one firing once per second in the main loop and it dropped more than 5% of overall performance.

So I like this one better? it has docs and this and that.

elfchief:
Ok, should have all those fixed up.

dormando:
Thanks! can you fix the one minor issue on the other PR please?

dormando:
(looks like you did fix it here, you just need to push it to the other branch)

dormando:
merged into next, thanks!

could I interest you in any other small tasks? :)

elfchief:
You could! Whatcha got?

dormando:
shoot me an e-mail directly and we'll discuss :)

PR #84, opened 2014-10-28
Fix text protocol output for too big flags

rustamserg:
Flags is declared as unsigned 32 bits value however if higher bit is used then text protocol returns negative number.

### DIFF
--- a/items.c
+++ b/items.c
@@ -83,7 +83,7 @@ uint64_t get_cas_id(void) {
 static size_t item_make_header(const uint8_t nkey, const int flags, const int nbytes,
                      char *suffix, uint8_t *nsuffix) {
     /* suffix is defined at 40 chars elsewhere.. */
-    *nsuffix = (uint8_t) snprintf(suffix, 40, " %d %d\r\n", flags, nbytes - 2);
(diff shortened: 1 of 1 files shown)


dormando:
I have a more complete fix for this in a branch... but need to poke through clients to see if anyone's expecting them to be signed now :(

rustamserg:
Thank you. Would really appreciate to get it fixed in next releases. It will help us with deployments.

dormando:
I have a feeling clients have been expecting and parsing the signed value to make use of all of the flag bits... so it may not be possible to fix.

rustamserg:
We actually found that the flag is signed from parsing exception raised by  https://github.com/enyim/EnyimMemcached C# memcached client.  This library expects that flag conforms to memcached format description which clearly define it as unsigned value.
<flags> is an arbitrary 16-bit unsigned integer. https://github.com/memcached/memcached/blob/master/doc/protocol.txt

dormando:
https://github.com/memcached/memcached/commit/181ef834151a37587faf2ee59079af1b8cf610af fully treats the value like an unsigned int... which it more or less is unless someone sets that last bit.

this is now merged into 'next'. I went back and forth for a while on whether to gate on a feature flag, but am boldly deciding not to.

PR #85, opened 2014-11-17
gatkq: return key in response

saaros:
GATKQ was incorrectly mapped to GAT instead of GATK in binary protocol
handling and thus didn't return a key in the response.  Fixed that and added
test cases for GAT, GATQ, GATK and GATKQ in testapp.

Noticed this while testing a new memcahe client library, OMcache:
    https://github.com/saaros/omcache/


dormando:
thanks! guess the perl tests missed GATK as well. pushed to 'next'

PR #86, opened 2014-11-19
Fix typo in doc/protocol.txt

mckelvin:
`data blockm` should be a typo.


dormando:
being generous.. typo fix is in 'next' now.

PR #87, opened 2014-11-20
Fix issue #379

rmccullagh:


dormando:
Oh. I ended up fixing this in another way... also please don't include text changes in the same PR as code changes.

rmccullagh:
Sorry, quite new to pull requests, and didn't realize it would be combined into one. Anyway you can share how you fixed it? Just curious.

dormando:
it's in the 'next' branch... I was just checking for it->nbytes <= 2 (0 byte items have \r\n in them, end up being 2 bytes in length). So ITEM_data(it) with 0 byte data size should still return a pointer to the \r\n value?

PR #88, opened 2014-11-24
Handle SIGTERM the same as SIGINT

md5:


dormando:
It's not quite clear what people want, but it's safe enough to merge this. Tests passed and is now in the 'next' branch.

It's probably better to break the loop as eric's PR did, but I'm not sure why he closed it off so I won't merge it. There're also some bugs with stopping the various maintenance threads that I'm working on in another branch, so it may be smarter to fix those things first.

md5:
Thanks @dormando. For those looking at this PR, it looks like the change landed in this commit: https://github.com/memcached/memcached/commit/f87ac298bb36ba14abe1f4407c6fd04ca11579e8

dormando:
it went out with 1.4.22

PR #92, opened 2014-12-04
incr/decr: the value should be part of extras according to doc

saaros:
protocol-binary.xml states:
  Increment, Decrement
    MUST have extras.
    MUST have key.
    MUST NOT have value.

but memcached did not set extlen for incr and decr responses.  Most client
applications didn't appear to care, they just used the 8 bytes of the body
as the delta value and thus should not break when incr and decr start saying
the new value is in extras.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1135,7 +1135,8 @@ static void complete_incr_bin(conn *c) {
         if (cas) {
             c->cas = cas;
         }
-        write_bin_response(c, &rsp->message.body, 0, 0,
+        write_bin_response(c, &rsp->message.body,
+                           sizeof(rsp->message.body.value), 0,
                            sizeof(rsp->message.body.value));
         break;
     case NON_NUMERIC:
@@ -1161,7 +1162,9 @@ static void complete_incr_bin(conn *c) {
 
                 if (store_item(it, NREAD_ADD, c)) {
                     c->cas = ITEM_get_cas(it);
-                    write_bin_response(c, &rsp->message.body, 0, 0, sizeof(rsp->message.body.value));
+                    write_bin_response(c, &rsp->message.body,
+                                       sizeof(rsp->message.body.value), 0,
+                                       sizeof(rsp->message.body.value));
                 } else {
                     write_bin_error(c, PROTOCOL_BINARY_RESPONSE_NOT_STORED,
                                     NULL, 0);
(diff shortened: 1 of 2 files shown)


dormando:
What clients have been tested? Despite the docs being wrong, things will be implemented to the way they are :(

saaros:
libmemcached should handle this fine, it just looks at bodylen and doesn't care about extlen. I'm not sure which other clients implement the binary protocol.

dormando:
Would it be terrible if I asked if you could check a few other binary clients? There're a few java ones that do binary, the Dalli ruby client. Most others should be implemented by way of libmemcached.

saaros:
AFAICT this Java client won't break, it treats all data as 'payload': https://github.com/dustin/java-memcached-client/blob/master/src/main/java/net/spy/memcached/protocol/binary/MutatorOperationImpl.java#L69

But the Dalli ruby client will break: https://github.com/mperham/dalli/blob/master/lib/dalli/server.rb#L324 https://github.com/mperham/dalli/blob/master/lib/dalli/server.rb#L491

So maybe we just need to update protocol documentation and protocol_binary.h to match the implementation.   Or do something different: add a protocol version number to requests (using `reserved`?) and fix the implementation when using newer protocol version.

dormando:
The first byte is technically the protocol version... We'll probably just have to update the documentation. I've learned in the past that even trivial changes to an existing protocol version are impossible to propagate.

saaros:
Ok, let's just update documentation to match the implementation.

Regarding protocol versions: changing the first byte for requests doesn't really work if you need to support multiple protocol versions at the same time and don't want to require an extra round trip and reconnection to fall back to the old version.  If the new version was mostly compatible with the old one (like this chnage) and clients set their requested version in another field a server which supports that version could reply with a different magic number in the first byte to indicate a response in the new format. Older servers would just ignore it and respond in the old format.

I don't think this thing alone is worth a new protocol version, but it might be interesting to collect issues like this and publish a new version of the protocol that fixes them at some point in the future.

dormando:
I don't really like binprot in general. we designed it as a committee partly during a hackathon. I don't really want to invest more time in it right now until I figure out if I'll rev the whole thing and replace it or drop it for ASCII, since many of the forks/proxies have simply ignored support for it entirely.

saaros:
Looks like the bit I quoted in this PR refers to the request format and not to response format.  The sample response in https://github.com/memcached/memcached/blob/master/doc/protocol-binary.xml#L951 has the new value in message body.

PR #93, opened 2014-12-10
Release memory back to system

zwChan:
Add a feature that support release memory back to system. If the ratio of used items in a slab is less than release_mem_start(default is 50%), then start release memory; If the ratio  of used items greater than release_mem_stop(default is 80%), then stop release memory.
  It is not complicated because it base on the slab_reassign(auto move) function. The different is that the picked page is not assign to other slab, but freed. And it also depend on the lru_crawler to clear the expired items.
  The feature is useful when a cache cluster support  a lot of services(or application). We assign each service several cache instance in fixed size(say 4GB) randomly among cache servers(it's memory is 32GB or 64GB). Most of the time the service just used a few memory of their caches, and only reach the peak of memory usage at certain time. What is worse is that some services just don't know how much memory they need, so they required a lot of memory for 'safety'. With this feature, we can support more services using the current cache servers.

### DIFF
--- a/items.c
+++ b/items.c
@@ -899,7 +899,7 @@ enum crawler_result_type lru_crawler_crawl(char *slabs) {
     }
 
     for (sid = 0; sid < LARGEST_ID; sid++) {
-        if (tocrawl[sid] != 0 && tails[sid] != NULL) {
+        if (tocrawl[sid] != 0 && tails[sid] != NULL && !crawlers[sid].it_flags) {
             if (settings.verbose > 2)
                 fprintf(stderr, "Kicking LRU crawler off for slab %d\n", sid);
             crawlers[sid].nbytes = 0;
--- a/memcached.c
+++ b/memcached.c
@@ -180,6 +180,7 @@ static void stats_init(void) {
     stats.hash_power_level = stats.hash_bytes = stats.hash_is_expanding = 0;
     stats.expired_unfetched = stats.evicted_unfetched = 0;
     stats.slabs_moved = 0;
+    stats.slabs_freed = 0;
     stats.accepting_conns = true; /* assuming we start in this state. */
     stats.slab_reassign_running = false;
     stats.lru_crawler_running = false;
@@ -239,6 +240,10 @@ static void settings_init(void) {
     settings.shutdown_command = false;
     settings.tail_repair_time = TAIL_REPAIR_TIME_DEFAULT;
     settings.flush_enabled = true;
+    settings.release_mem_sleep = 30;
+    settings.release_mem_start = 50;
+    settings.release_mem_stop  = 80;
+    settings.lru_crawler_interval = 0;
 }
 
 /*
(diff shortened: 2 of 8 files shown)


zwChan:
I have tested at CentOS 6.3 and Ubuntu 14.04, using t/slab_free_memory_compare.t.
In slab_free_memory_compare.t, I send 100k keys with random expire and value length to 2 memcached, one is disable release memory function, named 'server1'; the other enable release function, named 'server2'.
### At CentOS, 
the result is something like: 
> total_mem1=26345472 B,total_mem2=25690112 B,diff=655 kB,evict1_incr=0,evict_diff=0,sys_mem1=28752 kB,sys_mem2=27652 kB,release sys_memory:1100 kB. 

It is interesting that the system release memory more than the memcached malloc diff(as it show in the stats). The CentOS seems has no reserved  padding at the top of the heap.

### At Ubuntu,
 the result is something like:
> total_mem1=9699328 B,total_mem2=8257536 B,diff=1441 kB,sys_mem1=11772 kB,sys_mem2=10352 kB,release sys_memory:1420 kB

The Ubuntu seems reserved some padding at the top of the heap, but average we can release 2/3 of the freed memory to the system.

### About eviction
 I also test about the effect on eviction. When the evicted keys just happen sometimes(e.g. a peak traffic to the cache), release memory function can reduce the number of evicted keys significantly. However, it has no improvement if the memory is very deficient.

In sum, I think this feature can work "good enough" even if the OS reserves padding at the top of the heap, and it can reduce the occupied memory, also reduce the number of evicted keys.

zwChan:
In addition, I also test the malloc_trim(0) at Ubuntu, the result as following:
> total_mem1=8912896 B,total_mem2=7340032 B,diff=1572 kB,sys_mem1=10992 kB,sys_mem2=9432 kB,release sys_memory:1560 kB
total_mem1=8912896 B,total_mem2=7471104 B,diff=1441 kB,sys_mem1=10992 kB,sys_mem2=9564 kB,release sys_memory:1428 kB
total_mem1=8912896 B,total_mem2=7602176 B,diff=1310 kB,sys_mem1=10992 kB,sys_mem2=9836 kB,release sys_memory:1156 kB
....
total_mem1=9568256 B,total_mem2=8126464 B,diff=1441 kB,sys_mem1=11780 kB,sys_mem2=10492 kB,release sys_memory:1288 kB
total_mem1=9699328 B,total_mem2=8257536 B,diff=1441 kB,sys_mem1=11780 kB,sys_mem2=10500 kB,release sys_memory:1280 kB
total_mem1=9699328 B,total_mem2=7995392 B,diff=1703 kB,sys_mem1=11780 kB,sys_mem2=10232 kB,release sys_memory:1548 kB

  It works sometimes, not always. Maybe it is cause by the 'hold' of the heap, not the padding at the top of the heap.

dormando:
could do this in a more simple manner based off of this patch series: https://github.com/memcached/memcached/pull/113

still not completely convinced of the value of it but the heuristics could be a lot simpler. in the new code: when page mover is "returning" a page to the global pool, call a function which frees it instead. This is much less dangerous to use now that the slab mover rescues valid items.

what I'd like to see is adding runtime tunability to the limit_maxbytes variable. if we're returning a page to the global pool and current bytes is higher than limit_maxbytes, free it. If we need to pull from main memory and limit_maxbytes is higher than it was before, allow more memory to be allocated.

Then if you're doing this kind of live tuning, you can use an external process to clamp the memory down and wait for it to be freed up (or run a flush_all and wait for the lru_crawler to blow up everything).

dormando:
Done here: https://github.com/memcached/memcached/pull/170 (so small!) closing this out.

PR #95, opened 2014-12-15
Add drop_privileges() for Linux

viraptor:
Implement an aggressive version of drop_privileges(). Additionally add
similar initialization function for threads drop_worker_privileges().

This version is similar to Solaris one and prohibits memcached from
making any not approved syscalls. Current list narrows down the allowed
calls to socket sends/recvs, accept, epoll handling, futex (and
dependencies - mmap), getrusage (for stats), and signal / exit
handling.

Any incorrect behaviour will result in EACCES returned. This should be
restricted further to KILL in the future (after more testing).

The feature is only tested for i386 and x86_64. It depends on bpf
filters and seccomp enabled in the kernel. It also requires libsecomp
for abstraction to seccomp filters. All are available since Linux 3.5.

Seccomp filtering can be enabled at compile time with --enable-seccomp.

In case of local customisations which require more rights, memcached
allows disabling drop_privileges() with "-o no_drop_privileges" at
startup.

Tests have to run with "-o relaxed_privileges", since they require
disk access after the tests complete. This adds a few allowed syscalls,
but does not disable the protection system completely.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -47,6 +47,33 @@ If you export the environment variable
 T_MEMD_USE_DAEMON="127.0.0.1:11211" the tests will use an existing
 daemon at that address.
 
+* Debugging seccomp issues
+
+If new functionality fails when built with seccomp / drop privileges
+support, it can be debugged in one of two ways:
+
+Run the memcached via strace. For example:
+
+    strace -o /tmp/memcache.strace -f -- ./memcached
+    less /tmp/memcache.strace
+
+And look for calls which failed due to access restriction. They will
+show up with result: "-1 (errno 13)". Then add them to linux_priv.c.
+
+Alternatively, change the definition in linux_priv.c to:
+
+    #define DENY_ACTION SCMP_ACT_TRAP
+
+and the process will crash with a coredump on all policy violations.
+In strace output those can be seen as:
+
+    SIGSYS {si_signo=SIGSYS, si_code=SYS_SECCOMP,
+    si_call_addr=0x358a443454d, si_syscall=__NR_write,
+    si_arch=AUDIT_ARCH_X86_64} ---
+
+In that output, the si_syscall shows which operation has been
+blocked. In this case that's `write()`.
+
 * Sending patches
 
 See current instructions at http://contributing.appspot.com/memcached
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,6 +33,11 @@ if BUILD_SOLARIS_PRIVS
 memcached_SOURCES += solaris_priv.c
 endif
 
+if BUILD_LINUX_PRIVS
+memcached_SOURCES += linux_priv.c
+LDFLAGS += -lseccomp
+endif
+
 if ENABLE_SASL
 memcached_SOURCES += sasl_defs.c
 endif
@@ -59,6 +64,8 @@ memcached_debug_LDADD += memcached_debug_dtrace.o
 memcached_debug_DEPENDENCIES += memcached_debug_dtrace.o
 CLEANFILES += memcached_dtrace.o memcached_debug_dtrace.o
 endif
+
+memcached_debug_CFLAGS += -DMEMCACHED_DEBUG
 
 memcached_dtrace.h: memcached_dtrace.d
 	${DTRACE} -h -s memcached_dtrace.d
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ list to ask questions, github issues aren't seen by everyone!
 ## Dependencies
 
 * libevent, http://www.monkey.org/~provos/libevent/ (libevent-dev)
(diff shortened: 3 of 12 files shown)


jakesyl:
+1 on merge

viraptor:
This pr needs a refresh to work with me changes, but I wanted to check if anyone is interested in it first.
If there's a chance it will be looked at, I'm happy to refresh it.

dormando:
I can look, but it's very difficult for me to test it. It'd help if we can find some other people to do so, or figure out how I can get one of my cloud buildbots to do it. Any thoughts there?

viraptor:
@dormando Are you thinking of automated negative tests? As in verification that the blocking actually happens? The only thing I can think of right now is adding a command that's only available in the test build and only exists to trigger things that shouldn't happen. For example `DO_BAD_THING 1` tries to `system("any_command")`, `DO_BAD_THING 2` tries to bind a new listening socket, etc.

As for positive testing, the whole test suite should work just fine with slightly extended rights. They do need extended privileges as mentioned in the commit - `-o relaxed_privileges` allows reporting all the test results properly, but doesn't actually kill the protection system.

I can't find any configuration for the buildbot in the repository. If there's anything I can help there, I'm happy to have a look.

Let me know if that's what you had in mind.

dormando:
Need at least positive tests. Maybe this isn't as hard as I was thinking - but verifying that the code is getting some exercize on my builders. The pattern for those kind of tests is looking for a magic environment variable that then tries to do something... but that might end up in someone's production, so the test command would have to be carefully chosen to prevent a security problem for people who don't have the privdrop enabled.

Buildbot stuff isn't in the repo sadly, it's tied to my cloud accounts. You can browse it a bit here: http://build.memcached.org:8010/ - the master is always live, and I fire up the builders when I'm working on a release.

There's also this that came in recently: https://github.com/memcached/memcached/pull/203 - not sure if there's any common ground or improvements to be had there.

viraptor:
@dormando I added a test that tries to run a command and open a socket in the worker thread - both operations are stopped by seccomp.

The command itself is only built in memcached-debug. There's a MEMCACHED_DEBUG define flag available for it now. It also enables more allowed calls automatically so the performance profile can be saved.

The branch is now up to date with the current master. This includes the lru worker and everything that runs during `make test`. I also added some details in HACKING about how to debug issues of seccomp breaking new code.

dormando:
(late, again :\)

Are you missing a file from the commit? It doesn't build due to missing -lseccomp from LIBS flags. I see it being added in Makefile.am under LDCONFIG but that's not propagating. How were you building this?

(I can fix it; but generally think it's better to ask when I see something odd in this way).

Also, if you have a minute; can you add the the extra outputs to doc/protocol.txt and rebase against `next`?

dormando:
one more poke. since it wasn't building I hadn't returned to this.

viraptor:
Missed your comment, thanks for the poke! I'll have a look at this in the next day or two.

dormando:
raaad, thanks!

viraptor:
I uploaded the code rebased on current `next` branch. (https://github.com/memcached/memcached/pull/275) It required one extra syscall and has the extra command documented.

I'm happy to debug the build issue, but I can't reproduce it :( On a clean checkout `./autogen.sh ; ./configure --enable-seccomp ; make ; make test` just works for me. What's the error you're seeing?

dormando:
Sorry, again :( I immediately got sick after you rebased things.

Here's the error during compilation: https://gist.github.com/dormando/8069d6bb920eea03dc0db199ea61c570

It configures fine. As I said before I don't see -lseccomp being added anywhere, so it can't link toward the end there.

dormando:
err nevermind, I'm blind.. I see the -lseccomp. not obvious why it's failing though.

viraptor:
Ohh... interesting. What system is that on? I've tried on Arch and Debian without issues.

dormando:
ubuntu 16.04

dormando:
Any ideas (or something I should look for?) I'm probably cutting a release soon.

jakesyl:
Huh, that's wierd.  Works fine on my Ubuntu 16.04 version.

viraptor:
Sorry, I've been away the last 2 weeks. Will check it tonight.

viraptor:
I did get the error by staring from a clean `ubuntu:16.04` docker image:
```
apt install build-essential libevent-dev libseccomp-dev
./autogen.sh
./configure --enable-seccomp
make
```
Getting same error about missing `seccomp_rule_add` in that case. I'll investigate some more soon.

viraptor:
Fixed in the next branch PR (https://github.com/memcached/memcached/pull/275)

dormando:
merged fia the other PR

PR #97, opened 2015-01-05
[Work In Progress] LRU rework

dormando:
This branch is a functional reworking of memcached's core LRU. Looking for some help with folks willing to run tests or try swapping a development or production machine to see if the hit ratio improves (or if it crashes).

Direct link to branch: https://github.com/dormando/memcached/tree/lru_rework - you can click "download zip" if you don't want a git clone. Build: `./autogen.sh && ./configure && make && make test` - need libevent-devel or equivalent installed.

 * cache_lock is gone, LRU's are now independently locked.
 * LRU's are now split between HOT, WARM, and COLD LRU's. New items enter the HOT LRU.
 * LRU updates only happen as items reach the bottom of an LRU. If active in HOT, stay in HOT, if active in WARM, stay in WARM. If active in COLD, move to WARM.
 * HOT/WARM each capped at 32% of memory available for that slab class. COLD is uncapped.
 * Items flow from HOT/WARM into COLD.
 * A background thread exists which shuffles items between/within the LRU's as capacities are reached.

The primary goal is to better protect active items from "scanning". items which are never hit again will flow from HOT, through COLD, and out the bottom. Items occasionally active (reaching COLD, but being hit before eviction), move to WARM. There they can stay relatively protected.

A secondary goal is to improve latency. The LRU locks are no longer used on item reads, only during sets and from the background thread. Also the background thread is likely to find expired items and release them back to the slab class asynchronously, which speeds up new allocations. Further work on the thread should improve this.

There are a number of new statistics to monitor this. Mainly you'll just want to judge your hit ratio before/after, as well as any potential latency issues.

This change is rough but functional. Notes:

To enable: start with `-o lru_maintainer,lru_crawler`

To adjust percentage of memory reserved for HOT or WARM LRU's (default to 32% each):
`-o lru_maintainer,lru_crawler,hot_lru_pct=32,warm_lru_pct=32`

A recommended start line:
`-o lru_maintainer,lru_crawler,hash_algorithm=murmur3`

An extra option: `-o expirezero_does_not_evict` (when used with lru_maintainer) will make items with an expiration time of 0 unevictable. Take caution as this will crowd out memory available for other evictable items.

Some caveats exist:

* Many possible tunables are currently hardcoded.
* Max number of slab classes is now 62, instead of 200. The default slab factor gives 42 classes.
* It's run for a few hours at 550k sets/sec + 1.2m key fetches/sec with the background thread and lru_crawler running. Your mileage may vary.

Main TODO items (in rough order):

- [x] Add automated LRU crawling from the background thread. Keep a histogram of expiration times to judge how often to send the crawler out. This will reclaim any expired items early.
- [x] Repair slab rebalancer lock safety (I have notes for how already)
- [x] Repair more tests
- [x] Add compatibilty mode (which will be the default). Locks will still be split, but background thread will not start and only one LRU will be used per slab class. Item LRU bumping will be inline, so the same as the current code.
- [x] Need tests specific to LRU maintainer thread.
- [x] Make HOT/WARM ratios at least starttime tunable. Runtime tunable is ideal.
 * Runtime tunable is a bit more work, delaying it for now. Starttime works.
- [x] Background thread gates aggressiveness by a low watermark
 * Left a very high default for now.
- [ ] Manage slab rebalancer from main background thread. If a source class has at least 2 pages worth of free chunks, it's possible for the slab mover to copy items instead of evicting them (I think). Combined with the LRU crawler keeping tabs on all actual free memory, this should allow fast automated movement of slab pages to optimal destinations.
 * Might do this in a separate branch. Algorithm is easy if free space exists, but tricky if all slabs are full and evicting, which is a common problem.
- [x] Documentation
 * Names for things may not be final.
- [x] Add a new starttime option to use a fourth LRU for items set to not expire. If expiration time == 0, these items will not be subject to eviction. Items with TTL's will continue to be subject to eviction. Use with extreme care.

This is loosely inspired by the 2Q algorithm. More specifically the OpenBSD variant of it: http://www.tedunangst.com/flak/post/2Q-buffer-cache-algorithm

It's then extended to cope with the fact that memcached items do not behave the same way as a buffer pool. TTL's mean extra scanning/shuffling is done to improve memory efficiency for valid items.

### DIFF
--- a/assoc.c
+++ b/assoc.c
@@ -27,6 +27,7 @@
 
 static pthread_cond_t maintenance_cond = PTHREAD_COND_INITIALIZER;
 static pthread_mutex_t maintenance_lock = PTHREAD_MUTEX_INITIALIZER;
+static pthread_mutex_t hash_items_counter_lock = PTHREAD_MUTEX_INITIALIZER;
 
 typedef  unsigned long  int  ub4;   /* unsigned 4-byte quantities */
 typedef  unsigned       char ub1;   /* unsigned 1-byte quantities */
@@ -167,10 +168,12 @@ int assoc_insert(item *it, const uint32_t hv) {
         primary_hashtable[hv & hashmask(hashpower)] = it;
     }
 
+    pthread_mutex_lock(&hash_items_counter_lock);
     hash_items++;
     if (! expanding && hash_items > (hashsize(hashpower) * 3) / 2) {
         assoc_start_expand();
     }
+    pthread_mutex_unlock(&hash_items_counter_lock);
 
     MEMCACHED_ASSOC_INSERT(ITEM_key(it), it->nkey, hash_items);
     return 1;
@@ -181,7 +184,9 @@ void assoc_delete(const char *key, const size_t nkey, const uint32_t hv) {
 
     if (*before) {
         item *nxt;
+        pthread_mutex_lock(&hash_items_counter_lock);
         hash_items--;
+        pthread_mutex_unlock(&hash_items_counter_lock);
         /* The DTrace probe cannot be triggered as the last instruction
          * due to possible tail-optimization by the compiler
          */
--- a/doc/new_lru.txt
+++ b/doc/new_lru.txt
@@ -1,0 +1,39 @@
+In versions new enough to have the `-o lru_maintainer` option, a new LRU
+mechanic is available.
+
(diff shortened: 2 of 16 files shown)


zzz6519003:
:beer

PR #99, opened 2015-01-14
Setting the pthread id of the LIBEVENT_THREAD on thread creation.

samanbarghi:
It seems that the thread_id variable in LIBEVENT_THREAD struct is not being set, not at the thread creation time nor later through calling pthread_self(). Although, the thread_id is not being used anywhere but for dtrace probe MEMCACHED_CONN_DISPATCH, it should be set properly.


dormando:
pushed to 'next', thanks.

PR #100, opened 2015-01-16
fix build with clang 3.5

swills:
Building with clang 3.5 produces 'error: equality comparison with extraneous parentheses', this patch removes the extra parens.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -280,7 +280,7 @@ static int add_msghdr(conn *c)
     c->msgbytes = 0;
     c->msgused++;
 
-    if (IS_UDP(c->transport)) {
+    if IS_UDP(c->transport) {
         /* Leave room for the UDP header, which we'll fill in later. */
         return add_iov(c, NULL, UDP_HEADER_SIZE);
     }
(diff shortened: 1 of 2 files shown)


emcconville:
Wouldn't be better to add `-Wno-parentheses-equality` as part of the `DETECT_CLANG` autoconf? I'm not familiar with this warning, but I'm guessing changing the .c files would emit other warning for other folks.

swills:
I considered disabling the warnings, but I think this change actually makes the code a little bit easier to read. I'm generally not in the "just silence the warnings" camp (the clang people probably know what they're doing by enabling the warning by default). I don't think this should cause warnings or issues with other compilers, but was hoping to see Travis confirm that for me. If there are issues with other builds, maybe that would be the way to go.

emcconville:
Agreed, I'm not happy about silencing warnings. :) I'm also on clang 3.5, and I'm not seeing this warning. What's the diff?

swills:
Not sure why I'm seeing them and you're not, perhaps it's a local config issue. I can only point to this log:

http://poudriere.mouf.net/poudriere/data/11amd64-default/2015-01-16_15h43m04s/logs/errors/memcached-1.4.21.log

emcconville:
Looping back in this. Thanks for the log, it was very helpful. I was able to reproduce this, and it seems to be a issue within ccache's implementation.

    CC="/usr/local/bin/ccache /usr/bin/clang" ./configure

[my log on pastebin](http://pastebin.com/raw.php?i=8tkTdSPh)

This can be resolved with `run_second_cpp=true` option in ccache. It can be set in environment, or by autoconf m4 check.

    run_second_cpp=true CC="/usr/local/bin/ccache /usr/bin/clang" ./configure

`make` and `make clean` works as expected

swills:
Ah, you're right, good catch. I should have known, I've run into the same issue with macro expansion before. Reminding myself to always try testing with ccache disabled. Thanks and sorry for the noise. Closing!

PR #102, opened 2015-01-30
Reset tap slot after disconnect

bossmc:
This change allows a client to perform more than 10 taps over the lifetime of the memcached server.  There's still a limit of 10 concurrent taps, but previously we wouldn't forget a tap once it had finished (the TCP connection drops).

I initially thought this code was in place to allow recovery of taps, but the previous line destroys the engine's record of the tap so there's no chance of a recovery (and the pointer is left dangling).


dormando:
I'm officially giving up on the engine-pu branch :( Thanks and sorry.

PR #105, opened 2015-02-16
copy original arg to avoid changing it with getsubopt()

tony2001:
getsubopt() changes the original string, this is quite confusing.
Here is how I start it:
./memcached -o slab_reassign,slab_automove,lru_crawler_sleep=1000

And here is how it looks like in `ps` output:
./memcached -o slab_reassign slab_automove lru_crawler_sleep=1000

The patch makes sure the commas are left in place and not replaced by \0.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -5047,7 +5047,7 @@ int main (int argc, char **argv) {
     enum hashfunc_type hash_type = JENKINS_HASH;
     uint32_t tocrawl;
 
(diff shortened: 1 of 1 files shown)


mkevac:
Any updates?

dormando:
I'm not going to get to this immediately, sorry.

dormando:
pushed to the 'next' branch. thanks!

PR #106, opened 2015-04-17
"add" and "replace" commands return wrong error codes at binary protocol.

ledyba:
At binary protocol,
memcached answers "Key exists"(code: 2) for "add" command and
"Key not found"(code: 1) for "replace" command if the conditions are not met.

But the document says: 
```
- "NOT_STORED\r\n" to indicate the data was not stored, but not
because of an error. This normally means that the
condition for an "add" or a "replace" command wasn't met.
```
https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L196

So, the memcached should answer "Item not stored"(code: 5) instead.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1190,7 +1190,6 @@ static void complete_incr_bin(conn *c) {
 }
 
 static void complete_update_bin(conn *c) {
-    protocol_binary_response_status eno = PROTOCOL_BINARY_RESPONSE_EINVAL;
     enum store_item_type ret = NOT_STORED;
     assert(c != NULL);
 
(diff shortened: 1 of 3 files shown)


dormando:
this feels weird. it seems like in the binary protocol this was fixed to give a more explicit error compared to the ascii one. "NOT_STORED" is generic... "key exists" for an Add or "Key not found" says what the actual error is.

There's a separate document covering the binary protocol: https://code.google.com/p/memcached/wiki/BinaryProtocolRevamped - though it doesn't cover which response codes go with which command unfortunately.

so... I think it's correct as is? I'll tentatively close this. thanks though.

PR #107, opened 2015-04-23
[doc]data return by "stats sizes" missing a STAT

Alwayswithme:


dormando:
pushed to next. thanks.

PR #109, opened 2015-07-21
fix the misuse of settings.hot_lru_pct in process_stat_settings

CatKang:
there should be warm_lru_pct instead of hot_lru_pct


dormando:
whoops! Thanks for the catch :) pushed to 'next'.

PR #110, opened 2015-07-28
release memory before exiting

Abioy:
'buf' could be freed after calling atoi, providing a gentler way to exit

Signed-off-by: Yongyue Sun <abioy.sun@gmail.com>


dormando:
the error was harmless, but this is technically correct. the best kind of correct. pushed to 'next', thanks.

PR #111, opened 2015-09-08
FreeBSD doesn't ship with automake-1.4 by default, so include automake 1.5.

erikarn:


dormando:
fixed in 'next' branch by dropping the explicit version requirement.

PR #113, opened 2015-09-29
Slab rebalancer and slab automover improvements

dormando:
**This PR is release-candidate ready**

Use start options like: `-o slab_reassign,slab_automove,lru_crawler,lru_maintainer`

The slab rebalancer and automover have sat as a first-pass and proof of concept for several years.

This changeset aims to improve both to the point where it can be enabled by default in future versions as well as generally be more useful to people. The old automover would only move one page every 10-60 seconds, and did so very conservatively. Without an automover, long running instances of memcached can utilize memory poorly if the average size of items changes after the memory is full.

1. Slab rebalancer is improved to attempt to "rescue" items while moving a page. If free memory is available in the source slab class, copy and re-link items which are still valid, instead of silently evicting them.
2. New automover default mode will aggressively return pages to a "global pool" if there is more than 2*pagesize free chunks available in a slab class. Pages will then distribute as-needed back into any slab class.
3. Returned an experimental automove=2 mode, which aggressively requests random memory be assigned to a slab class on any eviction.
4. One more forthcoming patch to optionally make automove decisions even if all memory is full (based on eviction pressure, most likely).

Some work remains for this branch:
- [ ] automove=3 mode + tunables for automoving while classes are full. [may not do this, see below]
- [x] Review for cleanups, extra counters, or verbose logging. [partly done]
- [x] Documentation of new tunables and variables.

## automove=3

Will likely punt on this for now. I cannot determine the _value_ of an object being evicted, so it is difficult to write an algorithm to generically move pages between slab classes when all memory is full and there are nothing but evictions. The right choice there is very dependent on the use case. I'll discuss some options below.

### pull pages in from other classes if free chunks are below a watermark (ie; half a page)

 * Still need to decide how to pull pages from other classes most effectively. Could do it randomly, or pull from the one with the most pages at the moment.

### Weighted shuffling of pages between slab classes to amortize evictions

 * If class 2 has 5 evictions per second, and class 1 has 2 evictions per second, slowly move pages from class 1 to 2.
 * If class 2 has 5 evictions per second and class 1 has 0 evictions per second, slowly move pages from class 1 to 2.
 * If class 2 has evictions where the item has _been fetched before_, and class 1 has either 0 evictions, or evictions where the items _have not been fetched before_, slowly move pages from 1 to 2.
  * Weakness: if pages are moving through class 1 so quickly they never get a chance to be fetched. Could slew against the number of overall sets or evictions.
 * If class 2 has evictions where the item has _been fetched before_, and class 1 has either 0 evictions, evictions where the item _has not been fetched before_, and/or the last accessed time on class 2 is significantly lower then class 1, slowly move pages from 1 to 2.
  * Similar problem as the option above.

If you need to have memory always evicting but still rebalance the slab pages, there is still the option of manually running the reassign command. centralized or per-host daemons can monitor the various `stats` commands once per N seconds and reassign pages in a way that fits the needs of the particular usage scenario.

The other changes in this branch related to rescuing items when possible should make it less traumatic for the hit ratio to arbitrarily move pages. What would improve this even more is a way to signal to the system to evict items from the tail before moving a slab page, if the slab class is full in the first place. Then items could always be rescued and force evictions at the tail rather than simply be evicted if no free chunks are available.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -598,8 +598,13 @@ integers separated by a colon (treat this as a floating point number).
 | slab_global_page_pool | 32u     | Slab pages returned to global pool for    |
 |                       |         | reassignment to other slab classes.       |
 | slab_reassign_rescues | 64u     | Items rescued from eviction in page move  |
-| slab_reassign_evictions                                                     |
-|                       | 64u     | Valid items evicted druing a page move    |
+| slab_reassign_evictions_nomem                                               |
+|                       | 64u     | Valid items evicted during a page move    |
+|                       |         | (due to no free memory in slab)           |
+| slab_reassign_evictions_samepage                                            |
+|                       | 64u     | Valid items evicted during a page move    |
+|                       |         | (due to free memory being in the same     |
+|                       |         |  page as the source item)                 |
 | slab_reassign_busy_items                                                    |
 |                       | 64u     | Items busy during page move, requiring a  |
 |                       |         | retry before page can be moved.           |
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -601,10 +601,10 @@ integers separated by a colon (treat this as a floating point number).
 | slab_reassign_evictions_nomem                                               |
 |                       | 64u     | Valid items evicted during a page move    |
 |                       |         | (due to no free memory in slab)           |
-| slab_reassign_evictions_samepage                                            |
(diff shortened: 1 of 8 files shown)


dormando:
This is now merged to master. Release will take a little while as I have to divorce from google code.

PR #117, opened 2015-11-16
Documentation: update manpage for -I (slabsize) parameter

mattiasgeniar:
Update documentation to more accurately reflect the value that should be passed along as the parameter via command line.

I found myself interpreting the current text as if I should write ```memcached -I 10mb``` but it should really be ```memcached -I 10m```. I'm hoping this PR makes that more clear.


dormando:
this might be one of the shortest PRs-to-merge in the history of the project, given how bad I am at this.

I adjusted the title of the commit to be an appropriate length but otherwise didn't touch. now in the 'next' branch, thanks.

PR #128, opened 2015-12-04
Dump command learned to handle no-expire keys

jasonlotito:
memcache internals can use the timestamp generated by time - uptime from
stats to signal a no-expire key.  The dump command didn't account for
this.  This meant that a dump could not be imported into another memcached
instance with a guarantee that the data would be the same With this change, 
the dump command now uses the stats command's output to generate the 
no-expire timestamp.  Any exported timestamp that matches this no-expire 
timestamp is replaced with a 0 instead, so that when reimporting the data, 
the key is properly set as a no-expire key.

### DIFF
--- a/scripts/memcached-tool
+++ b/scripts/memcached-tool
@@ -79,6 +79,25 @@ die "Couldn't connect to $addr\n" unless $sock;
 if ($mode eq 'dump') {
     my %items;
     my $totalitems;
+    my $uptime;
+    my $time;
+    my $noexpiretime;
+
+    print $sock "stats\r\n";
+
+    while(<$sock>) {
+        last if /^END/;
+
+        if (/^STAT uptime (\d*)/) {
+            $uptime = $1;
+        }
+
+        if (/^STAT time (\d*)/) {
+            $time = $1;
+        }
+    }
+
+    $noexpiretime = $time - $uptime;
 
     print $sock "stats items\r\n";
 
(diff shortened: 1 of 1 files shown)


key-amb:
The same problem happend to me.
Dumped data with older exptime are not restorable.

I think we don't need to check server's uptime.
Because items with older exptime which are retrieved by `get` should never expire. (Unless you `touch` with new exptime or invalidate by `flush_all`.)

Here's my patch:

```diff
diff --git a/scripts/memcached-tool b/scripts/memcached-tool
index 82c6c3c..ca584f2 100755
--- a/scripts/memcached-tool
+++ b/scripts/memcached-tool
@@ -93,6 +93,7 @@ if ($mode eq 'dump') {
     print STDERR "  Number of buckets: " . scalar(keys(%items)) . "\n";
     print STDERR "  Number of items  : $totalitems\n";
 
+    my $now = time();
     foreach my $bucket (sort(keys(%items))) {
         print STDERR "Dumping bucket $bucket - " . $items{$bucket} . " total items\n";
         print $sock "stats cachedump $bucket $items{$bucket}\r\n";
@@ -114,6 +115,7 @@ if ($mode eq 'dump') {
                 my $len = $3;
                 my $val;
                 read $sock, $val, $len;
+                my $exptime = ($keyexp{$k} < $now) ? 0 : $keyexp{$k};
                 print "add $k $flags $keyexp{$k} $len\r\n$val\r\n";
                 # get the END
                 $_ = <$sock>;
```

And in my recent project [memcached-cli](https://github.com/key-amb/perl5-App-Memcached-CLI), I fixed this problem in the same way.

dormando:
These patches are kind of unreliable. I can trivially make it output a 0 if exptime is 0... I think that's safe because otherwise the timestamp is added so it can't ever not be zero?

Just added that patch to the next branch: https://github.com/memcached/memcached/commit/1b13a586624e8e315f23a482691c9bfb3ab36d9b - feel free to test it?

Closing this out.

PR #129, opened 2015-12-07
Fail to build on GNU/Hurd

guikcd:
The first easy step was to define missing macros **IOV_MAX, PATH_MAX & MAXPATHLEN**.
Patch that dynamic alloc buffer instead of harcode PATH_MAX is used in the Debian package (http://anonscm.debian.org/cgit/collab-maint/memcached.git/tree/debian/patches/03_fix_ftbfs4hurd.patch, all credits to Svante Signell).
For MAXPATHLEN, i haven't touched the code, just define it if undefined.

The second step was about using UNIX domain sockets on GNU/Hurd.
The transport was hardcoded to tcp in *drive_machine()* when calling *dispatch_conn_new()* and *getpeername(2)* was always called even if transport is not tcp. After using *c->transport*, there is still a problem : *t/stats-conn.t* verify that the socket name appears in "stats conn" output, which is not the case at least on GNU/Hurd (and *BSD, according to docs) but only "<AF 1>". *conn_to_str()* calls *getsockname(2)* which always return an empty string for UNIX domain sockets.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -4119,7 +4119,7 @@ static void drive_machine(conn *c) {
                 STATS_UNLOCK();
             } else {
                 dispatch_conn_new(sfd, conn_new_cmd, EV_READ | EV_PERSIST,
-                                     DATA_BUFFER_SIZE, tcp_transport);
+                                     DATA_BUFFER_SIZE, c->transport);
             }
 
             stop = true;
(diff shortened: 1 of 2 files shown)


dormando:
Well I have no idea how to validate this, but the changes look correct and the tests pass. It's merged into the next branch and will go out in the next release.

thanks!

Issue #126, opened 2015-11-20
extra statistics


dormando:
seems like these would be nice:
https://code.google.com/p/memcached/issues/detail?id=290
https://code.google.com/p/memcached/issues/detail?id=289

chyun:
I think get_expired is meaningless when lru_maintainer is turned on, but if not, it's really useful.

dormando:
some super nice person implemented this, which is merged into the next branch. closing this out.

PR #130, opened 2015-12-09
Implement get_expired stats

sergiocarlosmorales:
From issue #126, https://code.google.com/p/memcached/issues/detail?id=289:
Implement get_expired

### DIFF
--- a/t/stats.t
+++ b/t/stats.t
@@ -206,9 +206,8 @@ is(0, $stats->{'reclaimed'});
 is(0, $stats->{'lrutail_reflocked'});
 
 # item expired
-print $sock "set should_expire 0 1 6\r\nfooval\r\n";
+print $sock "set should_expire 0 2678400 6\r\nfooval\r\n"; #2678400 = 31 days in seconds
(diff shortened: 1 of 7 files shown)


t/stats.t
```diff
@@ -203,6 +205,15 @@ is(0, $stats->{'evictions'});
 is(0, $stats->{'reclaimed'});
 is(0, $stats->{'lrutail_reflocked'});
 
+# item expired
+print $sock "set should_expire 0 1 6\r\nfooval\r\n";
+is(scalar <$sock>, "STORED\r\n", "set item to expire");
+sleep(2);
```
sergiocarlosmorales: if there is a better way to do this, let me know

dormando: you can set an expiration time of 31 days. Past 30 days it's interpreted as a unix timestamp, so any still relatively-small value will be immediately expired.

sergiocarlosmorales: thanks, forgot about that, fixed.

dormando:
thanks! So sorry for the massive delay. This has been pushed to the 'next' branch and will go out in the next release.

PR #135, opened 2015-12-30
Allow compilation with -Werror on gcc 4.4.7

calebshay:
gcc 4.4.7 with -Werror fails without this change:

```
memcached.c: In function ‘process_bin_get_or_touch’:
memcached.c:1286: error: ignoring return value of ‘fwrite’, declared with attribute warn_unused_result
```


dormando:
Wish folks would upgrade stuff, but.. this isn't harmful. Merged for next release, thanks!

PR #139, opened 2016-01-18
fix build with musl libc

### LINKED ISSUE #138: build failure with musl libc
Building with [musl libc](http://www.musl-libc.org/) fails with this error:
````console
gcc -DHAVE_CONFIG_H -I.  -DNDEBUG   -g -O2 -pthread -pthread -Wall -Werror -pedantic -Wmissing-prototypes -Wmissing-declarations -Wredundant-decls -MT memcached-slabs.o -MD -MP -MF .deps/memcached-slabs.Tpo -c -o memcached-slabs.o `test -f 'slabs.c' || echo './'`slabs.c
In file included from slabs.c:13:0:
/usr/include/sys/signal.h:1:2: error: #warning redirecting incorrect #include <sys/signal.h> to <signal.h> [-Werror=cpp]
 #warning redirecting incorrect #include <sys/signal.h> to <signal.h>
  ^
cc1: all warnings being treated as errors
````

The warning is correct according [posix](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html).

ncopa:
musl libc will warn if you include sys/signal.h instead of signal.h as
specified by posix. Build will fail due to -Werror explicitly beeing
set.

Fix it by use the posix location.

fixes #138

### DIFF
--- a/assoc.c
+++ b/assoc.c
@@ -14,8 +14,8 @@
 #include "memcached.h"
 #include <sys/stat.h>
 #include <sys/socket.h>
-#include <sys/signal.h>
 #include <sys/resource.h>
+#include <signal.h>
 #include <fcntl.h>
 #include <netinet/in.h>
 #include <errno.h>
--- a/items.c
+++ b/items.c
@@ -2,13 +2,13 @@
 #include "memcached.h"
 #include <sys/stat.h>
 #include <sys/socket.h>
-#include <sys/signal.h>
 #include <sys/resource.h>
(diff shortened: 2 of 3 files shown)


ncopa:
Another option migtht be to add a --disable-werror or similar configure option.

aleiphoenix:
:+1: to this.

dormando:
The joys of super old codebases. merged into the 'next' branch. thanks!

PR #154, opened 2016-04-05
Update items.c

fowles:
APPEND_NUM_FMT_STAT is a macro that expands to multiple statements, so it must have braces when used in a conditional.


dormando:
Thanks! I changed the commit message to be more descriptive. Pushed for the next release.

PR #155, opened 2016-04-06
Fix typo of pid path

hiracy:
I tried daemon stop with memcached.sysv script.
So process was stopped, but remained the PID file.
Check this commit.


dormando:
thanks, pushed for next release.

PR #159, opened 2016-05-01
Specify the default value of settings.sasl to be false

biwing:
Specify the default value of settings.sasl to be false


dormando:
this is a no-op, but the rest of the code is explicit so I'll allow it. Thanks!

PR #160, opened 2016-05-01
Remove meaningless if statemet because an assertion already exist

biwing:
Remove meaningless if statemet because an assertion already exist


dormando:
Going to have to NACK this, sorry. The if isn't useless because assert()'s aren't actually compiled by default. If you build the source tree they're only in the memcached-debug binary that's run with the tests.

I've been bitten by that a number of times though and would like to find an opportune time to enable assert's by default.

rmccullagh:
Ye.p asserts will not do anything unless you compile with `-g` I believe.

PR #162, opened 2016-05-30
correct indent

sidgwick:
Just correct some bad indent, nothing changed.


dormando:
Ah... I did that on purpose because the extra indentation made the lines too long.. I need to do the actual right thing and reorg that code into functions.

Thanks for the reminder though. I'll leave this open until it gets cleaned up either way.

sidgwick:
:ok_hand:

PR #169, opened 2016-06-22
Online stats sizes and specifying exact slab class sizes

dormando:
This two-parter PR is handy for folks wanting to tune their slab class sizes.

The `stats sizes` command was generally off limits as it used to hang the cache. More recently it just hangs that particular worker thread and hammers the locks.

With this update a passive system can be enabled or disabled live which will track a histogram of item sizes as they're added or removed from the cache. This allows temporarily enabling it, or simply enable and leave it running if you don't mind the overhead. The command then dumps just the pre-created histogram on demand.

`stats sizes_enable` and `stats sizes_disable` enable and disable the feature.

If your `-f` and `-n` options can be tuned to give less memory overhead, that's great. However, what if you have a cluster where items only exist at exact sizes, or the factorial is never quite close enough?

You can now specify the slab classes manually via `-o slab_sizes=100-200-300-400-500`. Which will create 5 slab classes around those sizes, plus a sixth at the max item size.

TODO:

- [x] Update documentation and -h.
- [x] Decide on requiring CAS being enabled for stats sizes to work.
- [x] Start option for enabling stats_sizes

Optional:

- [ ] Automated tests (a nice to have since they weren't here before either)

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -720,6 +720,8 @@ other stats command.
 |                   | bool     | If yes, items with 0 exptime cannot evict    |
 | idle_time         | 0        | Drop connections that are idle this many     |
 |                   |          | seconds (0 disables)                         |
+| track_sizes       | bool     | If yes, a "stats sizes" histogram is being   |
+|                   |          | dymamically tracked.                         |
 |-------------------+----------+----------------------------------------------|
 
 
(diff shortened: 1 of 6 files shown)


dormando:
This is now complete so far as I'm concerned. I won't merge it for a day or two while I work on other things.

PR #170, opened 2016-06-23
cache_memlimit command for tuning runtime maxbytes

dormando:
Allows dynamically increasing the memory limit of a running system, if memory
isn't being preallocated.

`cache_memlimit N` where N is a value in megabytes. Value can go up or down.

If `-o modern` is in use, can also dynamically lower memory usage. pages are
free()'ed back to the OS via the slab rebalancer as memory is freed up. Does
not guarantee the OS will actually give the memory back for other applications
to use, that depends on how the OS handles memory.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3601,6 +3601,31 @@ static void process_watch_command(conn *c, token_t *tokens, const size_t ntokens
     }
 }
 
+static void process_memlimit_command(conn *c, token_t *tokens, const size_t ntokens) {
+    uint32_t memlimit;
+    assert(c != NULL);
+
+    set_noreply_maybe(c, tokens, ntokens);
+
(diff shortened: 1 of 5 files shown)


memcached.c
```diff
@@ -3601,6 +3601,31 @@ static void process_watch_command(conn *c, token_t *tokens, const size_t ntokens
     }
 }
 
+static void process_memlimit_command(conn *c, token_t *tokens, const size_t ntokens) {
+    uint32_t memlimit;
+    assert(c != NULL);
```
rmccullagh: If we're checking for `c != NULL`, why isn't they're also a NULL check for `tokens`?

dormando: I think the assert() there is cargo culted by me and could be done once at the top of the process_* stack. Tokens can't be null because you can't get to that function if it's null.

PR #172, opened 2016-06-24
Fix undefined behavior in vlen check

elfchief:
The size check for vlen in process_get_command was relying on
undefined behavior (signed integer overflow) to check to see
if vlen was greater than 2^31-2. This made certain compilers
optimize out part of the conditional and prevent memcache from
erroring when the specified size was exactly 2^31-1, leading to
an incorrect error message and a core dump. This patch does that
overflow test in a not-undefined-behavior way.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3234,11 +3234,11 @@ static void process_update_command(conn *c, token_t *tokens, const size_t ntoken
         }
     }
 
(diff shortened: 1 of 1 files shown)


dormando:
bonus points if the commit message notes which test it fixes under what platform/compiler :)

PR #176, opened 2016-06-28
systemd hardening

candrews:
memcached should be isolated and restricted as much as possible to improve security so that if memcached is compromised, the damage will be limited.


dormando:
Thanks! would you mind adding some extra comments as to the intent of each option? Will help users (and probably developers) if in the future something in the code changes which gets tripped up by these.

candrews:
@dormando I don't really think such comments are necessary because:
* the systemd documentation is easy to find and quite comprehensive
* other software (include distros) that have systemd units don't include such comments

But, I'm happy to add it - so I've made a commit doing so.

dormando:
I totally understand but appreciate the comments. It's not super intuitive but it does help people.

PR #180, opened 2016-07-06
Update memcached.c to allow stats cachedump on slab 63

davidfoliveira:
It's returning an error when we try to cachedump the slab 63


dormando:
merged, thanks

PR #181, opened 2016-07-07
Chunked (large) item support

dormando:
Work in progress PR for chunked item support. As of writing this is largely functional, but is incomplete and may need a refactor.

---
 * What is large item support?

Items have always required a single chunk large enough to hold its header, key, and data. This is done via a slab allocator, which is split into many sized slabs. The "largest available slab", became the largest item you could store into memcached. This is where the 1 megabyte limit came in.

The larger the item size limit, the less efficient the slab allocator is, as there is more space between slab classes.

With this change, the max slab chunk size is decoupled from the max item size, and items larger than the max chunk size now consist of multiple chunks chained together.

This has a number of benefits:
 * Performance is nearly identical, as larger items will fill bandwidth much earlier than small items.
 * The max slab class size can be much smaller (I've been settling toward 16k after testing), down from 1mb. This means the rest of the slab classes can be closer together.
 * Since larger items use data at a much smaller granularity (ie 16k), they have less memory overhead than before.
 * The item size max can be hundreds of megabytes without any major impact.

There are some downsides:
 * It complicates the slab allocator a bit and adds a few extra code paths in reading/writing items.
 * Freeing a single item no longer guarantees enough memory to store a new single item. Pulling chunked items to free will need some careful tuning. This may not be awful considering very small items will continue to be 1:1, and the number of loops necessary to free memory for large items is reasonable given the bandwidth required to store them.

---
 * The current status

This largely works. It's been bench tested but not torture tested. Some tests have been written and I've fiddled with it manually for a while. There's a decent amount of code left to write, and I feel like it might need a little rewriting or simplification. This implementation is making the deep slab allocator more aware of how items are structured which feels like a layer violation: items.c should instead ask the slab allocator for a specific number of chunks, but then we might have to loop through them twice, or pass a callback function into the slab allocator to initialize the chunks. Moving more of the implementation up to the item layer would allow it to more intelligently garbage collect when freeing up chunks for a new chunked item as well.

Could use some feedback on any obvious bugs I've left in.

Work remaining:

- [x] Write more tests (+ fix more existing tests + tests for 00-startup.t)
- [x] Protect binary sasl auth from getting a chunked item
- [x] Implement chunked-aware append/prepend commands
- [x] Implement chunked-aware slab page mover. This is also complicated.
- [x] Run burn-in test
- [x] Clean out debug code

Punting or doing later:

- [ ] Test using readv to reduce number of read() syscalls on store.

~~Run burn-in tests on very large items (hundreds of megabytes)~~
~~Any useful stats that could be added? Something to indicate how many large items have been seen, at least.~~

### DIFF
--- a/items.c
+++ b/items.c
@@ -61,6 +61,7 @@ static item *tails[LARGEST_ID];
 static crawler crawlers[LARGEST_ID];
 static itemstats_t itemstats[LARGEST_ID];
 static unsigned int sizes[LARGEST_ID];
+static uint64_t sizes_bytes[LARGEST_ID];
 static unsigned int *stats_sizes_hist = NULL;
 static uint64_t stats_sizes_cas_min = 0;
 static int stats_sizes_buckets = 0;
@@ -89,7 +90,7 @@ void item_stats_reset(void) {
 }
 
 static int lru_pull_tail(const int orig_id, const int cur_lru,
-        const unsigned int total_chunks, const bool do_evict, const uint32_t cur_hv);
+        const uint64_t total_bytes, const bool do_evict, const uint32_t cur_hv);
 static int lru_crawler_start(uint32_t id, uint32_t remaining);
 
 /* Get the next CAS id for a new item. */
@@ -120,7 +121,7 @@ static unsigned int noexp_lru_size(int slabs_clsid) {
     id |= NOEXP_LRU;
     unsigned int ret;
     pthread_mutex_lock(&lru_locks[id]);
-    ret = sizes[id];
+    ret = sizes_bytes[id];
     pthread_mutex_unlock(&lru_locks[id]);
     return ret;
 }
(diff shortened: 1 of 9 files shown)


dormando:
It's been passing the burn-in tests, finally.

Just did some cleanup and pushed the rebalancer updates, which were very challenging. Normally when updating the slab rebalancer I spend several hours drawing out interactions; I did not do that this time and suffered for it.

Most of the pain was preserving the (recent!) ability of the slab rebalancer to avoid evicting items when it has some free chunks in the slab class. The magic trick now extends to being able to rescue individual chunks from within an item: If a large, chunked item has a chunk that spans into a slab page that needs to be moved, the chunk can be swapped out with free memory without evicting or rewriting the entire item. If no free memory is available the entire item gets removed, of course.

This is going to need some refactoring once it's had time to sink in: There're too many layer violations between memcached.c, items.c, and slabs.c, but thinking through all of this on a higher level has given me ideas for the future.

More test cleanup is needed, then several runs through the buildbots and a looped-test to kick out some flaky tests. I'm low on TODO's so this should be very close to ready.

PR #185, opened 2016-07-18
Modernize unit file

remicollet:
AFAIK, "systemctl edit" is supported everywhere (including on RHEL-7) and is simpler to use (this also run a "systemctl daemon-reload")


dormando:
Were there any releases of major distros using systemd that don't have this feature? People tend to run memc on old stuff :(

remicollet:
@dormando old major distro don't have systemd ;)

The oldest I know is RHEL-7 with systemd 219 which support this feature.

dormando:
relased, thanks!

PR #189, opened 2016-08-01
Add missing parameters and escape hypens as minus

guikcd:
* -A, -vvv and -V were missing
* Precision on -t
* Escape hyphen as minus (old versions of groff)


dormando:
released, thanks!

PR #190, opened 2016-08-01
Handle end of line comment on memcached.conf

guikcd:
I've included this for the Debian project.
From http://code.google.com/p/memcached/issues/detail?id=283


dormando:
released, thanks!

PR #191, opened 2016-08-08
Add MemoryDenyWriteExecute to the systemd service

candrews:
MemoryDenyWriteExecute=true will prevent attempts to create memory mappings that are both writable and executable at the same time. This option improves service security, as it makes harder for software exploits to change running code dynamically.

This option was added in systemd 231. Since systemd ignores unknown options, it will have no effect on (but will also cause no problems for) users of systemd < 231.


dormando:
released, thanks!

PR #193, opened 2016-08-11
LRU crawler "metadump" command and refactor

dormando:
Finally! You can dump the cache. metadata.

---
 * New command: `lru_crawler metadump [x,y,x|all]`

ie: `lru_crawler metadump all`

Dumps output in the form `key=foo cas=1\n` etc from the bottom of the cache up to the top.

Notes:

 * This is not a multi-versioned consistent snapshot. This is a live-walk, which means it's possible to get dupes if items shift around, or are overwritten between when the LRU crawler first sees them at the bottom and when it gets to the top of the LRU. You can use the CAS values from the dump to avoid processing duplicates of the same exact key. Keys with different cas ids signify different values.
 * Only one dump can run at a time. This is a necessary restriction to avoid performance problems.
 * The value isn't included with the dump (hence meta[data]dump), which allows clients to fetch values in parallel while processing the feed dump. That is the fastest method of pulling all the data out of the cache.
 * This is highly performant and does hold not locks. It functions via a single side thread which holds locks for short periods of time and releases them between each item examined. There are default short sleeps between every 1000 items examined as well. It can use up to one extra core of CPU while the dump is being processed.
 * IE: My intel NUC can dump 50 million items in under a minute. It isn't that fast of a machine.

---
Other improvements:

 * The internal LRU crawler auto-run code for clearing expired items has been improved significantly. Should burn a lot less excess CPU. Still hoping for more feedback on how that should operate.
 * A number of bugs were fixed along the way, some will likely be backported before this is merged.
 * The LRU crawler is now its own unique file and has an internal plugin-type structure. This means adding further LRU crawler functions is trivial.

This feature and fixes have been sponsored by Netflix. Much thanks goes out to them for investing the time to do this feature correctly and ensure it is thoroughly tested.

As of this writing the feature is still being tested but the PR is ready to be opened and staged while we wait. Although there's more I'd like to do, I can do that on my own in future releases :) At this point the code is stabilized and we're only looking for bugs and some minor additions to the dump output.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,7 +20,8 @@ memcached_SOURCES = memcached.c memcached.h \
                     util.c util.h \
                     trace.h cache.h sasl_defs.h \
                     bipbuffer.c bipbuffer.h \
-                    logger.c logger.h
+                    logger.c logger.h \
+                    crawler.c crawler.h
 
 if BUILD_CACHE
 memcached_SOURCES += cache.c
--- a/crawler.c
+++ b/crawler.c
@@ -1,0 +1,630 @@
+/*  Copyright 2016 Netflix.
+ *
+ *  Use and distribution licensed under the BSD license.  See
+ *  the LICENSE file for full text.
+ */
+
+/* -*- Mode: C; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+#include "memcached.h"
+#include <sys/stat.h>
+#include <sys/socket.h>
+#include <sys/resource.h>
+#include <fcntl.h>
+#include <netinet/in.h>
+#include <errno.h>
+#include <stdlib.h>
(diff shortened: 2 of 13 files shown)


floatingatoll:
Is "does hold locks" meant to read "does not hold locks"?

dormando:
Yes, thank you! Text updated.

dormando:
merged

Tanton1394:
memcached-1.4.31
libmemcached-1.0.18  memcached-2.2.0  PHP 5.6.17

$m = new Memcached();
$m->addServers( array('127.0.0.1', 11211, 100) );
$m->getAllKeys();
echo $m->getResultCode().":".$m->getResultMessage();

result:
9:CLIENT ERROR

What version with recommendations？

PR #198, opened 2016-08-27
Inaccurate binary protocol examples for getk/getkq, flush

Nebopolis:
I have been going through the binary protocol document to implement it in rust, and I identified a few inaccuracies in the example packets provided.
- The length of the body for the example getk/getkq response should be 14 (4 + 5 + 5) bytes long.
- The length of the expirey provided in the flush request is three bytes not four. 

I also noticed a few slight typos/inconsistencies in formatting.

It seems that Github doesn't support making PRs for wiki entries, but these should be updated at https://github.com/memcached/memcached/wiki/BinaryProtocolRevamped as well.

### DIFF
--- a/doc/protocol-binary.xml
+++ b/doc/protocol-binary.xml
@@ -1199,6 +1199,8 @@ Total body   (8-11) : 0x00000000
 Opaque       (12-15): 0x00000000
 CAS          (16-23): 0x0000000000000000
 Extras              : None
+Key                 : None
+Value               : None
             </artwork>
           </figure>
           <figure>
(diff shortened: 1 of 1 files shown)


elfchief:
Hey @dormando, I checked these out and the first and last corrections have since been integrated, but b7395fd hasn't. If you just cherry-picked that one, you could close this PR.

dormando:
Does it look like a correct fix? :) think I just hadn't gone and squinted at it yet.

elfchief:
Yup, that one is correct! The first (550e7b5) was actually incorrect as submitted, but it's already been fixed (correctly) in master. So as long as you just cherry-pick b7395fd, you're golden.

dormando:
up with 1.5.1

PR #205, opened 2016-09-13
Update memcached.c

danial-cao:
the slab_sizes str point store by the slab_sizes_unparsed


dormando:
going into the next branch. since this is a non-functional typo fix I've committed it myself and credited you. (slab_sizes_unparsed is == subopts_value)
thanks!

PR #206, opened 2016-09-22
Fix typo

adamchainz:


dormando:
committed to next branch, but with the new policy of typo fixes committing from the dev with attribution in the commit message. thanks!

adamchainz:
Thanks, will make any pr's against `next` branch in future

PR #208, opened 2016-10-04
Update items.c

akisssa:
Add evicted_time


dormando:
Wow. Crazy how long ago I broke that... and just recently people seem to notice. thanks!

dormando:
need to come back to a test for this. there're a couple good places but requires adding a sleep in a funny way (since it's the latest evicted item that gets the age tag).

Also, I edited the commit message to be descriptive.

PR #216, opened 2016-11-01
Add a feature to restrict writing to local host only

GROpenSourceDev:
We have a requirement to have memcached instances that are visible network wide for reading, but can only be written by a process on the local machine for security reasons. We use it as a buffer in front of a datastore server where clients can read data from the cache from the network, but only the local server can push data into it as it is the authoritative source of data.

We also added a small fix to the logging in memcached as the millisecond type is different (unsigned on gcc/linux, unsigned long on Windows/msys2/gcc) - casting to unsigned long and adjusting the log format should not have any performance impact in a log line that only gets executed if verbosity is >= 2.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -246,6 +246,8 @@ static void settings_init(void) {
     settings.crawls_persleep = 1000;
     settings.logger_watcher_buf_size = LOGGER_WATCHER_BUF_SIZE;
     settings.logger_buf_size = LOGGER_BUF_SIZE;
+    settings.local_write_only = false;
+    settings.anonymous_reads = false;
 }
 
 /*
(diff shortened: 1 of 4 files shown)


dormando:
Thanks for sharing! I'll have to give it some thought. You've pretty specific use case here.

Also thanks for adding tests :)

memcached.c
```diff
@@ -5217,7 +5306,7 @@ static int server_socket_unix(const char *path, int access_mask) {
     }
     if (!(listen_conn = conn_new(sfd, conn_listening,
                                  EV_READ | EV_PERSIST, 1,
-                                 local_transport, main_base))) {
+                                 local_transport, main_base,false))) {
```
greexon: It's required here one more space.

dormando:
Going to close this out as it's too specific of a use case to build into the main code. Might have a different solution down the road. Sorry for not dealing with this years ago, and thanks again for the submission.

PR #217, opened 2016-11-02
Missing necessary header for atomic_inc_64_nv() used in logger.c

pyhalov:
logger.c uses atomic_inc_64_nv() function, but doesn't include atomic.h. This breaks compilation on OpenIndiana.


dormando:
released in 1.4.34

PR #223, opened 2016-11-13
fix compile error.

zzl0:
error: comparison of unsigned expression < 0 is always false

@dormando please review.

cf: https://github.com/memcached/memcached/commit/bd578fc34b96abe0f8d99c1409814a09f51ee71c


dormando:
this isn't right (but I need to look at it anyway). 0 is a valid number, less than zero wouldn't be. other parts of the stack are wrong though.

won't close this until I resolve

zzl0:
@dormando OK, thanks

dormando:
released in 1.4.34

Issue #229, opened 2016-12-07
segfault in version 1.4.33 while using key iterator


vuzilla:
We encountered a memcached segfault in the following situation:
- an administrative app was connected to memcached (TCP connection), using the key iterator feature.
- administrative app was terminated; we believe it was still connected and using the key iterator
- memcached segfault

Here's the log from a system log.  I'm not sure how to get the source code line number from this since the segfault is actually in libc library, but may be caused by an argument passed in from memcached.

Dec  7 02:40:21 ubuntu kernel: [92078.128093] memcached[3870]: segfault at 0 ip 00007fa3993f113b sp 00007fa3962e0940 error 6 in libc-2.19.so[7fa3992e8000+1ba000]

memcached was actively handling get/set/touch traffic at the same time at this:
- get =~ 1600/sec
- set =~ 1300/sec
- touch =~ 2300/sec

Here are the command line parameters:
memcached -p 11211 -u evcache_map_lt -l 0.0.0.0 -m 56253 -C -c 100000 -o slab_chunk_max=524288 -o slab_reassign,slab_automove,lru_maintainer,lru_crawler,hash_algorithm=murmur3 -I 20m

I can send further information, if needed.

dormando:
See #234 - pretty sure that's the bug.

dormando:
released in 1.4.34

vuzilla:
We've verified that this segfault doesn't happen anymore.  Sorry for the delay.

Thanks

Vu

dormando:
Thanks for following up!

PR #234, opened 2016-12-16
Metadump segfault

dormando:
Fix for #229 (and a second one I spotted).

Error handling was so close on this one :) Turned out to be easy to reproduce; which is weird, was fairly sure I tested this code path. Need a better method of extending the unit tests for it.

### DIFF
--- a/crawler.c
+++ b/crawler.c
@@ -402,7 +402,10 @@ static void *item_crawler_thread(void *arg) {
             active_crawler_mod.mod->finalize(&active_crawler_mod);
         if (active_crawler_mod.c.c != NULL) {
             lru_crawler_poll(&active_crawler_mod.c);
-            lru_crawler_release_client(&active_crawler_mod.c);
+            // Double checking in case the client closed during the poll
+            if (active_crawler_mod.c.c != NULL) {
+                lru_crawler_release_client(&active_crawler_mod.c);
+            }
         }
         active_crawler_mod.mod = NULL;
     }
--- a/crawler.c
+++ b/crawler.c
@@ -344,6 +344,9 @@ static void *item_crawler_thread(void *arg) {
                     lru_crawler_class_done(i);
                     continue;
                 }
+            } else if (active_crawler_mod.mod->needs_client) {
+                lru_crawler_class_done(i);
+                continue;
             }
             pthread_mutex_lock(&lru_locks[i]);
             search = do_item_crawl_q((item *)&crawlers[i]);


VictorYYW:
Hi, dormando! I am an undergraduate student doing research on various bugs. Could you please show me the way to reproduce this bug? (the original issue report mentioned _key iteration_ which I can't understand.)

dormando:
@VictorYYW `lru_crawler metadump all` is a command used to asynchronously list item metadata for everything in the cache. It's used for debugging, copying, analysis, etc.

The repro is to load a few thousand/million items into a cache server, run the metadump command, then disconnect the client without reading the results. It'll try deref a null pointer to find the client since metadump wasn't stopping.

VictorYYW:
That make a lot of sense! Thank you, @dormando !

VictorYYW:
Hi, @dormando ! I am still trying to reproduce this bug. I tried to set 9998 key-value pairs and run the metadump command. But the output immediately shows up in the console before I could disconnect the connect. Any hint on how I can "disconnect the client without reading the results".

dormando:
load a million keys.

VictorYYW:
But even after loading 9998 keys, the metadump command only gives around 6000 items. Do you think I should set up more memory when starting the memcached server?

dormando:
set smaller (0 byte) values and/or give the instance more RAM, yes.

VictorYYW:
I am trying to load that many keys into memcached. It's gonna take a long time. Hope that would work. Thank you for your explanation!

dormando:
https://github.com/memcached/mc-crusher/blob/master/bench-warmer

VictorYYW:
That seems to be exactly what I want. You are really helpful! However, when I tried to run the perl script with argument like
```./bench-warmer 127.0.0.1 key 0 1000000```
It seems that four arguments is parsed correctly.(exit on line 17) I verified that the arguments should have been parsed by printing out their value. Besides, I have done some researches on perl since I am totally new to it. If you still have time to look into it, I will very appreciate it.

dormando:
`./bench-warmer 127.0.0.1:11211 key 0 1000000`
you forgot the port, probably?

VictorYYW:
Actually, I have tried to append the port at the very beginning. Unfortunately, it exited with same error.

dormando:
what error?  you didn't list one in your comment

VictorYYW:
Oh, I am sorry. I should have made it clear. It exit at this line
```die "bench-warmer addr prefix size count" unless $addr && $prefix && $size && $count;```
So, I guess it should be the problem that the arguments are parsed properly.

dormando:
heh. set size to 1 instead of 0. dumb bug on my part, I never run it with less than 8. should be checking for defined instead of simply true..

VictorYYW:
Oh, that works. One last question (sorry for asking so many dump questions). How can I simulate the process of close connection before reading?
I tried `echo "lru_crawler memadump all" | nc 127.0.0.1 11211`as well as
establishing connection with telnet, try `lru_crawler metadump all`, use `^]` to escape and `quit`.
But both method can't resulting in segmentation fault of memcached.

dormando:
That's all I did to reproduce it. are you testing the correct version?

otherwise you can write a small program that connects, runs the command, then disconnects without calling any read's. or try loading even more items (5-10 million)

VictorYYW:
What do you mean by "calling reads"?

dormando:
The read() syscall.

VictorYYW:
I tried to use gdb on telnet to break before any read's. But in that case, `lru_crawler metadump all` gives no output and even `^]` or `^c` or `^d` can't help me escape the program or gdb. Am I missing anything?

dormando:
I'm a little lost. The bug was trivial to reproduce by loading a 1 million keys the ^C'ing a netcat. are you testing against the wrong version of memcached? the bug was patched months ago.

VictorYYW:
You are right! The bug should be reproduced easily. I made a very silly mistake when I was using docker to reproduce this bug. I simply forgot that memcached server is running in the background, so the error message will not show up in the console.

Thanks a lot, @dormando ! Really appreciate your time!

PR #250, opened 2017-01-29
Lru move buffer

dormando:
This PR is based off of #240

See commit message; with this PR the LRU algorithm now has a much better balance of not losing active items while still completely avoiding LRU relinks inline with get requests.

Work left:
- [x] Bench testing.
- [x] Decide on if the buffer size should be tunable. [doesn't seem necesary; at pretty high rates it's hard to overwhelm at 8k]
- [ ] Use unique settings to enable? Could be useful to use async bumps with the classic LRU. [punt]
- [x] Either remove the stats counters or surface them somewhere. Considering the hot path it might be best to remove (or stick to dropped only)


dormando:
going to be bullish on this one.

dormando:
wonder why this thinks it's not merged :/

PR #255, opened 2017-02-10
Settings command

ScottMansfield:
DO NOT MERGE - this will need more work. At the moment this is missing tests and a design review. I am creating this PR to review the design I have chosen before I go re-learn perl and write tests for it all.

Another caveat: I only have 64 bit x86 machines to test on, so I don't know if there will be issues with other architectures. The reason I say this is that I am simply doing an assignment to the settings struct and not doing any sort of atomic operation the values are ints, which should be atomic as they are machine size, but I don't have any guarantee. I don't think a torn read would matter much anyhow. I don't think the OS will matter but it might.

Adding a settings command:

The settings command includes 3 possible variations:

settings get
 - gets all settings, acts as a synonym for 'stats settings'

settings get <setting>
 - gets a single setting value
 - if it's an invalid setting, you get an error

good command:
`settings get reqs_per_event`

reply:
```
STAT reqs_per_event 20
END
```

bad command:
`settings get non_existent_setting`

reply:
`CLIENT_ERROR invalid setting`

And the last variation allows you to set any settable setting to a different value.
Things that are settable don't include things like the growth factor but do include
the reqs_per_event, for example.

`settings set <setting> <new value>`

good command:
`settings set reqs_per_event 40`

reply:
`OK`

bad command:
`settings set growth_factor 1.8`

reply:
`CLIENT_ERROR invalid setting`

The settings that can be set are the ones that are not set anywhere else at the moment (via other commands) and are scalars that are easily changeable and are just read in some places.

They are:
- verbosity - settings.verbose
`settings set verbosity <0|1|2|3>`

- evictions - settings.evict_to_free
`settings set evictions <on|off>`

- reqs_per_event - settings.reqs_per_event
`settings set reqs_per_event <num > 0>`

- expirezero_does_not_evict - settings.expirezero_does_not_evict
`settings set expirezero_does_not_evict <yes|no>`

- idle_timeout - settings.idle_timeout
`settings set idle_timeout <num > 0>`

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -2705,6 +2705,7 @@ typedef struct token_s {
 
 #define COMMAND_TOKEN 0
 #define SUBCOMMAND_TOKEN 1
+#define STAT_TOKEN 2
 #define KEY_TOKEN 1
 
 #define MAX_TOKENS 8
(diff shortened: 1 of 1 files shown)


dormando:
Hey,

Thanks for starting this! I'll roll it around in my head a bit more but to start:

- `settings get [etc]` looks nice for completeness but adds a redundant code path. Adding a new settings would require adding it in three places now (two areas of code + protocol.txt). (it's actually 4-5 but meh).
- redundancy with some other commands. phase those out? support them forever? Not sure I can reasonably remove "verbosity"
- I guess it doesn't (yet) make it obvious which settings are runtime tunable, which could lead to frustration as people guess.
- Probably would be nice to move the tested/supported but still recentish ones in there. like cache_memlimit, idle_timeout, etc.

Did you see https://github.com/memcached/memcached/pull/251/files or other recent changes? It probably is a mistake that I've been occasionally naming stats and settings commands differently :/ but things like the LRU commands are more specific to purpose and a bit easier for shorthand; there's one command for changing the LRU tunables instead of 4. "temp_ttl" internally is an on/off along with the actual TTL setting, but the command auto-enables when setting a TTL and disables if the setting is negative. This ideally lowers confusion if users didn't notice there's a separate temp_ttl option.

But, I dunno! There's still some flexibility for the recent commands. I'd welcome some thoughts there.

ScottMansfield:
Sorry for the long delay on my reply here, been a rough week. The following sentences should all be prepended with "In my opinion" or suffixed with "but that's just my opinion," if you prefer.

The individual `settings get <thing>` is one of the liberties I took when putting this together. I'm not opposed to just ripping it out and saying people need to do a very minor amount of extra work to get at the specific setting they want.

If that is removed, I'd vote to deprecate stats settings and instead use `settings get`. The output of `settings get` could include a flag to indicate if it's settable or not, or maybe that could be a documentation task.

After looking through pull 251 it's starting to make more sense to me to split based on subsystem, e.g. the lru command. This PR may be heading for the trash bin.

dormando:
I'm not entirely sure I can deprecate an interface like `stats settings` - that's the downside of being OSS :)

The LRU command is going out into a release but I'm still not convinced it should stay that way. A more generic setting command could still supplant it I think. Would be nice if I could get a wider opinion.. did you post a link to this on the mailing list? if not, go for it.

dormando:
I'm going to close this out. We should keep in touch about it though. I feel bad about all the random commands I've been adding.

PR #257, opened 2017-02-18
Sleep more aggressively in some threads

grantisu:
The logger and lru maintainer threads both adjust how long they sleep
based on how busy they are. This adjustment should be exponential, to more quickly adjust to workloads.

The logger thread in particular would only adjust 50 microseconds at a
time, and was capped at 100 milliseconds of sleep, causing many
unnecessary wakeups on an otherwise idle dev machine. Adjust this cap to 1 second.

### DIFF
--- a/items.c
+++ b/items.c
@@ -1165,7 +1165,9 @@ static void *lru_maintainer_thread(void *arg) {
         }
         if (did_moves == 0) {
             if (to_sleep < MAX_LRU_MAINTAINER_SLEEP)
-                to_sleep += 1000;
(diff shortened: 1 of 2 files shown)


dormando:
I'm giving this some thought. The logger thread should definitely have a higher max wait, but I'm worried about the wait time increasing too quickly since that's especially bad for the LRU thread.

Thanks!

dormando:
Yeah this is the right thing to do.. I updated your change for the newer code, as slab classes are individually scheduled now.

it'll go out in the next release.

PR #262, opened 2017-02-23
refactor chunk chaining for memory efficiency

dormando:
Memory chunk chains would simply stitch multiple chunks of the highest slab
class together. If your item was 17k and the chunk limit is 16k, the item
would use 32k of space instead of a bit over 17k.

This refactor simplifies the slab allocation path and pulls the allocation of
chunks into the upload process. A "large" item gets a small chunk assigned as
an object header, rather than attempting to inline a slab chunk into a parent
chunk. It then gets chunks individually allocated and added into the chain
while the object uploads.

This solves a lot of issues:

1) When assembling new, potentially very large items, we don't have to sit and
spin evicting objects all at once. If there are 20 16k chunks in the tail and
we allocate a 1 meg item, the new item will evict one of those chunks
inbetween each read, rather than trying to guess how many loops to run before
giving up. Very large objects take time to read from the socket anyway.

2) Simplifies code around the initial chunk. Originally embedding data into
the top chunk and embedding data at the same time required a good amount of
fiddling. (Though this might flip back to embedding the initial chunk if I can
clean it up a bit more).

3) Pulling chunks individually means the slabber code can be flatened to not
think about chunks aside from freeing them, which culled a lot of code and
removed branches from a hot path.

4) The size of the final chunk is naturally set to the remaining about of
bytes that need to be stored, which means chunks from another slab class can
be pulled to "cap off" a large item, reducing memory overhead.

As of this commit memory_requested does not work, and binary protocol does not
work. Passes most tests.

- [x] fix binary protocol. the \r\n isn't being allocated dynamically
- [x] figure out the lock safety issues; might have to move the flag fiddling and chain changes under the slab lock.
- [x] memory_requested does not work
- [x] add append/prepend tests within the main chaining tests, rather than just under binprot? [ok for now. need to refactor some more tests]
- [x] make a decision on allocating the first chunk as an object header with an empty internal chunk or with internal spacing. It does require some special code but if it boils down to just 1-2 more lines it would reduce the allocations a bit. [bailing: it's doable but requires more refactoring than I want to do right now]

### DIFF
--- a/items.c
+++ b/items.c
@@ -169,38 +169,14 @@ static size_t item_make_header(const uint8_t nkey, const unsigned int flags, con
     return sizeof(item) + nkey + *nsuffix + nbytes;
 }
 
-item *do_item_alloc(char *key, const size_t nkey, const unsigned int flags,
-                    const rel_time_t exptime, const int nbytes) {
-    int i;
-    uint8_t nsuffix;
+static item *do_item_alloc_pull(const size_t ntotal, const unsigned int id) {
     item *it = NULL;
-    char suffix[40];
-    // Avoid potential underflows.
-    if (nbytes < 2)
-        return 0;
-
-    size_t ntotal = item_make_header(nkey + 1, flags, nbytes, suffix, &nsuffix);
-    if (settings.use_cas) {
-        ntotal += sizeof(uint64_t);
-    }
-
-    unsigned int id = slabs_clsid(ntotal);
-    if (id == 0)
-        return 0;
-
+    int i;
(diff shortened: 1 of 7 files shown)


cjnadz:
Needs/review testing

cjnadz:
Wip

dormando:
no longer WIP. need to do a squash down and maybe add some macros, but if it survives torture tests it's good to go.

dormando:
after some huffing, the last major bug turned out to predate this patch... running through extra mc-crusher now.

PR #273, opened 2017-05-05
Avoid segfault if idle_timeout value is missing.

thatsafunnyname:
If memcached is called without a numeric argument for the idle_timeout option, so:
  -o idle_timeout
and not:
  -o idle_timeout=60
it segfaults inside stdlib/strtol_l.c:: ____strtol_l_internal from stdlib.h::atoi from the IDLE_TIMEOUT case.
I am using glibc-2.12 on RHEL6.  I think atoi(NULL) is undefined behaviour.
This change guards against this for the the idle_timeout option, in the same manner as the cases that use the subopts_value, printing that the numeric argument is missing for the idle_timeout option.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -6303,6 +6303,10 @@ int main (int argc, char **argv) {
                 settings.temporary_ttl = atoi(subopts_value);
                 break;
             case IDLE_TIMEOUT:
+                if (subopts_value == NULL) {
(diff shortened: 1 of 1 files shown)


dormando:
thanks!

I should probably make a macro for that check... also need to finish converting everything off of atoi.

dormando:
looks good, merged for next release. thanks again!

PR #275, opened 2017-05-11
Add drop_privileges() for Linux (base: next)

viraptor:
Implement an aggressive version of drop_privileges(). Additionally add
similar initialization function for threads drop_worker_privileges().

This version is similar to Solaris one and prohibits memcached from
making any not approved syscalls. Current list narrows down the allowed
calls to socket sends/recvs, accept, epoll handling, futex (and
dependencies - mmap), getrusage (for stats), and signal / exit
handling.

Any incorrect behaviour will result in EACCES returned. This should be
restricted further to KILL in the future (after more testing).

The feature is only tested for i386 and x86_64. It depends on bpf
filters and seccomp enabled in the kernel. It also requires libsecomp
for abstraction to seccomp filters. All are available since Linux 3.5.

Seccomp filtering can be enabled at compile time with --enable-seccomp.

In case of local customisations which require more rights, memcached
allows disabling drop_privileges() with "-o no_drop_privileges" at
startup.

Tests have to run with "-o relaxed_privileges", since they require
disk access after the tests complete. This adds a few allowed syscalls,
but does not disable the protection system completely.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -47,6 +47,33 @@ If you export the environment variable
 T_MEMD_USE_DAEMON="127.0.0.1:11211" the tests will use an existing
 daemon at that address.
 
+* Debugging seccomp issues
+
+If new functionality fails when built with seccomp / drop privileges
+support, it can be debugged in one of two ways:
+
+Run the memcached via strace. For example:
+
+    strace -o /tmp/memcache.strace -f -- ./memcached
+    less /tmp/memcache.strace
+
+And look for calls which failed due to access restriction. They will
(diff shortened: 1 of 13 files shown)


dormando:
out (after years! holy crap sorry) with 1.5.1. thanks for coming back so many times to fix things!

viraptor:
No worries. I understand this may be a big change and could run into edge cases in real world situations. Don't hesitate to mention me on any related issue - happy to help with them.

Also - yay sandboxing! Thanks for merging!

PR #281, opened 2017-06-06
Add curr_connections_pct to "stats".

thatsafunnyname:
Thanks for memcached.

This PR is to consider adding "curr_connections_pct" to "stats", so that percentage based threshold monitoring of the number of connections does not have to also query "stats settings" to calculate a percentage.
 
I noticed the "stats" output has had some items added ( https://github.com/memcached/memcached/blame/master/memcached.c#L2988 ), I thought I would chance a PR (maybe for the 1.5 release?).  Thank you for your consideration.

Related:
https://github.com/memcached/memcached/wiki/ServerMaint#important-stats 
https://code.google.com/archive/p/memcached/issues/206

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -628,6 +628,8 @@ integers separated by a colon (treat this as a floating point number).
 | bytes                 | 64u     | Current number of bytes used              |
 |                       |         | to store items                            |
 | curr_connections      | 32u     | Number of open connections                |
+| curr_connections_pct  | 32u.32u | Number of open connections as a           |
+|                       |         | percentage of the maximum number allowed  |
 | total_connections     | 32u     | Total number of connections opened since  |
 |                       |         | the server started running                |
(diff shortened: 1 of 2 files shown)


dormando:
I'll have to give it some thought. It's a little bit of a slippery-slope since a number of stats could be calculated that way.

For what it's worth, a lot of my monitoring scripts simply send something like "stats\r\nstats settings\r\n" (or "stats items\r\nstats slabs\r\n") and then just read until the second END.

dormando:
thinking of a compromise: since maxbytes is in both stats and stats settings, it makes some sense for max_connections to be in both. It's not a great general case to just hand people a percentage since they may want to cap to a specific value, or etc.

I can whip this up or you can resubmit if you get to it before me :) hope this is okay.

dormando:
went with adding a `max_connections` stat in the stats output.

dormando:
out with 1.5.1

thatsafunnyname:
Thanks.

PR #283, opened 2017-06-21
Slab automove algorithm revisit

dormando:
A few fixes:

 * If a slab page move loops too often, delete lockable busy items from the cache to unjam it.
 * Improve page balancer python script, adding test util.
 * Convert python script into C, use as new overall mover algorithm.

This can be tuned for more edge cases but I'd prefer to defer.


dormando:
it's had some testing. it's pushed to next. early and often, hopefully doesn't bite me too hard.

PR #285, opened 2017-07-05
systemd instancing support & rpm build improvements

elfchief:
The major things this does are adding systemd support to the rpm .spec
file, and adding systemd instancing support. This means that it is
possible to run multiple memcached instances without having to do
any additional configuration or hack on init scripts.

To use:
    systemctl start memcached@11211 memcached@11311 memcached@11411

sysconfig files at /etc/sysconfig/memcached.<port> will be read as
appropriate, to allow differing configurations per-port. Defaults
will be read from /etc/sysconfig/memcached before the port-specific
settings are read.

You can also still start memcached the standard way just by doing
"systemctl start memcached". This will read /etc/sysconfig/memcached
and nothing else.

The "enhanced security" lines in the existing systemd unit file have
been commented out, because they ause an error on common older versions
of systemd (like what ships with Redhat/CentOS 7). I'm not sure if we
want to have this file default to "what works everywhere" or "what is
more secure"; I can back out this change if needed.

There are two versions of the .service file included, one for standard
memcached invocations and one for instanced invocations. The two are
very similar, but not identical. Ideally, we'd only have one version
in the source tree and we'd massage it with sed or somesuch during the
rpm build, but couldn't think of a super clean way to do that, so erred
on the side of simplicity.

A decent amount of spec file work was needed to enable this functionality.
In the process, I also cleaned up several additional aspects of the spec
file (like using %{name} in places where it was appropriate). I also
commented out the automatic restart in the %postun section, for two main
reasons:

        1. The try-restart for instanced memcached will produce an error if
        instanced memcached isn't in use, which is probably quite confusing to
        people who aren't using that functionality and are just trying to update
        their package. (There's workarounds for this, but I try to keep pre/post
        scripts as simple as humanly possible)

        2. Automatic restarts on updates means the cache gets flushed, which
        means you can no longer safely use large-scale management tools (like
        puppet or chef) to roll out new versions, at least not without a lot
        of planning first. Not automatically dumping someone's caches feels
        safer, here.

### DIFF
--- a/memcached.spec.in
+++ b/memcached.spec.in
@@ -1,3 +1,21 @@
+# Set with_systemd on distros that use it, so we can install the service
+# file, otherwise the sysvinit script will be installed
+%if 0%{?fedora} >= 14 || 0%{?rhel} >= 7 || 0%{?suse_version} >= 1210
+%global with_systemd 1
+BuildRequires: systemd-units
+
+# Disable some systemd safety features on OSes without a new enough systemd
+# (new enough is systemd >= 233)
+%if 0%{?fedora} < 26 || 0%{?rhel} > 0
+%global safer_systemd 0
+%else
+%global safer_systemd 1
+%endif
+
+%else
+%global with_systemd 0
+%endif
+
 Name:           memcached
 Version:        @VERSION@
 Release:        @RELEASE@%{?dist}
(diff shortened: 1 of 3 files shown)


elfchief:
FWIW, there's a prebuilt RPM you can test this functionality with (if you have CentOS 7 or probably most versions of Fedora) at http://www.lupine.org/RPMS/memcached-1.4.36-5.wwp.el7.centos.x86_64.rpm . Slightly older version of memcached itself, but has all of these changes.

dormando:
nice. will look as soon as I can.

dormando:
apparently "as soon as I can" continues to be "way too long".

the extra security stuff not being supported in the latest centos kind of sucks. is there any way to detect if they're supported or not at install time?

elfchief:
At install time... well, probably, but it would be kind of ugly (having the unit file on disk be different from what's in the RPM is undesirable). At _build_ time ... sure (or just specify it as a build option), which should be fine, since RPMs are generally built for each major version of each given platform anyhow. Mostly just a matter of finding an easy way to manipulate the file at build time.

...let me think about that one. Should be possible to put something together.

dormando:
any luck? I'm doing another cut soon.

dormando:
Hey, poking on an old PR :P I kinda want this in for the next release; pretty sure I'll just merge it (though I should get tests for rpmbuild's working...). I'll leave it open a bit longer but if for some reason you're around and want to look at that last question again that'd be sweeeellllll

elfchief:
I am still around! Just ... really run down these days. :(

Let me see if I can poke at things some tonight.

elfchief:
Hokay, tree is updated. I added some conditionals to comment out (or not) the potentially unsupported lines. Right now it comments those lines out on RHEL/CentOS7, and on fedora < 26, and enables those lines on everything else. (Sadly, there's not really a clean way to determine if the systemd version is new enough (233 or later) at either build time or install time, so if like RHEL 7.5 suddenly has a newer systemd, we'll have to update the spec file. This isn't the end of the world, at least.)

A built and testable version for CentOS 7 is at https://www.lupine.org/RPM/memcached-1.5.5_1_g5dff36d-1.el7.centos.x86_64.rpm if you want to play with it.

dormando:
Sorry to hear, and thanks for re-appearing! I've shoved this into the 'next' branch with a "looks good to me" - since I can barely read RPM spec despite having written it on and off for over a decade.

Hopefully another pass at documentation would add rpm building/etc. I like the commit and all of the fixes in here, and I've definitely seen people ask about multi-instance systemd.

thanks again!

dormando:
went out in 1.5.6

Issue #265, opened 2017-03-02
Some way to have memcached hangup tcp sockets that are erroring


neerolyte:
Currently if you send memcached a bad command, it just errors and waits for a new command:

```
asd
ERROR
set foo 0 0 3
bar
STORED
```

I'd like some way to have it drop that tcp socket on any error - I couldn't see this in the doco, is it possible already?

Basically I'd like to harden up memcached so that if there's an open http proxy or other relay that could be used to send in some arbitrary data over tcp, it's most likely that memcached will drop the tcp socket before any attacker controlled data is executed.

It's not as relevant for UDP (which can't be hung up on anyway) or unix sockets as a http proxy seems like the most likely way someone would be getting to an internal memcached and that should be limited to TCP.

dormando:
not currently possible.. there are a lot of errors which do cause the connection to close but that's not one of them.

it's possible to add a setting that could do that for the unknown command case.. if you issue malformed other commands the error handling is deeper in and that's harder.

can't switch the defaults right off because it might goof people's existing systems. it's something I could add to -o modern and flip to default at a later date...

also probably need similar for binprot.

dormando:
Ah right; if the binprot sees a malformed magic header it closes the connection. so it probably only needs to be done for ASCII.

neerolyte:
Not sure if this would be easier but if you could some how alias all the normal HTTP verbs to exit commands and/or detect `<something verblike> .+ HTTP/.*` and treat that as an error that closes the connection - that'd mitigate issues from **almost** all open HTTP proxies.

I noticed while playing with postfix for much the same issue that it also liberally accepts real commands after bogus ones but drops out on HTTP verbs:

```
bash-4.1# nc localhost 25
220 f722397fcdcf.localdomain ESMTP Postfix
arbitrary
502 5.5.2 Error: command not recognized
junk
502 5.5.2 Error: command not recognized
POST
221 2.7.0 Error: I can break rules, too. Goodbye.
bash-4.1#
```

Although it doesn't block enough verbs, so you can do a `PUT / HTTP/1.0`, followed up by valid SMTP and have it send an email anyway - hence it'd be nice to scan for the HTTP pattern.

I'd prefer the strict behaviour, but if it's quicker to blacklist a couple of terms to block known attacks that's still pretty useful :)

neerolyte:
[...]

PR #291, opened 2017-07-27
Drop sockets from obviously malicious command strings

neerolyte:
This fixes most of the issue I have with #265.

Basically if memcached receives a string that's obviously someone doing something malicious, just drop the socket.

I also added a test for the `quit` command over the text protocol - I couldn't find a test that did this, I copied the basics of it from the binary protocol test.


dormando:
squashed, fixed minor style issue, merged into 'next' branch for next release.

thanks for doing this!

dormando:
out with 1.5.1

PR #300, opened 2017-08-30
Rework pledge support after seccomp support has been added

bigio:
Re-add pledge support,
now that seccomp(2) support has been imported and there is a save_pid(...) function it is even easier because it doesn't need anymore permissions to create pid files.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -38,6 +38,10 @@ if BUILD_LINUX_PRIVS
 memcached_SOURCES += linux_priv.c
 endif
 
+if BUILD_OPENBSD_PRIVS
+memcached_SOURCES += openbsd_priv.c
+endif
+
 if ENABLE_SASL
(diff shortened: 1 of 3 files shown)


dormando:
awesome, thanks!

dormando:
I don't actually have an OpenBSD buildbot right now :( but this is very short and makes sense. Sorry for the long wait; had a queue of stuff to go out first.

bigio:
Il 4 novembre 2017 21:24:28 CET, dormando <notifications@github.com> ha scritto:
>I don't actually have an OpenBSD buildbot right now :( but this is very
>short and makes sense. Sorry for the long wait; had a queue of stuff to
>go out first.

I have committed this diff to OpenBSD even before submitting it here and there has been no regressions.
 Thanks
     Giovanni

PR #302, opened 2017-10-03
Adding missing get and touch command for ascii protocol

theonajim:
Thank you for your consideration.


dormando:
The key gets printed via item_get(), called from item_touch(), if verbosity is high. so that's fine.

The code looks fine, but it's not idiomatic with the ASCII protocol unfortunately; you're matching the binary protocol with its CAS return, but normal ASCII only does that on "gets". So you'd more likely end up with a GAT and a GATS deal for CAS vs no CAS. There's also no multi-GAT for ascii here. I'm debating if the lack of multi-get is okay though.

Thanks for the submission! :) Hello Walmart, I think?

theonajim:
@dormando 
Yes, hello from WalmartLabs :) 
My rationale was not to repeat the pattern of "get" vs "gets" since "gets" was introduced for backwards compatibility of "get" (but you know the code history better than me). Since "gat" is a new command, we don't have those same backwards compatibility constraints. Multi-GAT can be introduced at a later time (I might be able to follow up with another pull request)

dormando:
that's correct, backcompat was half of the reason. the other half is adding that extra header takes some amount of time and it was important to preserve performance for people who don't use CAS. This was less important in binprot since it was a standard header and didn't require character conversion.

I *have* optimized that process recently though. But in the grand scheme of things, is GAT just another flag+param to process_get_command, which flips item_get to item_touch? :)

theonajim:
@dormando 
If we use a flag to process_get_command, then we need to designate one of the tokens as the exptime. It seems simplest to use the first key token as exptime. Otherwise, we would need a to break the nested do/while and while loops when reaching the last key_token (and ntokens is modified). 

The command format would be:
gat <exptime> <key>*
gats <exptime> <key>*

dormando:
Yeah, that looks right.

process_get_command is kind of large and gnarly already but it won't grow much if you pre-shift/parse the exptime token. in binprot land get/gat/touch are all the same code path anyway.

dormando:
nice. can you rebase the branch as well? (generally, we work against 'next' branch, but it's presently the same as master). might want to squash before you rebase.

I'll run a crush against it when I get a chance to ensure there isn't any major regression.

theonajim:
Thanks @dormando . I moved the changes to PR https://github.com/memcached/memcached/pull/303

PR #303, opened 2017-10-05
Adding missing get and touch command for ascii protocol

theonajim:
Recreating PR https://github.com/memcached/memcached/pull/302

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3276,8 +3276,14 @@ static inline int make_ascii_get_suffix(char *suffix, item *it, bool return_cas,
 }
 
 #define IT_REFCOUNT_LIMIT 60000
-static inline item* limited_get(char *key, size_t nkey, conn *c) {
-    item *it = item_get(key, nkey, c, DO_UPDATE);
+static inline item* limited_get(char *key, size_t nkey, conn *c, uint32_t exptime, bool should_touch) {
+    item *it;
+    if (should_touch) {
+        it = item_touch(key, nkey, exptime, c);
+    }
+    else {
+        it = item_get(key, nkey, c, DO_UPDATE);
+    }
     if (it && it->refcount > IT_REFCOUNT_LIMIT) {
         item_remove(it);
         it = NULL;
@@ -3332,7 +3338,7 @@ static inline char *_ascii_get_suffix_buf(conn *c, int i) {
 // FIXME: the 'breaks' around memory malloc's should break all the way down,
 // fill ileft/suffixleft, then run conn_releaseitems()
 /* ntokens is overwritten here... shrug.. */
-static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens, bool return_cas) {
+static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens, bool return_cas, bool should_touch) {
     char *key;
     size_t nkey;
     int i = 0;
@@ -3340,7 +3346,19 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
     item *it;
     token_t *key_token = &tokens[KEY_TOKEN];
     char *suffix;
+    int32_t exptime_int = 0;
+    rel_time_t exptime = 0;
     assert(c != NULL);
+
+    if (should_touch) {
+        // For get and touch commands, use first token as exptime
+        if (!safe_strtol(tokens[1].value, &exptime_int)) {
+            out_string(c, "CLIENT_ERROR invalid exptime argument");
+            return;
+        }
+        key_token++;
+        exptime = realtime(exptime_int);
+    }
 
     do {
         while(key_token->length != 0) {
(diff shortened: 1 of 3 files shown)


dormando:
excellent, thanks! Unfortunately it might take me a while to run it through the crusher as I'm busy this weekend. :( I'll get to it as soon as I can.

memcached.c
```diff
@@ -3332,16 +3332,28 @@ static inline char *_ascii_get_suffix_buf(conn *c, int i) {
 // FIXME: the 'breaks' around memory malloc's should break all the way down,
 // fill ileft/suffixleft, then run conn_releaseitems()
 /* ntokens is overwritten here... shrug.. */
-static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens, bool return_cas) {
+static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens, bool return_cas, bool should_touch) {
```
dormando: I can refactor this myself later, but better would be to pass in flags, then convert those to bools at the top of the function.

memcached.c
```diff
@@ -3359,7 +3371,12 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
                 return;
             }
 
-            it = limited_get(key, nkey, c);
+            if (should_touch) {
+                it = item_touch(key, nkey, exptime, c);
```
dormando: this should move into limited_get, of which this function is the only caller. Then you don't lose the refcount protection by calling GAT instead of GET.

memcached.c
```diff
@@ -3379,8 +3396,13 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
 
                 if (return_cas || !settings.inline_ascii_response)
                 {
-                  MEMCACHED_COMMAND_GET(c->sfd, ITEM_key(it), it->nkey,
+                  if (should_touch) {
```
dormando: remove this branch and leave it at GET. should help with perf enough to get it over the line, and it's still more a get than a touch (which doesn't return the data). not sure I care enough for it to be semantically correct.

memcached.c
```diff
@@ -3440,19 +3467,34 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
                 }
 
                 /* item_get() has incremented it->refcount for us */
-                pthread_mutex_lock(&c->thread->stats.mutex);
-                c->thread->stats.lru_hits[it->slabs_clsid]++;
-                c->thread->stats.get_cmds++;
-                pthread_mutex_unlock(&c->thread->stats.mutex);
+                if (should_touch) {
+                    pthread_mutex_lock(&c->thread->stats.mutex);
```
dormando: There isn't really contention on these locks, so it's okay to put them outside of the branches and save a few lines.

memcached.c
```diff
@@ -3440,19 +3467,34 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
                 i++;
 
             } else {
-                pthread_mutex_lock(&c->thread->stats.mutex);
-                c->thread->stats.get_misses++;
-                c->thread->stats.get_cmds++;
-                pthread_mutex_unlock(&c->thread->stats.mutex);
-                MEMCACHED_COMMAND_GET(c->sfd, key, nkey, -1, 0);
+                if (should_touch) {
+                    pthread_mutex_lock(&c->thread->stats.mutex);
```
dormando: same comments as above re: locks, kill the MEMCACHED_COMMAND_TOUCH and move the MEMCACHED_COMMAND_GET back outside of the branch.

t/getandtouch.t
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/perl
+
+use strict;
```
dormando: + use warnings;

memcached.c
theonajim: @dormando 
I will make these changes soon. I am trying to follow the code paths, there seems to be a missing item_remove(it) in following path. Is that intentional? 

if ((it->it_flags & ITEM_CHUNKED) == 0)

https://github.com/memcached/memcached/blob/69a4414178e1bd5f69db7810022a94799d7e1019/memcached.c#L3404

dormando: You mean there's a missing check to see if add_iov() fails if the item isn't chunked?

That's fine there (the iov's tends to get doubled, and it probably never actually fails). I'm going to refactor the loop soon so there's one iov space check at the top of the loop. That should fix it as well.

theonajim:
@dormando 
changes done and commits squashed. Thanks for taking a closer look.

dormando:
That was fast! Now I feel worse about being slow :)

I did the other bug fixes (ended up being easy) but should try to find a few other things to fix/merge to warrant a bugfix cut. Should be a few days or a week at most.

Thanks again!

dormando:
This is released in 1.5.3. If there's anything else you'd like to work on, I have plenty of stuff :)

theonajim:
@dormando 
Thanks for the merge! I would like to work on more stuff :) Can you send me a doc with a list of planned enhancements?

dormando:
@theonajim I don't really have anything so formal :)

First; what nature of changes would you be interested in?

I'm focusing on merging in the flash branch (since rebasing it is a pain in the ass), then I want to spend some time writing up issues for new contributors to chew through. Could be a couple weeks if that's okay.

PR #315, opened 2017-11-19
Apply the cast to the whole expressions, fixing build with clang

devnexen:


dormando:
apologies for the long idle, thanks for the submission! should go out with the next release

dormando:
went out in 1.5.6

Issue #312, opened 2017-11-04
Tests core dump on aarch64


cstrouse:
```bash
chronos@localhost /usr/local/tmp/crew/memcached-1.5.3.tar.gz.dir/memcached-1.5.3 $ make test
./sizes
Slab Stats      64
Thread stats    -6496
Global stats    160
Settings        224
Item (no cas)   32
Item (cas)      40
Libevent thread 96
Connection      336
----------------------------------------
libevent thread cumulative      6400
Thread stats cumulative         6304
./testapp
1..52
ok 1 - cache_create
ok 2 - cache_constructor
ok 3 - cache_constructor_fail
ok 4 - cache_destructor
ok 5 - cache_reuse
ok 6 - cache_redzone
ok 7 - issue_161
ok 8 - strtol
ok 9 - strtoll
ok 10 - strtoul
ok 11 - strtoull
ok 12 - issue_44
ok 13 - vperror
ok 14 - issue_101
Signal handled: Terminated.
ok 15 - start_server
ok 16 - issue_92
ok 17 - issue_102
ok 18 - binary_noop
ok 19 - binary_quit
ok 20 - binary_quitq
ok 21 - binary_set
ok 22 - binary_setq
ok 23 - binary_add
ok 24 - binary_addq
ok 25 - binary_replace
ok 26 - binary_replaceq
ok 27 - binary_delete
ok 28 - binary_deleteq
testapp: testapp.c:729: safe_recv: Assertion `nr != 0' failed.
make: *** [Makefile:1734: test] Aborted (core dumped)
```

dormando:
well.. that was fast.

What platform/etc?

cstrouse:
Sorry, should have included more info. I package `memcached` for use on ChromeOS devices. I have a Samsung Chromebook Plus (aarch64). This machine runs a `4.4.70-11386-g8271523ecf43` kernel, has a 64-bit ARMv8 Processor rev 4 (v8l) processor but runs a mostly 32-bit userland. `libevent` is on 2.1.8 currently. Any other info that would help?

dormando:
Ahh the changes looked trivial so I didn't do the full buildbot suite this time... I'll plug in my rpi for an ARM test locally.

In the meantime; if you want to try bisecting to find the offending commit that could save some time. If I can't repro I'll need the backtrace from the actual segfault (or at least try getting an addr2line from the segfault in dmesg)

dormando:
[...]

PR #317, opened 2017-12-02
Don't use unaligned access on ARM

guikcd:
On ARM, unaligned memory access is not guaranteed to succeed; depending on
the CPU and the kernel configuration, it may raise SIGBUS instead.

memcached defaults to using unaligned access, which caused the package to
fail to build on Ubuntu infrastructure where SIGBUS is raised.

All credits to: Nishanth Aravamudan <nish.aravamudan@canonical.com>
Original bug report: https://bugs.debian.org/879797

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -482,48 +482,7 @@ AC_CHECK_FUNCS(clock_gettime)
 AC_CHECK_FUNCS([accept4], [AC_DEFINE(HAVE_ACCEPT4, 1, [Define to 1 if support accept4])])
 AC_CHECK_FUNCS([getopt_long], [AC_DEFINE(HAVE_GETOPT_LONG, 1, [Define to 1 if support getopt_long])])
 
-AC_DEFUN([AC_C_ALIGNMENT],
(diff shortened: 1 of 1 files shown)


dormando:
Hey,

Well I found out about this a month ago in https://github.com/memcached/memcached/issues/312
- managed to get a repro after compiling things on my rpi3 for six hours.

Unfortunately this patch just forces aligned access for all platforms all the time forever, so I can't accept it :) I'll leave this open until I get time to fix the detector for ARM. It shouldn't be too hard since I do have a repro, but that code is tricky.

dormando:
figured this out. GCC O2 was eliding the alignment test in configure.

fixing it fixes the binprot test, but apparently I have some fairly bad unaligned accesses in the logging code. leaving a note here so I don't forget what I was doing and finish fixing it tomorrow.

dormando:
should be fixed properly in the next branch if you want to give it a try

dormando:
released in 1.5.7

Issue #313, opened 2017-11-09
Support transparent huge pages


wens:
It seems memcached only supports huge pages under Solaris, under which memcntl exists.

With modern Linux and FreeBSD, the kernel supports transparent huge pages.
This means the kernel will merge and migrate normal pages into huge pages.
Note that this is a possibility, but not a certainty. If the system does not have
sufficient free pages available, depending on the OS, the allocation may either
fail or fall back to normal pages.

On FreeBSD this is done by using mmap() with MAP_ALIGNED_SUPER.
On Linux this is done by using madvise() with MADV_HUGEPAGE on an
existing mapping, likely allocated by mmap or posix_memalign.

From what I can understand of the code, nothing is needed at the parameter
check phase, though on Linux it would be a good idea to check
/sys/kernel/mm/transparent_hugepage/enabled . Otherwise
memcached.c:enable_large_pages() can just return 0 on Linux and FreeBSD.

In slabs.c:slabs_init(), malloc() in the prealloc block would need to be replaced
with mmap() and madvise() (on Linux only) with some #ifdefs. The fall back
path stays the same.

Does that sound like a plan?

dormando:
Sounds about right. People will run into issues doing large mmap's, so there should be good error handling to communicate to the user as well.

you giving it a shot? :)

wens:
I think so. Can't give any guarantees on any timeline though. :)

dormando:
If you see how laggy I usually am here, you'll find yourself at home :P

not that I should be proud of that..

dormando:
this went in, but I never closed the issue :)

PR #318, opened 2017-12-13
Linux transparent huge pages support

wens:
Linux has supported transparent huge pages on Linux for quite some time.
Memory regions can be marked for conversion to huge pages with madvise.
Alternatively, Users can have the system default to using huge pages for
all memory regions when applicable, i.e. when the mapped region is large
enough, the properly aligned pages will be converted.

Using either method, we would preallocate memory for the cache with
proper alignment, and call madvise on it. Whether the memory region
actually gets converted to hugepages ultimately depends on the setting
of /sys/kernel/mm/transparent_hugepage/enabled.

If any step of the preallocation fails, we simply fallback to standard
allocation, without even preallocating slabs, as they would not have
the proper alignment or setting anyway.

This is my solution for #313 for Linux. It turned out slightly simpler than I thought.
I'm not sure whether the preallocation bit should be pulled out into a separate
configuration option or not though.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -6437,6 +6437,16 @@ static int enable_large_pages(void) {
     }
 
     return ret;
+#elif defined(__linux__) && defined(MADV_HUGEPAGE)
+    /* check if transparent hugepages is compiled into the kernel */
+    struct stat st;
+    int ret = stat("/sys/kernel/mm/transparent_hugepage/enabled", &st);
+    if (ret || !(st.st_mode & S_IFREG)) {
+        fprintf(stderr, "Transparent huge pages support not detected.\n");
+        fprintf(stderr, "Will use default page size.\n");
+        return -1;
+    }
+    return 0;
 #else
     return -1;
 #endif
--- a/slabs.c
+++ b/slabs.c
@@ -8,6 +8,7 @@
  * memcached protocol.
  */
 #include "memcached.h"
+#include <sys/mman.h>
 #include <sys/stat.h>
 #include <sys/socket.h>
 #include <sys/resource.h>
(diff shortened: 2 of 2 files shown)


dormando:
sorry for the delay, I'll be reviewing this soon! been working on a big feature merge-down

dormando:
Have you done any testing with this, btw?

I've been working on an mmap-based restartable cache thing which probably clobbers some of this, but I think I can rescue it. This looks right in either case, only used when -L is set.

Only question I have is if it should hard-fail if that file doesn't exist (hugepages are disabled), since users may ask for them via -L then simply not get them.

wens:
What sort of testing did you have in mind? My testing involved more functional testing, as opposed to stress testing or performance. Things like checking the exported memory map and stats to make sure that transparent pages were actually used, things were aligned, sizes were properly rounded, etc..

I've not actually put this in production, since we typically use distro packages for those.

wens:
As for the restartable cache, I guess it depends on how you implement it? If you're doing mmap, you'll have to align the size and address by yourself, as opposed to posix_memalign doing it for you, but that's just a minor inconvenience.

As for whether to hard fail or not, I think it should be the same as when -L is not used (if you are planning that case). Or maybe "restartable cache" is a separate option, which would hard fail.
IMHO, -L should be transparent to everything else, except for pre-allocating memory.

dormando:
Fwiw, I didn't forget about this. The current release has a large number of bugfixes, so I'm avoiding adding this in until the next cut. Sometimes it's nice to be mostly-bugfix or mostly-feature change-wise. Will do some performance regression testing too.

dormando:
Next release got delayed a bit :( Working on it now, and just reviewed this. If you're still around and want to fix I'd appreciate, but if not I can try:

https://github.com/torvalds/linux/commit/49920d28781dcced10cd30cb9a938e7d045a1c94

So the size tester only works on kernels newer than late 2016 or early 2017, which will exclude quite a lot of the userbase. A fallback/guess here would widen the usability quite a bit. For instance, I don't actually have a test server or buildbot new enough to even run this :)

Also any kind of documentation/manpage updates are nice.

wens:
2 MiB would probably be a good guess. I know it would work for x86 and ARM. Don't have any other platforms.

On the other hand, we would need some way to detect whether transparent huge page support is compiled in. The presence of /sys/kernel/mm/transparent_hugepage/enabled should work.

I can work on this during the weekends.

dormando:
Ok. I've been bug hunting, anyway.

wens:
I've rebased the patch onto memcached/next, and made it fall back to 2MiB pages when auto-detection of huge page size fails. memcached will also check if the kernel has tranparent huge pages support compiled in.

Note that this new version is only compile tested. I've not had any time to do anything else.

dormando:
@wens did you ever get to test/run this anywhere? I need to come up with a few tests and would love to get this thing in for the release I'm planning this week.

wens:
Just did. Works as expected, either with /sys/kernel/mm/transparent_hugepage/enabled
set to "always" or "madvise" (the default). It will not work when it is set to "never".

To verify, one option is to use strace and look for the madvise call.

Another way is to look at /proc/<PID>/smaps for the allocated block.
Best to use a different size than 64 MiB, as there seem to be several
64 MiB allocations from elsewhere. You'll have to set some values in
the cache so memory actually gets used, then you'll see for that memory
mapping, usage is a multiple of the huge page size.

Two notes come to mind:
a) When /sys/kernel/mm/transparent_hugepage/enabled is set to "always",
there will be some benefit even when not using "-L", but this really depends
on the usage pattern. And the pages are aligned for proper huge page treatment.

b) 2 MiB huge page size is not valid for all platforms. See
https://wiki.debian.org/Hugepages#Huge_pages_sizes
Maybe I should just use the "Hugepagesize" line from /proc/meminfo.
This goes back to 2008 when HugeTLB was introduced.

dormando:
Perhaps. the other platforms seem exotic and/or dead... a new enough kernel would get it right, or can deal with that when someone complains?

Up to you if you want to try parsing it from /proc/meminfo - I'll be doing testing and a mergedown within the next day.

wens:
OK. I updated the patch to use /proc/meminfo, and to bail out if that fails, instead of using some non-universal fallback value.

Also made it preallocate slabs if the huge page allocation succeeded. This was somehow missing from my earlier versions.

So now with the default 64 MiB and huge pages enabled, you'll get the following mapping right after startup:

```
7fc5d6e00000-7fc5dae00000 rw-p 00000000 00:00 0
Size:              65536 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:               40960 kB
Pss:               40960 kB
Shared_Clean:          0 kB
Shared_Dirty:          0 kB
Private_Clean:         0 kB
Private_Dirty:     40960 kB
Referenced:        40960 kB
Anonymous:         40960 kB
LazyFree:              0 kB
AnonHugePages:     40960 kB
ShmemPmdMapped:        0 kB
Shared_Hugetlb:        0 kB
Private_Hugetlb:       0 kB
Swap:                  0 kB
SwapPss:               0 kB
Locked:            40960 kB
VmFlags: rd wr mr mw me ac sd hg
```

dormando:
tabs leaked in (t/whitespace.t fails).

I'll fix myself, but I have to run for a few hours so if you have a chance please do :)

wens:
My bad.

I fixed it and updated the branch.

dormando:
Woot, thanks! merged for next, finally. I'll have to refactor my restart branch + this a little bit, but it didn't look too bad.

dormando:
released in 1.5.9, thanks!

PR #339, opened 2018-02-12
limit crawls for metadumper

dormando:
LRU crawler metadumper is used for getting snapshot-y looks at the LRU's.
Since there's no default limit, it'll get any new items added or bumped since
the roll started.

with this change it limits the number of items dumped to the number that
existed in that LRU when the roll was kicked off. You still end up with an
approximation, but not a terrible one:

- items bumped after the crawler passes them likely won't be revisited
- items bumped before the crawler passes them will likely be visited toward
  the end, or mixed with new items.
- deletes are somewhere in the middle.

### DIFF
--- a/crawler.c
+++ b/crawler.c
@@ -530,6 +530,9 @@ static int do_lru_crawler_start(uint32_t id, uint32_t remaining) {
         crawlers[sid].next = 0;
         crawlers[sid].prev = 0;
         crawlers[sid].time = 0;
+        if (remaining == LRU_CRAWLER_CAP_REMAINING) {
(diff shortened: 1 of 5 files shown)


crawler.c
```diff
@@ -651,7 +654,7 @@ enum crawler_result_type lru_crawler_crawl(char *slabs, const enum crawler_run_t
         }
     }
 
-    starts = lru_crawler_start(tocrawl, settings.lru_crawler_tocrawl,
+    starts = lru_crawler_start(tocrawl, remaining,
```
jmswen: One line?

dormando:
merged

PR #346, opened 2018-02-17
Replace event_init() with event_base_new() for thread safety

qianl15:
Change to use event_base_new() instead of obsolete event_init() for creating
new event base. Also add evthread_use_pthreads() before using any libevent
function to enable thread-safe functions and structures in libevent.

By default, using this combination of APIs will add a lock to the event base.
This can prevent a potential race condition where one worker thread is polling on
its event base and handling requests while a new connection is dispatched and
added to that worker's event base. The performance is unlikely to be affected
by the lock.

This patch may relate to an [ancient issue from 2010](https://groups.google.com/forum/#!topic/memcached/pESfuEq7QAw), or [the one from 2012](https://code.google.com/archive/p/memcached/issues/141). According to [libevent documentation](http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/event__compat_8h.html#a1bf74386dd3725e1538fed2d70c1c113), `event_init()` is deprecated and is totally unsafe for multithreaded use.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -7573,7 +7573,17 @@ int main (int argc, char **argv) {
     }
 
     /* initialize main thread libevent instance */
+#if defined(LIBEVENT_VERSION_NUMBER) && LIBEVENT_VERSION_NUMBER >= 0x02000101
+    /* If libevent version is larger/equal to 2.0.2-alpha, use newer version */
+    struct event_config *ev_config;
+    ev_config = event_config_new();
+    event_config_set_flag(ev_config, EVENT_BASE_FLAG_NOLOCK);
+    main_base = event_base_new_with_config(ev_config);
+    event_config_free(ev_config);
+#else
+    /* Otherwise, use older API */
     main_base = event_init();
+#endif
 
     /* initialize other stuff */
     logger_init();
@@ -7757,6 +7767,9 @@ int main (int argc, char **argv) {
       free(l_socket);
     if (u_socket)
       free(u_socket);
+
+    /* cleanup base */
+    event_base_free(main_base);
 
     return retval;
 }
--- a/thread.c
+++ b/thread.c
@@ -309,7 +309,16 @@ void accept_new_conns(const bool do_accept) {
  * Set up a thread's information.
  */
 static void setup_thread(LIBEVENT_THREAD *me) {
+#if defined(LIBEVENT_VERSION_NUMBER) && LIBEVENT_VERSION_NUMBER >= 0x02000101
+    struct event_config *ev_config;
+    ev_config = event_config_new();
+    event_config_set_flag(ev_config, EVENT_BASE_FLAG_NOLOCK);
+    me->base = event_base_new_with_config(ev_config);
+    event_config_free(ev_config);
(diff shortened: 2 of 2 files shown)


dormando:
So there was an actual bug with very old versions of libevent which was causing those errors. I haven't personally seen a report of one in probably 6 years.

This change probably causes a performance regression; currently workers only modify their own libevents, and nothing changes cross-thread. If you look at thread.c and look around the cq_push/pop/etc stuff, you'll see that new connection dispatch works via pinging a pipe. The pipe is then read as a normal libevent event, and the worker thread modifies its own queue. This avoids having to lock/unlock around every libevent operation when connection handoffs are relatively rare.

qianl15:
Thanks for your detailed explanation! I also worry about the performance regression caused by lock.

But I still think using newer API would be good (more robust and easy to adapt to new features). So in my updated commit, I changed from old API to `event_base_new_with_config()`, and explicitly disabled locking on event_base by using `event_config_set_flag(cfg, EVENT_BASE_FLAG_NOLOCK)`.

dormando:
I'm a little paranoid about people building this against ancient libevents that may not have these API's...

I'll have to check (or you can and update here) on when it was added, in what version, or if it's reasonable to test for them before use.

I'd also probably remove the pthread/thread bits entirely from the libevent stuff until they're actually used, since that may have a side effect of making it harder to build for some people.

qianl15:
I removed pthread/thread bits from my PR entirely.

According to [libevent changelog](https://raw.githubusercontent.com/libevent/libevent/patches-2.0/ChangeLog), `event_base_new_with_config()` and `EVENT_BASE_FLAG_NOLOCK` first appeared in Libevent 2.0.1-alpha.

So here are two choices: 
1) Detect the libevent version: if version < 2.0.1, then use event_init(), otherwise use the newer APIs.
2) Ask users to upgrade to newer version of libevent if they want to use the newer version of memcached.

dormando:
thanks!

2. is definitely not happening :) Many people switch to other key/value stores when posed with the trouble of having libevent on the system at all. Telling them to upgrade in the past has been a total wash, and there's no pressing feature need to force people to do it (yet).

There's an example of a version check on libevent in the code somewhere already.. there was some ancient version which broke everything and it does error on that.

qianl15:
I updated my PR with a version check and choose the corresponding API.

The event_base_free() function is defined in <event2/event.h>. It was first implemented in Libevent 1.2. So it is not affected by the version.

dormando:
does that compile against older versions? :P (sorry, I'm being picky at this point)

qianl15:
Sorry for that :P 

The solution is to check `LIBEVENT_VERSION_NUMBER`. If it is larger/equal to 0x02000101 (>= 2.0.2-alpha), it indicates that we can use newer APIs.

dormando:
rad! pretty sure I'm okay with this :P maybe we'll use more features, maybe we'll rip out libevent. we'll see.

this change goes with the next release, either way.

dormando:
went out in 1.5.6

PR #349, opened 2018-02-28
Support seccomp on musl

### LINKED ISSUE #347: Memcached 1.5.5 is broken on alpine linux
How to reproduce:
```
$ docker run --net=host -it alpine:edge /bin/sh
/ # apk update
/ # apk add memcached
/ # memcached -V
memcached 1.5.5
/ # memcached -u root -l 127.0.0.1
```
When you run `printf "set mykey 0 60 4\r\ndata\r\n" | nc 127.0.0.1 11211` from inside or outside docker, it does not get any response from memcached. Note that `alpine:3.6` with memcached `1.4.36` works fine.

viraptor:
Musl-based distributions like Alpine didn't work with seccomp profile
enabled. This was due to musl using different syscalls than glibc under
the cover.

- writev/readv and epoll_pwait have been allowed
- specific ioctl has been allowed to enableprinting the error/exit
  message
- brk has been allowed at runtime
- worker writes to stderr have been allowed (this was also broken on
  glibc)

Fixes #347

### DIFF
--- a/linux_priv.c
+++ b/linux_priv.c
@@ -2,6 +2,7 @@
 #include <seccomp.h>
 #include <errno.h>
 #include <stdlib.h>
+#include <sys/ioctl.h>
 #include "memcached.h"
 
 // In the future when the system is more tested this could be switched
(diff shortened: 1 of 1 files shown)


dormando:
thanks! for some reason I thought the process would exit if the failure were severe enough. hopefully this fixes it enough, I'll try to validate

viraptor:
Would you prefer if it did?

As per:

https://github.com/memcached/memcached/blob/master/linux_priv.c#L7

This is a simple change. It may be easier to tell that something's wrong that way...

dormando:
Yeah probably. it's still gated behind a compile flag.. it didn't exactly "fail open" this time, it seemed to have "Failed into a confused user" :)

viraptor:
This change is also needed to run on recent glibc. I checked on an updated Arch with glibc 2.26 and it also uses `epoll_pwait` syscall internally.

dormando:
time for another merge round. seems to work locally at least :) I'm going to attempt to add a separate buildbot builder for seccomp before release, but it's merged into next for now.

dormando:
released in 1.5.7

PR #351, opened 2018-03-01
Add Dockerfile definitions

viraptor:
Add alpine (musl-based) and ubuntu (glibc-based) descriptions for test
images. To run the testapp in either of them, run:

```
docker-compose run alpine
docker-compose run ubuntu
```

### DIFF
--- a/devtools/Dockerfile.alpine
+++ b/devtools/Dockerfile.alpine
@@ -3,5 +3,10 @@ FROM alpine:latest
 RUN apk update && apk add --no-cache musl-dev libevent-dev libseccomp-dev linux-headers gcc make automake autoconf
 
 ADD . /src
+WORKDIR /src
 
-CMD cd /src && ./autogen.sh && ./configure --enable-seccomp && make -j && ./testapp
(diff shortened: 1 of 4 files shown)


dormando:
ooh fancy.. I finally got the new buildbot master going late last night. will see if I can figure out how to add a buildworker that does this.

if I wanted to add variants for other configure options, is there a better way than just adding more of these docker files with different names?

viraptor:
Depends what you mean by configure options. You can parameterize the dockerfile and provide the variables in the compose file like the `args` described here: https://docs.docker.com/compose/compose-file/ There are other things like the command to run, which are exposed explicitly as well.

For bigger changes like a different base image though, a new dockerfile is likely a better option.

dormando:
just configure flags for now. ie; I'm pretty sure extstore and seccomp don't work together :)

viraptor:
No, but I'll add it to my mental todo one day. As long as workers don't reopen files with different names at runtime, it should be possible to support.

viraptor:
Added Arch support as well.

PR #358, opened 2018-03-12
Enable Arch seccomp support

viraptor:
Add syscalls found when running on Arch.

PR is based on https://github.com/memcached/memcached/pull/349 so that one should be merged first.


dormando:
wonder how fragile this is going to be going forward :( I guess people can just build without it.

dormando:
released in 1.5.7

Issue #335, opened 2018-02-07
memcached-tool dump: infinite expiration time is not dumped as 0


ronald:
I dumped using `memcached-tool dump` and restored using `nc`.
All items without expiration time (value 0) are missing.

```$ echo -e 'add my_key 0 0 11\r\nhello world\r' | nc localhost 11211
STORED
$ /usr/share/memcached/scripts/memcached-tool localhost:11211 dump 
Dumping memcache contents
  Number of buckets: 1
  Number of items  : 1
Dumping bucket 1 - 1 total items
add my_key 0 1518017432 11
hello world
```

expected result 
```add my_key 0 0 11```

The article https://blog.elijaa.org/2010/12/24/understanding-memcached-stats-cachedump-command/ indicates that the timestamp must be combined with uptime & current time from memcached stats but there is still a difference of 60 seconds.

dormando:
There was a PR someone had last year trying to fix this :(

The tool should really use the new lru_crawler metadump command, which is able to print 0 exp time values correctly (displays them as -1).

If you're curious, you can telnet in from a recent version and run: `lru_crawler metadump 1` and it'll dump everything in slab class 1.

I'm going to leave this open for a bit.. maybe I should patch that or add a wiki document.

calin-iorgulescu:
I took a crack at updating `memcached-tool`'s `dump` command to use `lru_crawler`. However, I simply pull all key information using `lru_crawler metadump all`, instead of pulling each the keys for each slab separately.

Please see #361  .

dormando:
released in 1.5.7

PR #361, opened 2018-03-15
Rewrite memchached-tool 'dump' method to use new lru_crawler interface.

calin-iorgulescu:


calin-iorgulescu:
Should also solve #335 .

dormando:
nice. will take a look when I can. really regretting not adding a run limit to the lru_crawler command :(

dormando:
tabs in the file (see t/whitespace.t failure), devtools/clean-whitespace.pl can fix.

isn't dumping 0 exptime values, since regex is matching (\d+) and they're dumped with "-1".

slightly nervous since there's no cap to the amount of items being dumped, buffering the keys in memory is going to go poorly for people. might be okay, but perhaps a dump limit default/option would be prudent.

calin-iorgulescu:
Argh, missed the whitespaces, sorry about that. Changed the commit, whitespace tests pass now. I also fixed the regex to catch the `-`.

Would limiting the number of items fetched from `memcached-tool` suffice? Or are you considering parametrizing `lru_crawler metadump` s.t. only the first `limit` keys are returned?

dormando:
Yeah counting rows would suffice, but you have to disconnect and reconnect to stop the dump and start fetching stuff... I do need to actually parameterize lru_crawler commands but that's going to be painful.

I can accept it as is and can adjust again in the future though.

calin-iorgulescu:
Wrote a quick draft of what this might look like, feedback welcome. Please see f0e30ae.

dormando:
Looks fine to me. just add a comment for why the close is there (and to revisit in the future when the correct parameter appears)?

dormando:
was there anything in particular you are concerned about with your change? I just tested it and it works fine too

calin-iorgulescu:
Nothing in particular, thanks for looking it over! Added a comment as suggested, and also added a brief description to the `memcached-tool` manpage.

dormando:
rad. thanks for the docs update!

dormando:
released in 1.5.7

Issue #362, opened 2018-03-15
lru_crawler metadump misses last entry


calin-iorgulescu:
Steps to reproduce:

1. Load some random records (10 records here):
```
add usertable-user6284781860667377211 0 2147483647 1149
STORED
add usertable-user8517097267634966620 0 2147483647 1143
STORED
add usertable-user1820151046732198393 0 2147483647 1144
STORED
add usertable-user4052466453699787802 0 2147483647 1151
STORED
add usertable-user3232700585171816769 0 2147483647 1146
STORED
add usertable-user1000385178204227360 0 2147483647 1150
STORED
add usertable-user7697331399106995587 0 2147483647 1147
STORED
add usertable-user5465015992139406178 0 2147483647 1144
STORED
add usertable-user6873002678636213555 0 2147483647 1149
STORED
add usertable-user9105318085603802964 0 2147483647 1155
STORED
```
Confirm loading:
```
stats items
STAT items:13:number 10
STAT items:13:number_hot 0
STAT items:13:number_warm 0
STAT items:13:number_cold 10
STAT items:13:age_hot 0
STAT items:13:age_warm 0
STAT items:13:age 161
STAT items:13:evicted 0
STAT items:13:evicted_nonzero 0
STAT items:13:evicted_time 0
STAT items:13:outofmemory 0
STAT items:13:tailrepairs 0
STAT items:13:reclaimed 0
STAT items:13:expired_unfetched 0
STAT items:13:evicted_unfetched 0
STAT items:13:evicted_active 0
STAT items:13:crawler_reclaimed 0
STAT items:13:crawler_items_checked 20
STAT items:13:lrutail_reflocked 0
STAT items:13:moves_to_cold 10
STAT items:13:moves_to_warm 0
STAT items:13:moves_within_lru 0
STAT items:13:direct_reclaims 0
STAT items:13:hits_to_hot 0
STAT items:13:hits_to_warm 0
STAT items:13:hits_to_cold 10
STAT items:13:hits_to_temp 0
END
```
2. Try to use `lru_crawler` to get items (only _9_ are returned, last one is omitted):
```
lru_crawler metadump all
key=usertable-user6284781860667377211 exp=2147483647 la=1521142095 cas=1 fetch=yes cls=13 size=1241
key=usertable-user8517097267634966620 exp=2147483647 la=1521142095 cas=2 fetch=yes cls=13 size=1235
[...]

PR #363, opened 2018-03-15
Fix lru_crawler skipping over last entry in slab in item_crawler_thread.

calin-iorgulescu:
When using `lru_crawler metadump all` to dump cache contents, there is always one less key returned per slab.
Steps to reproduce and more details in issue #362 .

Problem seems to stem from a decrement before comparison in `item_crawler_thread`, on the following line:
https://github.com/memcached/memcached/blob/dbb7a8af90054bf4ef51f5814ef7ceb17d83d974/crawler.c#L389
When `crawlers[i].remaining` reaches 1, it gets decremented to 0, and `lru_crawler_class_done(i)` gets called, without calling `eval()` on the last item.

Please review and confirm.

### DIFF
--- a/crawler.c
+++ b/crawler.c
@@ -533,6 +533,14 @@ static int do_lru_crawler_start(uint32_t id, uint32_t remaining) {
         if (remaining == LRU_CRAWLER_CAP_REMAINING) {
             remaining = do_get_lru_size(sid);
         }
+        /* Values for remaining:
+         * remaining = 0
+         * - scan all elements, until a NULL is reached
+         * - if empty, NULL is reached right away
+         * remaining = n + 1
+         * - first n elements are parsed (or until a NULL is reached)
+         */
+        if (remaining) remaining++;
         crawlers[sid].remaining = remaining;
         crawlers[sid].slabs_clsid = sid;
         crawlers[sid].reclaimed = 0;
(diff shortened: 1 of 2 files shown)


dormando:
nice catch! can you retitle the PR so it describes the issue and mentions the issue?

calin-iorgulescu:
Done, and thank you for responding so quickly!

Frankly, I'm not sufficiently familiar with the code to figure out if the decrement before the check is necessary, so I'm not sure whether this doesn't break something else.

On the other hand, running `make test` shows all tests passed.

dormando:
Yeah I'll take a closer look at it before merge, no worries. Very few people other than me have ever looked at the crawler code :(

dormando:
yeah this one is tricky because of how I did the flags. if remaining == 0 at the start of that loop, it means run infinitely. The always-use-remaining thing was a recent change, and it might be a decent small change to just +1 in that code if you want to give it a shot.

calin-iorgulescu:
Ah, I think I see what you mean. If `remaining` is used to limit the number of slabs, and `remaining > nr_slabs` (i.e., `search != NULL` when `remaining == 1`), then it will decrement `remaining` to 0, and keep going.

Ok, I'll change the patch. I'll also try adding a test to expose this issue.

calin-iorgulescu:
Ok, amended the previous commit. Also added a test in `lru-crawler.t` to expose this problem. Please see 225641d.

dormando:
that looks better, thanks! I may or may not include the whitespace commit though.

calin-iorgulescu:
Sure, no problem. I can just take out that commit altogether, just let me know.

dormando:
cherry picked the off-by-one fix + test. thanks!

dormando:
released in 1.5.7

PR #367, opened 2018-03-24
Drop supplementary groups in addition to setgid

anthonyryan1:
When launching as root, we drop permissions to run as a specific user, however the way this is done fails to drop supplementary groups from the process.

On many systems, the root user is a member of many groups used to guard important system files. If we neglect to drop these groups it's still possible for a compromised memcached instance to access critical system files as though it were running with root level permissions.

On any given system `grep root /etc/group` will reveal all the groups root is a member of, and memcached will have access to any file these groups are authorized for, in spite of our attempts
to drop this access.

It's possible to test if a given memcached instance is affected by this and running with elevated permissions by checking the respective line in procfs: `grep Groups /proc/$pid/status`

If any groups are listed, memcached would have access to everything the listed groups have access to. After this patch no groups will be listed and memcached will be locked down properly.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -7536,6 +7536,10 @@ int main (int argc, char **argv) {
             fprintf(stderr, "can't find the user %s to switch to\n", username);
             exit(EX_NOUSER);
         }
+        if (setgroups(0, NULL) < 0) {
+            fprintf(stderr, "failed to drop supplementary groups\n");
+            exit(EX_OSERR);
+        }
         if (setgid(pw->pw_gid) < 0 || setuid(pw->pw_uid) < 0) {
             fprintf(stderr, "failed to assume identity of user %s\n", username);
             exit(EX_OSERR);
--- a/memcached.h
+++ b/memcached.h
@@ -18,6 +18,7 @@
 #include <pthread.h>
 #include <unistd.h>
 #include <assert.h>
+#include <grp.h>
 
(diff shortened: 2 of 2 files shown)


dormando:
weird. I've been doing this a long time and I'm not sure I've had a system with root in other groups (by default?). Ah well.

Need to check the portability of setgroups, otherwise I have no complaints.

thanks!

jsuter1:
This was originally noticed on Gentoo systems where the default behaviour is to have root in multiple groups. From a fresh stage3 install base:

```
# groups
root bin daemon sys adm disk wheel floppy tape video
```

For comparison I checked an Ubuntu 16.04 install in AWS and it does not have root in any other groups. Gentoo may be exposing this more directly but any distro is at risk without the use of setgroups.

anthonyryan1:
`make test` is reporting a success both with and without seccomp, it's also working as expected in my usage.

I can't comment authoritatively on the portability, but I'm not aware of any issues. I know some older BSDs used to have problems with setgroups being passed 0, but all the affected distros I was aware of no longer exist or fixed the problem more than a decade ago.

Let me know if I can be of any further assistance.

dormando:
Makes sense. pushed to 'next'. Seems fine locally, if it passes the buildbots things should be okay, if not I'll revise it.

dormando:
released in 1.5.7

PR #368, opened 2018-03-24
Bugfix: missing initialization code for extstore maint thread objects

orishu:
The extstore maintenance thread takes permanent ownership of its mutex.
Grabbing an uninitialized mutex can result in undefined behavior. If
the mutex uninitialized data happens to match an existing mutex, this
can cause a deadlock.


dormando:
Whoops, good catch! That structure is calloc'ed to zero so it's not going to cause problems, but better to be correct there.

orishu:
The code should actually be added *before* the pthread_create... sorry

PR #369, opened 2018-03-24
Bugfix: missing initialization code for extstore maint thread objects

orishu:
The extstore maintenance thread takes permanent ownership of its mutex.
Grabbing an uninitialized mutex can result in undefined behavior. In
this case the memory is zeroed so probably no harm.


dormando:
fwiw, you don't need to open a new PR, you can just push over your original branch and the PR updates.

dormando:
thanks! pushed in for the next release.

dormando:
released in 1.5.7

Issue #379, opened 2018-05-23
segfault when sasl enabled and using the java spymemcached client


PaulFurtado:
When using the java spymemcached client configured with invalid credentials, we're seeing segfaults, even with a single client and a single-threaded memcached server. This seems to mirror the experience of the users in this old issue: https://code.google.com/archive/p/memcached/issues/278

When using valid credentials in the client, we do not see any segfaults. We experience this up through memcached 1.5.7. We've tested on both CentOS 6 and Ubuntu 18.04.

When running under valgrind, it outputs:
```
==23422== Thread 6:
==23422== Invalid read of size 8
==23422==    at 0x528C89E: sasl_server_step (server.c:1430)
==23422==    by 0x1150B2: complete_nread (memcached.c:2146)
==23422==    by 0x119FD7: drive_machine (memcached.c:5567)
==23422==    by 0x4E40F8B: event_base_loop (event.c:1350)
==23422==    by 0x1227CA: worker_libevent (thread.c:385)
==23422==    by 0x549EAA0: start_thread (pthread_create.c:301)
==23422==    by 0xABBD6FF: ???
==23422==  Address 0x10 is not stack'd, malloc'd or (recently) free'd
==23422==
==23422==
==23422== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==23422==  Access not within mapped region at address 0x10
==23422==    at 0x528C89E: sasl_server_step (server.c:1430)
==23422==    by 0x1150B2: complete_nread (memcached.c:2146)
==23422==    by 0x119FD7: drive_machine (memcached.c:5567)
==23422==    by 0x4E40F8B: event_base_loop (event.c:1350)
==23422==    by 0x1227CA: worker_libevent (thread.c:385)
==23422==    by 0x549EAA0: start_thread (pthread_create.c:301)
==23422==    by 0xABBD6FF: ???
==23422==  If you believe this happened as a result of a stack
==23422==  overflow in your program's main thread (unlikely but
==23422==  possible), you can try to increase the size of the
==23422==  main thread stack using the --main-stacksize= flag.
==23422==  The main thread stack size used in this run was 8388608.
==23422==
==23422== HEAP SUMMARY:
==23422==     in use at exit: 3,395,838 bytes in 216 blocks
==23422==   total heap usage: 3,372 allocs, 3,156 frees, 14,194,509 bytes allocated
==23422==
==23422== LEAK SUMMARY:
==23422==    definitely lost: 0 bytes in 0 blocks
[...]

PR #383, opened 2018-05-30
Fix segfault: Prevent calling sasl_server_step before sasl_server_start

PaulFurtado:
Fixes: https://github.com/memcached/memcached/issues/379

If sasl_server_step is called on a sasl_conn which has not had sasl_server_start called on it, it will segfault from reading uninitialized memory.

Memcached currently calls sasl_server_start when the client sends the SASL_AUTH command and sasl_server_step when the client sends the SASL_STEP command. So if the client sends SASL_STEP before SASL_AUTH, the server segfaults.

For well-behaved clients, this case never happens; but for the java-memcached-client, when configured with an incorrect password, it happens very frequently. This is likely because the client handles auth on a background thread and the socket may be swapped out in the middle of authentication. You can see that code here: https://github.com/dustin/java-memcached-client/blob/master/src/main/java/net/spy/memcached/auth/AuthThread.java

The specific line that causes the segfault in libsasl is in sasl_server_step in server.c:
```c
    ret = s_conn->mech->m.plug->mech_step(conn->context,
					s_conn->sparams,
					clientin,
					clientinlen,
					serverout,
					serveroutlen,
					&conn->oparams);
```
it segfaults when doing the `s_conn->mech->m.plug` part with an error like `Cannot access memory at address 0x10`. This is because `s_conn->mech` is `NULL`.

---

I'm not sure that this is the best fix since it adds another field to the conn struct, but I could not find a way to cleanly determine if the sasl_conn had had sasl_server_start called on it using the existing data structures since `sasl_conn_t` is private so I can't just do `c->sasl_conn->mech == NULL`. Using the public libsasl API, I tried calling `sasl_getprop` like:
```c
const char *sasl_mech_name;
sasl_getprop(c->sasl_conn, SASL_MECHNAME, (const void **) &sasl_mech_name);
```
calling that from `process_bin_complete_sasl_auth`, `sasl_mech_name` ends up getting set to the value of `c->sfd`. Depending on where you call that from, you end up with a different piece of the conn struct, or a segfault. I'm almost wondering if libsasl should really have this guard since it seems fairly prone to exploitation.

I'm more than willing to modify this PR as necessary, but I wanted to post something to get the ball rolling.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -612,6 +612,7 @@ conn *conn_new(const int sfd, enum conn_states init_state,
     c->iovused = 0;
     c->msgcurr = 0;
     c->msgused = 0;
+    c->sasl_started = false;
     c->authenticated = false;
     c->last_cmd_time = current_time; /* initialize for idle kicker */
 #ifdef EXTSTORE
@@ -2106,8 +2107,16 @@ static void process_bin_complete_sasl_auth(conn *c) {
         result = sasl_server_start(c->sasl_conn, mech,
                                    challenge, vlen,
                                    &out, &outlen);
+        c->sasl_started = (result == SASL_OK || result == SASL_CONTINUE);
         break;
     case PROTOCOL_BINARY_CMD_SASL_STEP:
+        if (!c->sasl_started) {
+            if (settings.verbose) {
+                fprintf(stderr, "%d: SASL_STEP called but sasl_server_start "
+                        "not called for this connection!\n", c->sfd);
+            }
+            break;
+        }
         result = sasl_server_step(c->sasl_conn,
                                   challenge, vlen,
                                   &out, &outlen);
--- a/memcached.h
+++ b/memcached.h
@@ -555,6 +555,7 @@ typedef struct _io_wrap {
 struct conn {
     int    sfd;
     sasl_conn_t *sasl_conn;
+    bool sasl_started;
     bool authenticated;
     enum conn_states  state;
     enum bin_substates substate;
--- a/t/binary-sasl.t
+++ b/t/binary-sasl.t
@@ -192,7 +192,9 @@ for my $dir (split(/:/, $ENV{PATH}),
     }
 }
 
-system("echo testpass | $saslpasswd_path -a memcached -c -p testuser");
+my $sasl_realm = 'memcached.realm';
+
+system("echo testpass | $saslpasswd_path -a memcached -u $sasl_realm -c -p testuser");
 
 $mc = MC::Client->new;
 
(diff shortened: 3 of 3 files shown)


dormando:
excellent! I was going to poke at this today while doing benchmarks, but you just freed up a bunch of time :)

That does look like a problem, and a bool might be the right way to do it. You can run the `./sizes` util from the source directory to see if the connection structure actually gets any bigger or not. If not, or even if so, this is probably fine. I'll give it a more thorough review as soon as I can.

dormando:
(also; bonus points if you add a test under t/binary-sasl.t, but I can also do that since it's perl)

PaulFurtado:
> (also; bonus points if you add a test under t/binary-sasl.t, but I can also do that since it's perl)

@dormando do the sasl tests on master currently pass for you? I'm getting:
```
t/binary-sasl.t ............. ERROR: You cannot allow the ASCII protocol while using SASL.
t/binary-sasl.t ............. 1/33 ERROR: You cannot allow the ASCII protocol while using SASL.
t/binary-sasl.t ............. 2/33 ERROR: You cannot allow the ASCII protocol while using SASL.
t/binary-sasl.t ............. 3/33 
#   Failed test 'authenticated'
#   at t/binary-sasl.t line 206.
#          got: '32'
#     expected: '0'

#   Failed test at t/binary-sasl.t line 210.

#   Failed test ' = somevalue'
#   at t/binary-sasl.t line 107.
#          got: ''
#     expected: 'somevalue'
send: Cannot determine peer address at t/binary-sasl.t line 356.
# Looks like your test exited with 107 just after 19.
t/binary-sasl.t ............. Dubious, test returned 107 (wstat 27392, 0x6b00)
Failed 17/33 subtests 
```

PaulFurtado:
Interesting, yeah, they still fail for me on the next branch running on arch linux without my patch. I'll debug and see what I find. It might be picking up a system-wide sasl config and making the tests fail.

dormando:
haven't tried them recently on a variety of platforms yet, just ubuntu.

PaulFurtado:
The failure I was hitting was that the perl test program was not passing a realm when authenticating. I'm guessing that on ubuntu, libsasl may be automatically filling in the realm server-side, but that doesn't occur for me on centos or arch linux.

If you don't pass a realm to saslpasswd2, it defaults to the hostname, so by appending `@<my hostname>` to every user in the tests, I was able to get the tests to pass. To make the tests pass consistently on all platforms, I added a hardcoded realm in the test file and pass it to saslpassd2 with `-u` and append it to every username in the `authenticate` function.

I added that fix in a commit in this PR, let me know what you think

dormando:
that works, thanks!

PaulFurtado:
And also just added the test for the segfault. If you run the new test against master, it segfaults, but succeeds on this branch.

dormando:
test looks good. there's a spurious newline in it.

also the title of "Add test to ensure memcached" etc is too long (try to keep to 70 chars).

other than that, at a read things look right. hopefully I'll have time to actually poke it today

dormando:
Looks good; the auth error counter will still go up, there's a verbose output for inspecting why you get those counters, the tests look fine, etc.

Checked the connection structure byte size with a few different configure options and the extra bool doesn't change the size due to pre-existing padding. I just merged this all into next.

Thanks so much for taking care of it!

PaulFurtado:
> Thanks so much for taking care of it!

And thank you for all the guidance and quick responses, you were really helpful throughout the whole process!

dormando:
released in 1.5.9, thanks!

Issue #391, opened 2018-06-12
item client flag conversion function (code cleanup)


dormando:
```
        if (settings.inline_ascii_response) {
            flags = (uint32_t) strtoul(ITEM_suffix(it), (char **) NULL, 10);
        } else if (it->nsuffix > 0) {
            flags = *((uint32_t *)ITEM_suffix(it));
        } else {
            flags = 0;
        }
```

this snippit appears in several places; need a utility function to replace it (probably a macro)

PR #393, opened 2018-06-13
add utility macro to replace repetitive code snippit.

Linkerist:
As you noted in #391 , I just added an utility macro to replace repetitive code snippit.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1640,13 +1640,7 @@ static void process_bin_get_or_touch(conn *c) {
         rsp->message.header.response.cas = htonll(ITEM_get_cas(it));
 
         // add the flags
-        if (settings.inline_ascii_response) {
-            rsp->message.body.flags = htonl(strtoul(ITEM_suffix(it), NULL, 10));
-        } else if (it->nsuffix > 0) {
-            rsp->message.body.flags = htonl(*((uint32_t *)ITEM_suffix(it)));
-        } else {
-            rsp->message.body.flags = 0;
-        }
+        rsp->message.body.flags = htonl(FLAGS_CONV(settings.inline_ascii_response, it));
         add_iov(c, &rsp->message.body, sizeof(rsp->message.body));
 
         if (should_return_key) {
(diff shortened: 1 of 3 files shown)


dormando:
Excellent, thanks! Will double check and merge, I think/hope the tests cover it somewhat.

embarrassing it got up to four of them :) thought it was "only" three...

memcached.h
```diff
@@ -141,6 +141,19 @@
 #define APPEND_NUM_STAT(num, name, fmt, val) \
     APPEND_NUM_FMT_STAT("%d:%s", num, name, fmt, val)
 
+/** Item client flag conversion */
+#define FLAGS_CONV(iar, it) (__extension__ ({ \
```
dormando: instead of masking C errors with __extension__ and returning the flag value, could this instead take a pointer to flag and set it directly?

memcached.c
```diff
@@ -2858,15 +2852,7 @@ enum store_item_type do_store_item(item *it, int comm, conn *c, const uint32_t h
...
-
-                if (settings.inline_ascii_response) {
-                    flags = (uint32_t) strtoul(ITEM_suffix(old_it), (char **) NULL, 10);
-                } else if (old_it->nsuffix > 0) {
-                    flags = *((uint32_t *)ITEM_suffix(old_it));
-                } else {
-                    flags = 0;
-                }
-
+                flags = FLAGS_CONV(settings.inline_ascii_response, old_it);
```
Linkerist: Over here, your suggestion is obvious better choice.
we can use
```c
FLAGS_CONV(settings.inline_ascii_response, old_it, flags);
```
instead of
```c
flags = FLAGS_CONV(settings.inline_ascii_response, old_it);
```
to avoid ISO C forbids without extension.

memcached.c
```diff
@@ -1640,13 +1640,7 @@ static void process_bin_get_or_touch(conn *c) {
...
 
         // add the flags
-        if (settings.inline_ascii_response) {
-            rsp->message.body.flags = htonl(strtoul(ITEM_suffix(it), NULL, 10));
-        } else if (it->nsuffix > 0) {
-            rsp->message.body.flags = htonl(*((uint32_t *)ITEM_suffix(it)));
-        } else {
-            rsp->message.body.flags = 0;
-        }
+        rsp->message.body.flags = htonl(FLAGS_CONV(settings.inline_ascii_response, it));
```
Linkerist: But over here, we may take further dispose for flag, with htonl util.
Should we use this solution
```c
FLAGS_CONV(settings.inline_ascii_response, it, rsp->message.body.flags);
rsp->message.body.flags = htonl(rsp->message.body.flags);
```
instead of
```c
rsp->message.body.flags = htonl(FLAGS_CONV(settings.inline_ascii_response, it));
```
to adjust the flag value after updated FLAGS_CONV routine.
Or just use extension?

dormando: Yup, that's fine. Or use an explicitly named temporary variable. GCC will likely optimize it away.

I'd much prefer that than doing something forbidden by the C standard :)

dormando:
Looks good, thanks! I squashed it to one commit and merged for the next release.

dormando:
released in 1.5.9, thanks!

PR #395, opened 2018-06-16
Include keys with non-[\w.~-] bytes in memcached-tool dump

charmander:
`lru_crawler metadump` uses `uriencode`.


dormando:
nice catch, thanks!

dormando:
released in 1.5.9, thanks!

PR #397, opened 2018-06-17
t/issue_67: default UPD port is 0 (off)

VadimPushtaev:
`t/issue_67.t` failed for me. There are two problems:

* `udp port disabled` fails if I have a running memcached instance. Just like the `no arguments` test that was commented before. I fixed it with a naive check whether `11211` is already in use.
* `specifying tcp ephemeral port` works only because `validate_port` doesn't check whether a random port is defined. Current memcached version has `0` as a default UDP port, not `11211`.

### DIFF
--- a/t/issue_67.t
+++ b/t/issue_67.t
@@ -1,11 +1,12 @@
 #!/usr/bin/perl
 
 use strict;
-use Test::More tests => 22;
+use Test::More tests => 24;
 use FindBin qw($Bin);
 use lib "$Bin/lib";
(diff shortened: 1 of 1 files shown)


dormando:
thanks! I need to think on this a bit, I worry about tests that can be silently disabled.

dormando:
looks good, merged for the next release! Sorry for the lag, I was taking a break for a week.

dormando:
released in 1.5.9, thanks!

PR #398, opened 2018-06-18
Add .swp/.swo to .gitignore

VadimPushtaev:
`t/whitespace.t` fails for me because I have `.swo` and `.swp` files created by `vim` in the project root. I believe vim is pretty common and it can be a good idea to add them to `.gitignore`.

Also, we probably don't need to explicitly exempt files in the test that are mentioned in `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,4 +42,6 @@ doc/protocol-binary.txt
 /doc/doxy
 /memcached.spec
 .*.swp
+.swp
+.swo
 /compile_commands.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -42,4 +42,6 @@ doc/protocol-binary.txt
 /doc/doxy
 /memcached.spec
 .*.swp
+.swp
```
Linkerist: > A gitignore file specifies intentionally untracked files that Git should ignore.

The software itself should not care about what editor you are using.  
As you know, vim is indeed a common editor, but other people prefer emacs, nano, etc. It's difficult to define _common_.  
Obviously, we cannot add editors' ignored pattern here.

VadimPushtaev: Reasonable. I thought `.*.swp` was already there for the same reason.

PR #403, opened 2018-07-01
Mark seccomp experimental

viraptor:
To make it less likely that distros enable seccomp by default, mark it
as EXPERIMENTAL in both readme and configure help.


dormando:
looks good, thanks!

PR #408, opened 2018-07-08
Add a handler for seccomp crashes

viraptor:
When seccomp causes a crash, use a SIGSYS action and handle it to print
out an error. Most functions are not allowed at that point (no buffered
output, no ?printf functions, no abort, ...), so the implementation is
as minimal as possible.
Print out a message with the syscall number and exit the process (all
threads).

### DIFF
--- a/linux_priv.c
+++ b/linux_priv.c
@@ -3,10 +3,58 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <sys/ioctl.h>
+#include <signal.h>
+#include <string.h>
 #include "memcached.h"
 
+static char *kill_msg;
+// Make sure to preserve the ??? position in the string, or correct the offsets
+// in the syssig handler.
+#define KILL_MSG_STR "Seccomp policy failure. Caught syscall ???. This is " \
+    "either an exploit attempt, or your system is not supported yet.\n"
+
+static void handle_syssig(int signum, siginfo_t *si, void *thread_context) {
+#if defined(si_syscall)
+    int syscall_no = si->si_syscall;
+#else
+    // If system has no support for si_syscal, the information may not be
+    // precise.
+    int syscall_no = si->si_value.sival_int;
+#endif
+
+    // Replace the characters in the kill message with the syscall number. We
+    // can't safely printf (even write is not really valid, but we're crashing
+    // anyway).
+
+    kill_msg[39] = (syscall_no / 100) % 10 + '0';
+    kill_msg[40] = (syscall_no / 10) % 10 + '0';
+    kill_msg[41] = syscall_no % 10 + '0';
+    if (write(2, kill_msg, strlen(kill_msg)) == -1) {
+        // An error occurred, but we can't do anything about it here. This check
+        // is mostly to avoid the "ignoring return value of 'write'" error on
+        // distributions with broken gcc (no "ignore via cast to void" support).
+    }
+
+    // We can't use the nice exit() version because it causes at_exit handlers
+    // to be looked up and run. We can't take any locks while handling the
+    // signal, so _exit() is the only thing to do safely.
+    _exit(EXIT_FAILURE);
+}
+
+static const struct sigaction act = {
+    .sa_sigaction = handle_syssig,
+    .sa_flags = SA_SIGINFO,
+};
+
+void setup_privilege_violations_handler(void) {
+    kill_msg = malloc(strlen(KILL_MSG_STR)+1);
+    strcpy(kill_msg, KILL_MSG_STR);
+
+    sigaction(SIGSYS, &act, NULL);
+}
+
 // If anything crosses the policy, kill the process.
-#define DENY_ACTION SCMP_ACT_KILL
+#define DENY_ACTION SCMP_ACT_TRAP
 
 void drop_privileges(void) {
     scmp_filter_ctx ctx = seccomp_init(DENY_ACTION);
--- a/devtools/Dockerfile.alpine
+++ b/devtools/Dockerfile.alpine
@@ -1,6 +1,6 @@
 FROM alpine:latest
 
-RUN apk update && apk add --no-cache musl-dev libevent-dev libseccomp-dev linux-headers gcc make automake autoconf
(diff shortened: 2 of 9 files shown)


viraptor:
Concerns left:
- Do you have to support compilers which can't do struct assignments? (https://github.com/memcached/memcached/pull/408/files#diff-638e2e370b9ff7d3e12c4c2dbe2c0777R37)
- Is `_exit()` guaranteed to kill all threads? Or put another way - do you support systems with glibc <= 2.3? If yes, we may need an `exit_group()` fallback, but the last version like that was released in 2005, so... :man_shrugging: ?

dormando:
re: 1) I couldn't quickly google when that became supported. seems like c99?
2) hmm, similar in thinking to the first one:
...since this whole code is ifdef'ed and I believe the current syscall version of seccomp is from 2015ish, so long as the constructs aren't ultra-modern it should be fine. I want to update the codebase to be consistently c99 at some point anyway.

linux_priv.c
```diff
@@ -3,10 +3,51 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <sys/ioctl.h>
+#include <signal.h>
+#include <string.h>
 #include "memcached.h"
 
+static char *kill_msg;
+#define KILL_MSG_STR "Seccomp policy failure. Caught syscall ???. This is either an exploit attempt, or your system is not supported yet.\n"
```
dormando: split line?

dormando: also, add clear warning comment that the code below is directly indexing into it, please!

viraptor:
Comments addressed.

dormando:
think a blank setup_etc_handler needs to be added to openbsd_priv.c as well

dormando:
@viraptor bouncing back to you:

1) fails whitespace.t
2) compile error on write() call for unused return value (I just wrapped it with an if and had it exit either way, since we really don't care in this case..)
3) missing stub for openbsd_priv.c
4) t/misbehave.t fails:
```
$ prove t/misbehave.t 
t/misbehave.t .. Seccomp policy failure. Caught syscall 041. This is either an exploit attempt, or your system is not supported yet.
t/misbehave.t .. 1/1 
#   Failed test 'did not allow misbehaving'
#   at t/misbehave.t line 26.
#          got: ''
#     expected: undef
t/misbehave.t .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/1 subtests 

Test Summary Report
-------------------
t/misbehave.t (Wstat: 256 Tests: 1 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
Files=1, Tests=1,  2 wallclock secs ( 0.01 usr  0.00 sys +  0.02 cusr  0.00 csys =  0.03 CPU)
Result: FAIL
```

it prints the death message but doesn't seem to exit :( or maybe not in a way the test noticed, I didn't dig further.

thanks!

viraptor:
Interesting, thanks for double-checking! I'll have a look ~Sunday.

dormando:
@viraptor how're things? :)

dormando:
So what should I do about seccomp support? :( It's half broken and people keep enabling it. It's hard to find time to support it myself since it only breaks on bleeding edge builds.

viraptor:
Ok, let's get this done. I'll polish this as much as I can. Once it works fine on most distros, I don't think there will be much "maintenance" to do around it, since syscalls usage should change very rarely, but it's up to you to decide about available manpower.

Re. `fails whitespace.t`, could you tell me what the error is? It doesn't fail for me.

dormando:
Alright, thanks! Had to clear out the TLS work first. Can we get a clean rebase rather than merged?

What distro's have you tested on? I have the buildbots but it seems like Arch/alpine/etc are the most fiddly and I don't have builders for those.

dormando:
poke again. one day we'll sync up properly? :)

viraptor:
Updated and rebased.
I fixed up the alpine docker test (it now needs an explicit package for `prove`) and added fedora. That's pretty much how I test them right now and the reasoning here is:
- ubuntu / fedora - major 2 different base systems
- alpine - musl based repo
- arch - likely testing with most up-to-date packages

dormando:
Sigh, I completely missed your update. Will look.

dormando:
```
memcached.c:8232: undefined reference to `setup_privilege_violations_handler'
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1
```

looks like adding the worker drop privs to the obsd/fbsd files was wrong, but this one still fails. you added it to the solaris one. I don't have an obsd builder right now, just an fbsd 11. Tests do seem to pass with a trivial change... so I'll add it.

dormando:
Released in 1.5.17!

viraptor:
Thank you.

BTW, I saw you commenting about the sentiment on seccomp somewhere. I agree more and more with that recently. It's a lot of headache with different systems / libc-s handling syscalls differently, new kernels implementing/renaming them, etc. It was supposed to be the great solution to eliminating lots of security issues, but ended up... meh.

If we get another series of issues with practical usability of this implementation, I'd be happy to either restrict this protection to either a blacklist (like: prevent spawning new processes and opening new network connections), or scaling it back to non-seccomp things (prevent gaining gaining new privileges/capacities). Or maybe something else like including selinux/apparmor markers in threads (change context / change hat respectively), so systems/people who are interested can write their own policies instead. Linux side getting security protection would be great, but if it's currently causing more problems than it's solving - I'll be there to help rip it out :D

dormando:
Thanks for the note, sorry for the grumpiness. Think I'd be more okay with it if people weren't so vitriolic about it.. invariably those who complain seem to strictly believe it doesn't have any of these faults, making conversation impossible.

Usually I just slowly collect feedback about new features and we improve it over time, but this one has been a nuisance. Hopefully this is the last of it, though I've still not tested it with extstore or TLS :\

viraptor:
I vaguely remember making privileges XOR extstore work... or maybe I was just thinking about it? I think this case should either scream at you to reconsider or just not work, since it could lead to unexpected failure modes.

I'll test TLS later, but don't expect it to break.

PR #410, opened 2018-07-10
add missing va_end() call to logger_log()

mlichvar:


dormando:
thanks!

dormando:
in 1.5.10, thanks!

Issue #386, opened 2018-06-08
Fix signal handler


elfring:
[The variable “`caught`”](https://github.com/memcached/memcached/blob/dbb7a8af90054bf4ef51f5814ef7ceb17d83d974/timedrun.c#L10 "Update candidate") should [use the data type “`sig_atomic_t`”](https://www.securecoding.cert.org/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers).

VadimPushtaev:
This code is somehow naive anyway. For instance, if you send the second signal immediately after the first one, that first signal can be ignored due to races.

I'm not sure it is worth to be fixed since that is used for tests only.

***

On my machine, the problem can be reproduced like this. First, create a script to be run:

```perl
#!/usr/bin/env perl

use strict;
use warnings;

$SIG{INT} = sub { print "INT\n" };
$SIG{TERM} = sub { print "TERM\n" };

sleep 1 while 1;
```

Then run it with `timedrun`:

```
./timedrun 100 ./script.pl
```

Now send two different signals
```
perl -e '$x = shift; kill("TERM", $x) && kill("INT", $x)' 12345
```

If `timedrun` is not nice enough, it works fine:

```
$ ./timedrun 100 ./script.pl
TERM
TERM
```

But we can get a different result by simply lowing priority:

```
$ nice -n 10 ./timedrun 100 ./script.pl
INT
```

dormando:
Yeah the report is pedantic; this project gets a lot of static analyzer noise. I've marked this as "help wanted" since there's no real harm in fixing it, but anyone can do it :)

danielschemmel:
The field [settings.sig_hup](https://github.com/memcached/memcached/blob/1939cf9edc0656f7b1cfab6365f313b56b07a333/memcached.h#L401) should also be similarly using `sig_atomic_t`.

dormando:
please open new issues/pr's.

PR #413, opened 2018-07-27
Fix issue #386 (fix signal handler)

gleicon:
Short summary trying to follow my first tentative fix for issue #386: 

- Checked out a new branch from next
- Changed timedrun.c
- Configured the commit hooks
- Ran configure/make from zero and tested the daemon w/ memslap after the commit tests

### DIFF
--- a/timedrun.c
+++ b/timedrun.c
@@ -7,11 +7,11 @@
 
 #include <assert.h>
 
-static int caught = 0;
+volatile sig_atomic_t caught_sig = 0;
 
-static void caught_signal(int which)
+static void signal_handler(int which)
 {
-    caught = which;
+    caught_sig = which;
 }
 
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! timedrun.c turns into a binary that the test suite uses, so I'm not sure the memslap step was necessary for this run :)

fwiw, you can test timedrun specifically by looking at how it's used by the test suite and starting up a binary by hand then sending it signals. looks fine though.

gleicon:
Cool, thanks for the heads up. I guess I got used to run the same commands (including memslap) for anything memcached - which until this patch was just setting it up, creating cookbooks or testing a new client. Muscle memory :)

dormando:
in 1.5.10, thanks!

PR #415, opened 2018-08-03
add support for travis-ci

phantom9999:
".travis.yml" is the config for the services.
todo list: 
settings -> integrations & services -> add services -> select "travis ci"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+sudo: required
+dist: trusty
+language: cpp
+
+compiler:
+    - gcc
+
+install:
+    - sudo apt-get update -y
+    - sudo apt-get install -y build-essential automake1.11 autoconf libevent-dev libseccomp-dev git
+
+script:
+    - ./autogen.sh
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
+sudo: required
+dist: trusty
+language: cpp
```
dormando: I haven't looked at travis lately. I'm guessing there's no plain "c", if you've specified cpp here? memcached is pure C.

phantom9999: oh sorry, it's my mistake. actually, there is a plan whose language is c, but the builder can compile pure c project successfully, because it will choose gcc instead of g++ to compile the project in c.

dormando:
btw, you don't need to close and re-open a PR, you can force-push back to your branch after a squash and it'll auto-update.

I've added this in to try it. if it's weird or obnoxious I'll look at how to do it safely via the buildbot network I already use instead.

thanks!

dormando:
in master now, thanks!

PR #419, opened 2018-08-15
Remove some unused variables from main

nickfrostatx:
Since `l_socket` and `u_socket` are both checked and freed, the compiler didn't notice that they are otherwise unused.


dormando:
nice catch. looks like they've been dangling since 2008! it's merged for the next release.

PR #432, opened 2018-11-03
extstore: tiered storage

dormando:
earlier commits added the concept of free_buckets and were able to run
simple tests of tiered device storage.

this commit rewrites the way compaction is handled in extstore to be more efficient (write less), and enables concepts of tiered storage: items can be moved to different files/devices depending on their status.

run with: `ext_path=/path/f1:64m,ext_path=/path2/f2:128m:compact`
creates a 64m file for "default" pages and a 128m file for compacted
pages. Also works with 'chunked' and 'lowttl'

This also adds buckets `old` and `coldcompact`. IE:
`ext_path=/path/f1:64m,ext_path=/path2/f2:128m:old`

If `old` exists and we run out of disk space in the main buckets, items are moved from their original location to the device holding the `old` file. This is useful if you have a load/reload pattern where you may need the previous days's data to stick around, but it could be on a cheaper/slower device (like networked block storage)

`coldcompact` is similar to `old` but only takes items which reside in the COLD LRU. Both `old` and `coldcompact` can be used at the same time.

Other changes:

- fixes off-by-one preventing first page in each file from being used.
- page evictions are now done on-demand instead of async
- use a dedicated IO thread for writes and compaction work
- kick in compaction much later (1% of pages free by default, instead of 25%)
- check if compaction should run after writing occurs, rather than simply sleeping, to improve responsiveness.
- compact _the most fragmented pages first_ instead of simply the oldest.
- reduce CPU usage in various scenarios related to eviction/compaction.

- removes the "fragmentation slew" calculation from when compaction
should kick in. Instead of writing early, should wait until we've hit
the compact_under page limit then start compacting the emptiest pages. This may reduce total write amplification significantly depending on the workload.

### DIFF
--- a/extstore.c
+++ b/extstore.c
@@ -83,11 +83,10 @@ typedef struct {
     unsigned int depth; // queue depth
 } store_io_thread;
 
-typedef struct {
+// sub-struct for maintenance related tasks.
+struct store_maint {
     pthread_mutex_t mutex;
-    pthread_cond_t cond;
-    store_engine *e;
-} store_maint_thread;
+};
 
 struct store_engine {
     pthread_mutex_t mutex; /* covers internal stacks and variables */
@@ -95,8 +94,7 @@ struct store_engine {
     _store_wbuf *wbuf_stack; /* wbuf freelist */
     obj_io *io_stack; /* IO's to use with submitting wbuf's */
     store_io_thread *io_threads;
-    store_maint_thread *maint_thread;
-    store_page *page_freelist;
+    store_io_thread *bg_thread; /* dedicated thread for write submit / compact ops */
     store_page **page_buckets; /* stack of pages currently allocated to each bucket */
     store_page **free_page_buckets; /* stack of use-case isolated free pages */
     size_t page_size;
@@ -110,6 +108,7 @@ struct store_engine {
     unsigned int io_depth; /* FIXME: Might cache into thr struct */
     pthread_mutex_t stats_mutex;
     struct extstore_stats stats;
+    struct store_maint maint;
 };
 
 // FIXME: code is duplicated from thread.c since extstore.c doesn't pull in
@@ -164,9 +163,10 @@ static store_io_thread *_get_io_thread(store_engine *e) {
 static uint64_t _next_version(store_engine *e) {
     return e->version++;
 }
+// internal only method for freeing a page up
+static void _free_page(store_engine *e, store_page *p);
 
 static void *extstore_io_thread(void *arg);
-static void *extstore_maint_thread(void *arg);
 
 /* Copies stats internal to engine and computes any derived values */
 void extstore_get_stats(void *ptr, struct extstore_stats *st) {
@@ -194,10 +194,33 @@ void extstore_get_stats(void *ptr, struct extstore_stats *st) {
 
 void extstore_get_page_data(void *ptr, struct extstore_stats *st) {
     store_engine *e = (store_engine *)ptr;
-    STAT_L(e);
-    memcpy(st->page_data, e->stats.page_data,
-            sizeof(struct extstore_page_data) * e->page_count);
-    STAT_UL(e);
+    pthread_mutex_lock(&e->maint.mutex);
+    struct extstore_page_data *pd = st->page_data;
+
+    for (int i = 0; i < e->page_count; i++) {
+        store_page *p = &e->pages[i];
+        pthread_mutex_lock(&p->mutex);
+
+        pd[p->id].free_bucket = p->free_bucket;
+        pd[p->id].version = p->version;
+        pd[p->id].bytes_used = p->bytes_used;
+        if (p->active) {
+            pd[p->id].active = true;
+        }
+        if (p->active || p->free) {
+            pthread_mutex_unlock(&p->mutex);
+            continue;
+        }
+        if (p->obj_count > 0 && !p->closed) {
+            pd[p->id].bucket = p->bucket;
+        }
+        if ((p->obj_count == 0 || p->closed) && p->refcount == 0) {
+            _free_page(e, p);
+        }
+        pthread_mutex_unlock(&p->mutex);
+    }
+
+    pthread_mutex_unlock(&e->maint.mutex);
 }
 
 const char *extstore_err(enum extstore_res res) {
@@ -339,18 +362,13 @@ void *extstore_init(struct extstore_conf_file *fh, struct extstore_conf *cf,
 
     // free page buckets allows the app to organize devices by use case
     e->free_page_buckets = calloc(cf->page_buckets, sizeof(store_page *));
-    e->page_bucketcount = cf->page_buckets;
+    e->free_page_bucketcount = cf->page_buckets;
 
-    for (i = e->page_count-1; i > 0; i--) {
-        e->page_free++;
-        if (e->pages[i].free_bucket == 0) {
-            e->pages[i].next = e->page_freelist;
-            e->page_freelist = &e->pages[i];
-        } else {
-            int fb = e->pages[i].free_bucket;
-            e->pages[i].next = e->free_page_buckets[fb];
-            e->free_page_buckets[fb] = &e->pages[i];
-        }
+    e->page_free = e->page_count;
+    for (i = e->page_count-1; i >= 0; i--) {
+        int fb = e->pages[i].free_bucket;
+        e->pages[i].next = e->free_page_buckets[fb];
+        e->free_page_buckets[fb] = &e->pages[i];
     }
 
     // 0 is magic "page is freed" version
@@ -380,6 +398,7 @@ void *extstore_init(struct extstore_conf_file *fh, struct extstore_conf *cf,
 
     pthread_mutex_init(&e->mutex, NULL);
     pthread_mutex_init(&e->stats_mutex, NULL);
+    pthread_mutex_init(&e->maint.mutex, NULL);
 
     e->io_depth = cf->io_depth;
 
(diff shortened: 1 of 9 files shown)


dormando:
Resurrected the oldest open PR! :boom: 

Turns out this change looks super awesome. I rebased this sucker (which was nuts, four years of drift), and split off the `ext_max_sleep` change into its own commit and upstreamed that into `next`. I've audited the remaining change and found a few problems to fix:

eviction:
 - if DEFAULT is out of pages, it needs to evict
 - if at the same time another bucket is out of pages, the evicted page should be the oldest of the [default, otherbucket, etc] list.
 
- ensure error on start if no pages assigned to DEFAULT bucket
- add specific tests for tiering; might need stats counters for pages/usage in particular buckets.

Stuff to add:
- add COLDCOMPACT - wherein it would normally do a `drop_unread`, write those COLD_LRU item headers into the special bucket, but only if enabled
- maybe: OLDCOMPACT - add a command to demarcate old vs new via CAS id. If CAS is too old, direct compacted items into this bucket instead.

lidaohang:
mark

dormando:
another year and a half, still no merge!

I'd like to do this soon.

Just had a maybe great idea to add: Allow specifying a client flags mask to direct data into specific buckets.

- Requires using client flags, which adds 4 bytes of RAM per item: audit in case there's another place for this.
- Allow creating user-specified buckets that match to specific bits from client flags.
- Route data according to the flag, if it exists.

This would allow a lot of shenanigans, such as routing certain types of data into cheaper storage.

Would be nice to not have to rely on the client flags but at most I can wring one more byte out of the header and that's a break-glass-in-emergency byte.

storage.c
```diff
@@ -1090,10 +1169,10 @@ struct extstore_conf_file *storage_conf_parse(char *arg, unsigned int page_size)
     }
 
     // TODO: disabling until compact algorithm is improved.
-    if (cf->free_bucket != PAGE_BUCKET_DEFAULT) {
+    /*if (cf->free_bucket != PAGE_BUCKET_DEFAULT) {
         fprintf(stderr, "ext_path only presently supports the default bucket\n");
         goto error;
-    }
+    }*/
```
mrghosti3, line 1175: This could just be removed instead of commenting out since these lines are stored in git.

dormando:
rebased without any issue.

just finished re-reading the code and need to take another pass at the eviction/defrag routine. It looks like I did improve the algo last year, but now it can try to evict pages to make space in `DEFAULT` by removing them from buckets with dedicated storage, which can't be added back to `DEFAULT`, thus useless.

Still need to do the `compact_under` adjustment as well.

dormando:
Since the tiering code doesn't call the "maint thread" nearly as often, the compaction could easily get behind, especially if the system isn't writing for a long time and then suddenly starts writing again.

There are bugs around accounting the free pages and deciding when to evict vs compact that I will work on next.

dormando:
Had to redo everything again. This was trickier than expected :)

Next step is to write up a bunch of mcshredder scenarios and run them. The main comparison should be the total amount of bytes written to storage between upstream and this branch. This branch should be much less, if working as intended.

Need to also compare and ensure memory evictions don't happen, page evictions only happen when expect them, etc.

dormando:
Added a change which gives a dedicated IO thread to the flushing and compaction IO's. In basic testing this doesn't seem to do anything unless you've configured way too few IO threads.

Did a bunch of basic benchmark tests. Performance is... decent.

- If disk space is extremely constrained (> 95%) the amount of data written is similar between `next` and this branch.
- But with a little extra breathing room (~ 85-95%) the amount of writing is 30% less or more compared to `next`
- The compaction thread is much more aggressive in this branch and is able to keep up with higher write rates without evicting.

Need to see how to tune the tests so the write rate is realistic. Loading data as fast as possible overwhelms the flush thread no matter how I tune it (ie; removing the sleep entirely isn't enough), but this loads data at rates much faster than the network would allow.

Need a set of tests like:
- load at n gigabits per second, roughly
- with at least a few seconds of runway in memory and/or page data
... which should be both realistic and give the threads a chance to catch up.

A blunt sleep every N probably does it...

dormando:
Loading at n gigabits (roughly) was pretty easy to do so I did that real quick.

So long as the individual write batches aren't too huge it keeps up okay at 8 gigabits+ on this hardware, which I think is enough.

It seems like the write buffer size _does_ impact the performance here, but I think this fill rate is higher than realistic usage.

The flush thread goes to sleep while waiting for the single write buffer to flush before it can start writing into the buffer again. Another attempt at having A/B write buffers could have a real impact on flush performance, in batch write scenarios..

dormando:
Mcshredder freakin' rules. Did some more tuning and a few more larger test runs. Nand SSD's really hate reading and writing at the same time.

Feel like this change could be complete, but I feel we need one more mode: We already have a `drop_under` feature, which will allow compaction to drop items that haven't been read at all. Earlier I talked about a `coldcompact` feature, which would compact `COLD_LRU` items... but to make tiering make sense we'd add something like:

- `forcecompact`, where if this bucket has dedicated storage, we will compact all of the data from either the oldest _or_ most fragmented page when storage is low.
- This gives us a primary storage drive (DEFAULT/etc)
- Then older data gets rewritten wholesale onto the `forcecompact` device, instead of dropped.

Feels like this is needed for non-overwrite scenarios, as fragmentation won't increase but there is a clear separation between new and old data.

pkarumanchi9:
Were any of these tests run on HDD. When referring to tiered, one setup that might be efficient from cost perspective is hot data one one device(Faster and expensive storage), and cold data on another(Slower and cheap storage).

dormando:
@pkarumanchi9 Unfortunately extstore's data access pattern requires at least some kind of SSD. I was envisioning more like: recent/hot data on local SSD, and cold or old data (coldcompact or oldcompact) on cheaper or network-backed slower SSD.

The iops on HDD are like ~100 per disk, you'd need extremely cold data for that to function.

dormando:
had a false start with december for the forcecompat/oldcompat change. took a break for the end of dec and will try again as soon as I can.

the disk load was too high and the approach felt a bit wrong.

extstore.c
```diff
@@ -658,6 +704,18 @@ int extstore_submit(void *ptr, obj_io *io) {
...
 
+int extstore_submit(void *ptr, obj_io *io) {
+    store_engine *e = (store_engine *)ptr;
+    store_io_thread *t = _get_io_thread(e);
+    return _extstore_submit(ptr, io, t);
+}
+
+int extstore_submit_bg(void *ptr, obj_io *io) {
+    store_engine *e = (store_engine *)ptr;
+    store_io_thread *t = _get_io_thread(e);
```
dormando, line 715: wtf

dormando:
Got an in-progress change for OLD/COLDCOMPACT. will post when done. The logic is difficult.

If COLDCOMPACT or OLD are enabled:
- If either runs out of pages completely, the oldest is immediately evicted
- If normally compacting, and COLDCOMPACT exists, COLD LRU headers split into there instead of COMPACT bucket
- If there's nothing to defragment, the oldest page in any stream is sent to OLD

There's weird special casing to work through and the logic is fiddly but after stabbing at this all day I believe this is the right approach finally. The idea is if you turn these features on we will try to evict data out of specifically those buckets, then copy COLD/OLD data back into the freed up space.

dormando:
Latest attempt is much shorter/simpler. Need to add specific unit tests and try mcshredder again.

Change is officially feature complete; just needs final validation.

dormando:
looks like it works after a couple small bug fixes. fixed an issue with it over-defragmenting as well.

want to add a couple stats counters for good measure and then spend some mcshredder time with it, I think.

dormando:
kinda wish I could get someone to prod-QA-bake this change. overall it's a simplification but it's more complex of a change than I wanted.

I'll just have to chip at it with mcshredder. Probably do a test where some subset data gets fetched frequently and to blow up if it disappears, then run that for a while.

dormando:
Need to do a lot of testing to convince myself this is good to merge. The mcshredder tooling was useful during development so I should be able to expand on that a bit.

Basically run the same workload before/after and compare the total bytes read/written, pages created/evicted/etc, item evictions, average latency, latency outlier counts.

Workloads:
- Random cache pattern (kinda worst case)
- Power-law cache pattern (don't think I have zipf in mcshredder so maybe an approximation of some kind)
- Load/reload overwrite (at 50/70/90% disk util)
- Power-law continual load with COLDCOMPACT enabled
- Load/reload overwrite with OLD enabled (at 50/70/90% of main disk util)
- Full load, no deletes/expirations, evict from disk, constant slow add of data

Hopefully that's enough.

dormando:
got shredders mostly set up... added a special container to the test machine and been writing specific bootstrap code. not enough gas in the tank to get it actually running tests today though.

kinda hard to ensure the tests are covering realistic use cases...

dormando:
ended up doing a bunch more work on shredders. new changes now start properly and it can run a basic test fine. just pushed the new code to the public repo.

still working out how to write all of the tests I want and how to evaluate the results of the runs. glad I can reuse this work going forward...

dormando:
first set of comparison tests are back and ... well so far no regressions at least!

The timing clustering looks _much_ better in the newer code. I'm not sure how much of this is from the bgthread vs just writing less in general. not sure if I care to test separately to find out.

In general the new code writes much less to disk. When the disk is 50%-75% full the delta in writing is huge (50g written vs 66g during the test). As the disk usage grows toward 100% the different algorithms get closer, but the new one still beats it out and writes less.

Got one test running the OLD code and it worked.

I want to add a handful more workloads to validate but this feels pretty so far. mcshredder/shredders has been fanastic for getting this done in a repeatable way.

dormando:
feel like I want to try to get the compact checker to run less often before merge... not really sure how to do it.

it's a balance of latency vs checking too many times. getting a faster "quick check" so the thread can go back to sleep would reduce CPU usage at least and might be good enough for this release.

dormando:
Solved! went from 15% CPU to 0.7% (for the compact thread) while doing normal amounts of writes. This should be close to what I was doing before with just sleeping from the compact thread, except now a burst of writes will wake it up very quickly instead of having to wait for the sleep timeout.

dormando:
I think this is basically done. Might add a couple more shredders tests if I can think of them but I'm out of testing ideas and the performance is obviously better even without using the new features.

PR #433, opened 2018-11-04
FreeBSD superpages checking.

devnexen:
This is a knob existing from 7.0 (2008), can be only changed
at boot time. It is enabled by default, on usual archs at least,
but in some cases it might not be desired so we check it
whatsoever.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -55,6 +55,9 @@
 #include <getopt.h>
 #endif
 
+#if defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


devnexen:
@dormando any opinion on this one ? ;-)

dormando:
@devnexen sorry, I'm a month overdue for a PR/issue burndown :( it'll happen soonish, but probably not this week.

dormando:
out in 1.5.14

PR #436, opened 2018-11-20
correct comment in cache.h

hujiecs:
Hi, the comment above cache_destructor_t() seems to be wrong, fixed it


dormando:
out in 1.5.14

PR #440, opened 2018-12-05
Securing Memcached with TLS

tharanga:
This is an experimental implementation of TLS for memcached. 

Implementation details : https://gist.github.com/tharanga/9ffa856d4abb478150d02b9c1d9dd84d

To build with TLS
-----------------------
$ ./autogen.sh
$ ./configure --enable-tls
$ make -j

To test
---------
$ make test_tls

Current usage
-------------------
TLDR; -Z : enables SSL
(other options are omitted for brevity)
$ ./memcached -h

-l, --listen=<addr>       interface to listen on (default: INADDR_ANY)
                          if TLS/SSL is enabled, 'notls' prefix can be used to
                          disable for specific listeners (-l notls:<ip>:<port>) 

-Z, --enable-ssl          enable TLS/SSL

-o, --extended            comma separated list of extended options
                          most options have a 'no_' prefix to disable
   - ssl_chain_cert:      certificate chain file in PEM format
   - ssl_key:             private key if not in -ssl_chain_cert
   - ssl_verify_mode:    peer certificate verification mode, default is 0(None).
                          valid values are 0(None), 1(Request), 2(Require)
                          or 3(Once)
   - ssl_keyform:         private key format (PEM, DER or ENGINE) PEM default
   - ssl_cipher:          specify cipher list to be used
   - ssl_ca_cert:  PEM format file of acceptable client CA's
   - ssl_wbuf_size:       size in kilobytes of per-connection SSL output buffer

$ memcached -Z -o ssl_chain_cert=/home/user/cert -o -ssl_key=/home/user/pkey

To-do tasks
- [x] refactor network read/write and ensure existing unit tests are passing
- [x] experiment with OpenSSL API and Bufferevents
- [x] hack mc-crusher tool set to test performance
- [x] ensure existing unit tests pass with TLS
- [x] use standard SSL command line parameters
- [x] implement certificate refresh without restarting the process
- [x] make TLS support an optional feature
- [x] add TLS specific unit tests 

#441


dormando:
Thanks! Seems like a decent starting point. I'll poke at this as I have time. Remember there are side threads that can take temporary ownership of connections: logger.c, lru_crawler.c, etc.

tharanga:
> Thanks! Seems like a decent starting point. I'll poke at this as I have time. Remember there are side threads that can take temporary ownership of connections: logger.c, lru_crawler.c, etc.

Is it happening via conn_worker_readd? I will take a closer look and keep this in mind when adding unit tests. Thanks!

dormando:
there're a couple functions under thread.c for sending/receiving connections. I sadly haven't standardized code for managing the conns once they get to the side threads, you tend to see poll in use there.

tharanga:
@dormando this PR is ready for a detail review.

dormando:
@tharanga thanks! I'm working on getting time to look at this. it's quite a lot of effort to validate patches, sadly :(

tharanga:
Sure @dormando. This change-set is not so small. Thanks!

dormando:
Hey,

It's quite a big change for the amount of stuff it touches, but also relatively minimal for changing the guts. Good job on getting this started :)

Took a few hours to do a first read through, get it running locally and run tests/etc. Need to refresh on the SSL API and manually test things, but I have some high level stuff to start with:

 - I see you ticked off a note about mc-crusher modification. I don't see the branch anywhere, do you have it still? Otherwise I'll have to edit myself to test.
 - Any thoughts on a command for refreshing certs? Just having SIGUSR1 doesn't feel right.
 - Some code style doesn't match the rest of the project. I can line-by-line but you should make a first pass yourself. It's a lot closer than a lot of stuff I see though!
 - If --enable-tls is compiled in, at least a minimal amount of tests should run without requiring "make_tls". Otherwise this doubles the amount of time it'll take my buildbots to run through things.
 - The alloca/malloc/memcpy routine in ssl_sendmsg has too many edge cases.

Especially if fetching larger objects (even from multigets of medium sized objects), you'll end up constantly malloc/freeing memory of randomized sizes which will negate a lot of what we've strived for :) Also, if a server has a relatively small send buffer, you could end up memcpy'ing megabytes of stuff and then only writing 32-64k at a time.

I'm pretty sure it's safe to just use a static allocation on the worker threads. c->t->ssl_buf. Default of 8-16k but sized via an -o option. In testing a long time ago 64k seemed to be a sweet spot due to SSL framing, but that'll need to be re-tested. The static allocation should help attenuate both situations, and alloca() creeps me out anyway.

Also the initial handshake and re-negotiation stuff should be explicitly tested/understood. IIRC you'll get EAGAIN or something like a "needs to read" after SSL_write sometimes (and vice-versa, if you try to read during a negotiation). I'm not sure how the state machine handles that as-is. Another thing I have to refresh my memory on... There're some utility functions for finding the current TLS state of a connection. We should consider at minimum to do initial negotiation in the listening thread, probably a thread pool to throw things into when they need to re-negotiate as well, but can do that in the longer term since I'm working towards having more generic worker pools.

I'm hopeful after a few back-and-forths we can merge this with the compile flag and some EXPERIMENTAL marking. After a frontend refactor it'll be a lot easier to get the performance up and other features supported.

dormando:
Guess sidethreads as-is would make a shared buffer hard. c->thread doesn't change when sent to a side thread. It uses c->thread to redispatch back to the original worker.

Still, not many places use side threads so it should be possible to figure out.

tharanga:
Hey Dormando,

Thanks for checking this. 

- I didn't submit a PR for mc-crusher, only had local modifications, let me check that and send across.
- [ ] send the mc-crusher changes

- SIGUSR1 is not right. We discussed about a command, didn't see any issues with that, I will change it accordingly.
- [ ] use a command for certificate refresh

- I followed the existing coding style, let me do a close review and fix things.
- [ ] Fix code style violations

- I will add minimal tests when compiled with --enable-tls
- [ ] add minimal tests when compiled with TLS

- I will refactor ssl_sendmsg with your suggestions.
- [ ] refactor and fix ssl_sendmsg

- Yes, SSL state machine is rather complex, SSL_ERROR_WANT_WRITE and SSL_ERROR_WANT_READ can happen during ssl_read and ssl_write respectively. Even initial handshake doesn't happen right away as we are using non-blocking sockets. After a couple of iterations, I realized current state machine handles this, let me write-down my understanding so we can review it. I wanted to disable renegotiation for the first version, let me explicitly add tests there. I see your concern about worker threads doing the initial handshake, could negatively affect the performance of existing connections when new SSL connections are flowing in.
- [ ] write down SSL state machine
- [ ] do the initial handshake via the listening thread
- [ ] test renegotiation and disable if needed

regards
Tharanga

dormando:
Excellent, thanks!

In the past I would disable re-negotiation because it was often a security fault, but we should review carefully.

Code style is tough since we used to cross-accept patches with... whatever style people wanted. Some of the more recent files are just what I do, but we should pick a thing and stick with it from now on.

I've been fiddling with mc-crusher recently, so you probably have to port changes, sorry!

dormando:
Took another quick skim; I'll wait for some updates from you then go over it more closely all at once. Just letting you know to not wait on any more comments from me right now.

tharanga:
Sounds good. I will keep you posted. Thanks!

dormando:
Still looking through the other changes, it looks like we didn't sync properly on the c->ssl_wbuf bit. I'd said:

"I'm pretty sure it's safe to just use a static allocation on the worker threads. c->t->ssl_buf. Default of 8-16k but sized via an -o option. "

I mean c->thread I guess. So there's no reason to have the ssl_wbuf attached to every connection, since it's only used temporarily during an actual SSL write.. and on a worker thread only one write can be happening at a time. So we have 16k * the number of worker threads, rather than 16k * 50,000 connections or what have you. This helps keep things fitting in cpu cache as well, and is something I'll be expanding to the wbuf/rbuf's when I get to the frontend rewrite (though that's more complicated).

This is slightly complicated because of the sidethread business: c->thread doesn't get updated but writes can happen while a connection is on a side thread (lru_crawler metadump, watch, etc). I haven't thought about how to cope with this yet but can propose something if you don't get to it first :)

Thanks!

dormando:
maybe: keep c->ssl_wbuf pointer, but set it to c->thread->ssl_wbuf during dispatch and redispatch. when getting assigned to a side thread use a local one. unless I'm blind there're still only two active side threads: crawler and logger which are both single threaded. So should be okay to have a static allocation for those. Actually the logger doesn't even redispatch, it runs until connection close because I never added a parser to end/return :) Also extstore does a redispatch but it doesn't write to the socket itself. So that means we probably only have to address the crawler/logger.

I'll be writing a real set of side thread structs/etc before adding more side threads, so then a dispatch-to-sidethread function would properly manage that pointer.

dormando:
Tracking these:
https://github.com/awslabs/s2n/issues/1012
https://github.com/awslabs/s2n/issues/1013

I'd prefer to use s2n but probably better for a longer term change.

s2n It still has the SSL_write() interface issues. Which we might be able to overcome with some BIO hackery... but I don't understand that 100% yet and need to code dive. Asked them some questions in the issues above.. but not sure I'll get a timely response.

I prefer it because it lacks any sort of locking for the network code. With OpenSSL's network code you can end up slamming locks if you generate errors, etc. Global locks are just all over the code internally and severely limits the CPU scalability no matter what you do. Though I will be auditing this again once I get mc-crusher working; it's easy enough to see under a stack trace profiler.

However s2n also does seem like a tough dependency... now instead of OpenSSL you also need to pull in s2n which probably isn't a -dev lib anywhere. It's also un-released (no versions), ugh. Maybe it's not too bad complexity wise to make it conditional.

tharanga:
I missed the c->thread->ssl_wbuf memo, let me fix that as you suggested. In fact we internally reflect on the issue of having the buffer at the connection level as 50k connections are not rare in our environment.
+1 on the s2n idea. We briefly looked at both boringSSL (Google) and s2n (Amazon), but decided to stick with openssl as we were not sure about how community would accept that kind of a dependency.

dormando:
Ok, great!

Despite its marketing s2n doesn't replace openssl, it supplants the network frontend. It still requires the encryption engine and x509 parsing from opnssl (or boringssl), which means it's always two dependencies. so that's too bad.

...but I'll still do a comparison while this feature is experimental, maybe bumping the priority depending on how the mc-crusher tests go. openssl will scale to a couple cores okay, at least.

dormando:
Looks about right, thanks! Just a few TODO's left from the original notes?

tharanga:
I didn't allocate buffers in side threads as none of them use the sendmsg, I wish SSL has a writev.

I'm working on doing the initial handshake via the listening thread, it's straightforward except all ifdef's at method signatures (dispatch_conn_new and conn_new), which I like to avoid if possible.

I tested renegotiation manually but need to add automated tests.

Next is to write down the SSL state machine. Let me update the gist (https://gist.github.com/tharanga/9ffa856d4abb478150d02b9c1d9dd84d) so we can create a doc/tls.txt out of that.

dormando:
sounds good. make sure you get my attention if you want me to help with anything :) The s2n folks seem amenable to a writev, but I want to qualify and release this PR as-is before going down that path. I toyed around a bit with BIO_f_filter and BIO_f_buffer but either I don't really understand them or they're not that helpful in this situation. Don't think I ended up using them in the past either.

Over the weekend I reimplemented a flat buffer mode for mc-crusher. Ended up sidetracked yesterday but working on putting TLS stuff on top of this refactored work.

tharanga:
Thanks Dormando! +1 for mc-crusher refactoring and doing the s2n change as a follow-up work.
I moved SSL handshake to the listening thread, couldn't get rid of having ifdef's all over methods, didn't want to change the stateless nature of those methods. Let me know if you have a better suggestion.
It would be great if you can shed some light on automating SSL renegotiation tests.

dormando:
Cool, thanks! I'll give some thought to the ifdef soup here and get back to you.

What've you tried so far with renegotiation? You were manually triggering it? In the past I've simply disabled renegotiation. A number of older SSL vulnerabilities and DoS's had to do with renegotiation. Is it actually necessary to support here?

memcached.c
```diff
@@ -4926,6 +5012,13 @@ static void process_command(conn *c, char *command) {
 #ifdef EXTSTORE
     } else if (ntokens >= 3 && strcmp(tokens[COMMAND_TOKEN].value, "extstore") == 0) {
         process_extstore_command(c, tokens, ntokens);
+#endif
+#ifdef TLS
+    } else if (ntokens >= 1 && ntokens <= 3 && (strcmp(tokens[COMMAND_TOKEN].value, "refresh_certs") == 0)) {
+        set_noreply_maybe(c, tokens, ntokens);
+        refresh_certificates();
```
dormando: refresh_certificates() prints to stderr. Should create a string and return to the caller, which is then copied or freed.

tharanga:
I used "R" option with openssl client.
$openssl s_client -connect localhost:11211
> stats
> R

There's no need to support the renegotiation, in fact we prefer not to do so until there's a need. We can easily disable renegotiation with openssl 1.1.0, but the undocumented option doesn't work with 1.0.1, I had a hack to do it (register a callback and abort the ssl session) which I'm not fond of. This also brings up the question about the opensssl version, maybe we can bump up the lowest supported version to 1.1.0. What are your thoughts on this? I prefer this so we can also use the TLS_server_method instead of the deprecated SSLv23_server_method.

dormando:
I thought you folks were stuck on pre-1.1.0? I was just going over the code and noticing the deprecated stuff everywhere. I'd have to add #if OPENSSL_API_COMPAT in a bunch of places, or else when the deprecated code gets removed in newer versions we'd fail to compile.

Also the locking callback stuff is completely gone in 1.1.0, which would simplify things a bit. Since this is new and there're good reasons to avoid running > 2.5 year old versions of OpenSSL I'm actually okay with this. If someone makes a significant argument for the older stuff we can add it in as necessary.

... and yeah. The callback trick is what I've used in the past to disable it. It's not great, and if there's a 1.1.x bit that does it without the cb dance that's an extra bonus.

dormando:
Note (partly for myself): OpenSSL locking appears to be substantially improved in 1.1.0

IE: In SSL_CTX_up_ref(), 1.1.0+ are using atomic counters and localized locks. Prior to that, any modification on *any* CTX acquired a global lock (that used the callback system to find the specific lock).

SSL_free() (which can be called from any thread) internally calls SSL_CTX_free(), which pokes the CTX atomic counter/lock/etc.

Still should verify a few more functions, but this looks like if testing shows lock scaling issues, we can potentially scale more by having one CTX per worker thread. This may be necessary for users who don't use persistent connections, but I don't think troubles your use case specifically.

The locking is weird; internally it uses locks here and there, but externally we still need to lock around modifications to a CTX, SSL objects, etc. Like it was designed at the wrong layer. :)

dormando:
https://github.com/memcached/mc-crusher/pull/8 - after resting my brain a bit. Using "modern" OpenSSL 1.1.0 initialization. Much simpler than the older modes? No locking since connects all happen in the original thread.

Unfortunately I had to read a good chunk of the OpenSSL codebase to figure this out, heh. Most tutorials, even on the site, are for the old code.

Ran a few tests and it seems to perform okay in throughput. Need to work the patch a bit more and run latency tests/perf/etc on my larger machines.

configure.ac
```diff
@@ -354,6 +362,93 @@ if test $ac_cv_libevent_dir != "(system)"; then
   fi
 fi
 
+trylibssldir=""
```
dormando: Remember to add a minimum version test in here

crawler.c
```diff
@@ -293,7 +293,8 @@ static int lru_crawler_poll(crawler_client_t *c) {
 
     if (to_poll[0].revents & POLLIN) {
         char buf[1];
-        int res = read(c->sfd, buf, 1);
+        int res = c->c == NULL ? read(c->sfd, buf, 1) :
```
dormando: when would c->c be NULL?

logger.c
```diff
@@ -443,7 +443,8 @@ static int logger_thread_poll_watchers(int force_poll, int watcher) {
          */
         if (watchers_pollfds[nfd].revents & POLLIN) {
             char buf[1];
-            int res = read(w->sfd, buf, 1);
+            int res = w->c == NULL ? read(w->sfd, buf, 1) :
```
dormando: Same as crawler. when is this NULL?

memcached.c
```diff
@@ -5458,8 +5551,51 @@ static void drive_machine(conn *c) {
                 stats.rejected_conns++;
                 STATS_UNLOCK();
             } else {
+#ifdef TLS
+                SSL *ssl = NULL;
+                if (IS_TCP(c->transport) && settings.ssl_enabled) {
+                    // Setup the SSL object at each connection level when it's enabled
+                    // at the correct port if ssl_port is specified.
+                    int port = 0;
+                    if (settings.ssl_port > 0) {
```
dormando: Is this because SSL and non-SSL listeners can co-exist?

memcached.c
```diff
@@ -5972,12 +6108,20 @@ static int server_socket(const char *interface,
                 int per_thread_fd = c ? dup(sfd) : sfd;
                 dispatch_conn_new(per_thread_fd, conn_read,
                                   EV_READ | EV_PERSIST,
-                                  UDP_READ_BUFFER_SIZE, transport);
+                                  UDP_READ_BUFFER_SIZE, transport
+#ifdef TLS
+                                  , NULL
```
dormando: think instead of #ifdef soup we can always have the extra parameter here.

but inside the function, use an UNUSED() macro or actually use the TLS bits.

tls.c
```diff
@@ -0,0 +1,194 @@
...
+
+#ifdef TLS
+
+#include "tls.h"
+#include <string.h>
+#include <sysexits.h>
+
+static pthread_mutex_t ssl_ctx_lock = PTHREAD_MUTEX_INITIALIZER;
+
+#define MIN(a,b) (((a)<(b))?(a):(b))
```
dormando: drop the single-use macro

tls.c
```diff
@@ -0,0 +1,194 @@
...
+        memcpy((void*)bp, (void*)msg->msg_iov[i].iov_base, copy);
+        bp +=  copy;
+        to_copy -= copy;
+        if (to_copy == 0)
+            break;
+    }
+    /* TODO : document the state machine interactions for SSL_write with
+        non-blocking sockets/ SSL re-negotiations
+    */
+    return SSL_write(c->ssl, c->ssl_wbuf, bytes);
```
dormando: Is it really safe to never check SSL for SSL_write()?

tls.c
```diff
@@ -0,0 +1,194 @@
...
+        pthread_mutex_init(&(ssl_locks[i]), NULL);
+    }
+
+    CRYPTO_set_id_callback(get_thread_id_cb);
+    CRYPTO_set_locking_callback(thread_lock_cb);
+
+    // SSL context for the process. All connections will share one
+    // process level context. We should use TLS_server_method when we
+    // start using the version 1.1.0
+    settings.ssl_ctx = SSL_CTX_new (SSLv23_server_method());
```
dormando: style: funcname()

tls.c
```diff
@@ -0,0 +1,194 @@
...
+    // Clients should use at least TLSv1.2
+    int flags = SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 |
+                SSL_OP_NO_TLSv1 |SSL_OP_NO_TLSv1_1;
+    SSL_CTX_set_options(settings.ssl_ctx, flags);
+
+    // The sevrer certificate, private key and validations.
```
dormando: sevrer -> server

tls.c
```diff
@@ -0,0 +1,194 @@
...
+    if (!SSL_CTX_use_certificate_chain_file(settings.ssl_ctx,
```
dormando: could pull these out into a function and share with fresh_certs()

dormando:
Oh, took a note yesterday but forgot to add:

```
SSL_CTX_set_mode -> SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER
  Make it possible to retry SSL_write() with changed buffer location (the buffer contents must stay the same). This is not the
  default to avoid the misconception that non-blocking SSL_write() behaves like non-blocking write().
```

I got partway through trying to understand this; From reading the code it wasn't obvious which types of errors requires the input buffer pointer to be the same for a retry :\ Sometimes the docs don't match up with reality though and they might've fixed this, or the error is obscure. Your original sendmsg would fail in these scenarios (alloca or malloc), somewhat ironically the per-thread buffer scenario should work just fine... buf location will never change and it'll re-fill with the same data if there's a retry?

configure.ac
tharanga: I added the new TLS_server_method to the snippet below which fails when tried to build with any version prior to 1.1.0. I modified the error message accordingly. To test this, I ran a local docker with build (travis) commands. How should we go about the build?

Option 1. Update the base OS to bionic, install the standard packages which installs openssl 1.1.0g (March 2018)

Option 2. Keep the OS version as xenial (I saw the other request on this), download the openssl tarball (openssl-1.1.0j.tar.gz, Nov-2018), make and install it. With this option we have the freedom to control the version and go to even 1.1.1 if desired.

I verified both approaches locally with different docker images (bionic and xenial).

crawler.c
tharanga: Coming to think of it, I don't believe it can be.

logger.c
tharanga: Ditto.

memcached.c
tharanga: Yes. Tests are in t/ssl_ports.t

tls.c
tharanga: +1

memcached.c
dormando: This seems confusing as implemented, or would be hard to explain to a user, maybe. I'll think about it more but it wasn't obvious to me and I wrote half this code :)

tharanga: Let me refactor the code, probably the previous version was more clear, and add a better comment.

dormando: I mean from an end-user standpoint too.
./memcached -Z (--enable-ssl) -l 127.0.0.1:11211 -l 127.0.0.1:11212 -l 0.0.0.0:11311 -o ssl_port=11211

... means "secure any listener on just port 11211" - I'm not sure how someone would come to that conclusion by looking at the help output.

might be better to track tls-or-not on the c-> level and mark it in the listener object during setup.

I'm not sure I like this idea but something like: `-Z -l 0.0.0.0:11211 -l notls:127.0.0.1:11211

Then you're explicitly stating which listeners do what, and if TLS is enabled it's simply the default unless otherwise specified.

configure.ac
dormando: I'm fine with either; I actually don't really care about the travis thing. I use a large set of buildbots that actually covers different platforms for release testing. The travis integration is just to help spot check PR's because people wanted it and I can't easily integrate buildbot for that. So travis isn't used to qualify releases, at any rate.

configure.ac
```diff
@@ -405,7 +405,7 @@ if test "x$enable_tls" = "xyes"; then
       # Can I compile and link it?
       AC_TRY_LINK([#include <sys/time.h>
   #include <sys/types.h>
-  #include <openssl/ssl.h>], [ SSL_CTX* ssl_ctx = SSL_CTX_new (SSLv23_server_method()); ],
+  #include <openssl/ssl.h>], [ SSL_CTX* ssl_ctx = SSL_CTX_new(TLS_server_method()); ],
```
dormando: Is this what OPENSSL_API_COMPAT is for? though this is probably fine. if we want to bump up the version again we'd have to find different function though.

tharanga: Looks like OPENSSL_API_COMPAT is to hide deprecated APIs in new builds. Yes, for the current version bump, this validates both compile/link and the version, but ideally we should check the exact version with "$openssl version" or something alike.

dormando: looks like assert(OPENSSL_VERSION_NUMBER >= 0x10100000L) might be what we want?

tharanga: +1, let me check that.

tls.c
dormando: er, "Is it really safe to never check SSL specific errors for SSL_write()?"

I english good.

dormando:
Your handshake commit only does the initial part of the handshake in the listener thread, not the actual RSA calculation?

We can probably leave that as-is for now, since it could require some plumbing to fix.

tharanga:
Yes, I was thinking of blocking on SSL_accept using SSL_ERROR_WANT_READ or SSL_ERROR_WANT_WRITE with a timeout, but wasn't sure about other complications of doing so in the lister thread.

tls.c
tharanga: Not really if we implement the renegotiation. I manually triggered renegotiations at different parts of the state machine and found out the SSL_write is broken as it doesn't return SSL_ERROR_WANT_READ when renegotiation request is in the flight. I suspect this is probably due to the way we set socket flags at the memcached state machine level which interferes with the internal SSL state machine, but still looking through the code to understand how this happens.
This is ok for the moment as I'm disabling the renegotiation, but looks like a bigger change when we want to support it.

memcached.c
tharanga: So we implement this at the interface+port level, getting rid off the ssl_port. Let me give this a thought, sounds like this suggestion aligns with the original intention of this feature (do plain text over the localhost).

tharanga:
re: SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER, wonder how we advance the buffer pointer? My understanding was we should retry SSL_write with the same buffer pointer unless return value is > 0.

tls.c
dormando: Yeah lets leave renegotiation disabled. I don't think there's any benefit to it, especially considering these connections should be intra-datacenter and can reconnect quickly. Is that the only time we'd have errors though? I was looking at the code and there're a ton. I can go read more I guess.

dormando:
Yeah; don't block in the listen thread, it needs to be in the state machine. I'll give that a shot myself, so you can focus on all the other crap I tossed at you :)

re: SSL_MODE_ACCEPT - I have no idea! Have to read the sauce. I assume it'll just work given the buffers are more static now. I can look into this too if you want, just let me know.

I'm holding back a bit while you refactor things, then will deep dive into tests/etc. I did a line-by-line the other day and made my other comments.

dormando:
lets get this sucker merged this week :)

tharanga:
A timely suggestion :)
I addressed code review comments so far, including getting rid of the ssl_port option. Still need to update the gist, but need to understand what other protocol messages can cause SSL_ERROR_WANT_READ/SSL_ERROR_WANT_WRITE after a successful handshake (except renegotiation).

tls.c
tharanga: Not sure about other scenarios except renegotiation, let me check this further.

tls.c
tharanga: I kept the macro as it's used in two places, but dropping this would be also fine if we want to stay away from macros.

memcached.c
```diff
@@ -5016,8 +5016,11 @@ static void process_command(conn *c, char *command) {
 #ifdef TLS
     } else if (ntokens >= 1 && ntokens <= 3 && (strcmp(tokens[COMMAND_TOKEN].value, "refresh_certs") == 0)) {
         set_noreply_maybe(c, tokens, ntokens);
-        refresh_certificates();
-        out_string(c, "OK");
+        if (load_server_certificates()) {
```
dormando: closer! What I was asking for was that load_server_certificates() returns a string (then use write_and_free here or free() in the initializer), so the user isn't mystified by what caused the error over the protocol.

memcached.c
```diff
@@ -5557,40 +5558,31 @@ static void drive_machine(conn *c) {
...
-                    // at the correct port if ssl_port is specified.
-                    int port = 0;
-                    if (settings.ssl_port > 0) {
-                        struct sockaddr_in peer_addr;
-                        socklen_t peer_addr_size = sizeof(peer_addr);
-                        getsockname(sfd, (struct sockaddr *)&peer_addr, &peer_addr_size);
-                        port = ntohs(peer_addr.sin_port);
+                if (IS_TCP(c->transport) &&
+                    settings.ssl_enabled &&
+                    c->ssl_enabled) {
```
dormando: can the first two conditions ever be false if c->ssl_enabled is true?

memcached.c
```diff
@@ -5557,40 +5558,31 @@ static void drive_machine(conn *c) {
...
+                   if (settings.ssl_ctx == NULL) {
+                        fprintf(stderr, "SSL context is not initialized\n");
```
dormando: needs an "if verbose" check, along with the other errors below.

memcached.c
```diff
@@ -6343,6 +6359,10 @@ static void usage(void) {
            "-A, --enable-shutdown     enable ascii \"shutdown\" command\n"
            "-a, --unix-mask=<mask>    access mask for UNIX socket, in octal (default: 0700)\n"
            "-l, --listen=<addr>       interface to listen on (default: INADDR_ANY)\n"
+#ifdef TLS
+           "                          if TLS/SSL is enabled, 'notls' prefix can be used to exclude ports\n"
```
dormando: "can be used to disable for specific listeners". doesn't have anything to do with ports anymore :)

memcached.c
tharanga: +1, I used the out_string and stack as a return string cannot differentiate between success and a malloc failure without further hand holding.

memcached.c
tharanga: They can't be, they in fact invariants, added as asserts.

memcached.c
tharanga: +1

memcached.c
tharanga: +1

dormando:
Cool! looks like you have a couple things left before I can go nuts?

The handshake state machine docs, couple more comments from the latest round, and some research into SSL_write/read error handling? There're a few small nits but I'll adjust those myself.

Anything you want me to do in specific?

tharanga:
I just pushed some changes addressing your recent comments. 
I'm working on the state machine docs and checking openssl code about SSL_write/read error handling.
I ran valgrind with and without TLS and didn't see any differences, though I may have not exhausted all paths (just ran mc-crusher for some time).
Shashi had looked at flame graphs in the past and saw crypto lib stands out in the CPU usage, a good thing to check again.
I believe you plan to benchmark both latency and throughput. One important thing notably absent from my tests is to test on other OS platforms that are common to both memcached and openssl.

dormando:
Looks like you missed a few of the "if (verbose) fixes" in the accept path? or haven't done it yet.

I'll be walking through the unit tests and doing a performance review using the mc-crusher changes I did, once you confirm you're done changing the code.

I'll also try building it on my various buildbots, ARM, etc.

tharanga:
My bad, just fixed them. Let me know if I missed anything or you can go nuts while I work on docs/SSL errors.

dormando:
alright, will jump into it tomorrow.

tharanga:
I updated the gist with details on how/when the SSL handshake happens.
Looking further into the SSL_write, I see 3 possible cases where it should return (but doesn't return with the current code) SSL_ERROR_WANT_READ. Handshake message, change cipher message and an alert. AFAIK change cipher message only happens with the handshake/renegotiation, so we should be good there. However, an alert could happen during a normal application data transfer as well.
The result of the current code is anything other than the application data after an handshake causes client to fail and disconnect iif client's state machine is not expecting application data.

dormando:
great, thanks! I've had a bit of a scheduling snafu but will wrap this up asap.

dormando:
https://github.com/dormando/memcached/tree/ssl_prototype_d

made a few changes here, otherwise it's down to testing and "release early and often". I posted a link to this PR in the mailing list as well; hoping to get some feedback on usability.

ended up undoing the out_string() change. In reality malloc never fails except for extremely exotic situations, and I'd rather have tls.c not interface to the protocol. Also snprintf'ed for safety.

rest are nits + removing one loop from ssl_sendmsg.

some quick perf tests seemed okay. I'm just going to accept that there will be bumps until I think about the handshaking more later. So I'll do some more testing there, as well as an ARM build manually. Unfortunately most of my buildbots are too old to actually run this without manually building openssl :)

@tharanga are you intending that the gist gets added as doc/tls.txt? Aside from that + a wiki page it's releasable (unless testing today finds something).

dormando:
Also, I'm going to end up squashing this all down for release. a lot of the commits aren't very descriptive or are non-working/missing things. This makes it harder to bisect bug reports.

dormando:
really sucks testing on a raspberry pi3. got it to compile and run help but it segfaults in tests.

going to try again on an ubuntu 18.04, as the ARM OS I just tried was debian 8 and I can just declare that as "too old" for this TLS stuff.

dormando:
note: apparently if you don't have IO::Socket::SSL installed the perl tests all bail, even when running non-tls tests. should attempt to fix.

ubuntu 18.04 64bit ARM builds and is passing tests so far (but it's super slow). I didn't have to compile openssl this time, either!

my 32bit ARM install is too old. I don't think it matters much anymore.

tharanga:
+1 for the changes you made. Squashing make sense. I just added the gist into the doc. How do we go about a wiki page? Shall draft one? 
re: perl package for testing, something I attempted to fix at the beginning but couldn't find a clean way to doing so. It's inconvenient for non TLS users to install the perl lib.

dormando:
I'll take a whack at the perl thing later.

go ahead and draft a wiki page if you want, you can gist it and I'll copy/paste it... sadly no way to give edit rights to the wiki without giving admin to the whole damn repo (maybe, i haven't checked lately). Just be explicit, clear, and terse/simple. Busy and tired people just want to know how to make it work and move on.

ARM tests did finish, hooray.

dormando:
spent some time with perf testing machines. I set up latency sampler running over localhost without tls, to measure the impact of TLS related requests more clearly.

it's trivial to get some requests to go over the 1ms mark (sometimes up to 10ms) by running just one perl script in a connect/disconnect loop.

with mc-crusher, there's a mystery causing the performance to waver constantly. With notls I get a nice consistent load, but with TLS it seems to waver around a lot... from 50krps to 3m and all along the middle. If I crank up the stats checking rate I can see it dropping to 0 for short periods of time.

Haven't quite caught what's going on, so I'm not sure if it's mc-crusher or memcached causing the fault. Tried gdb sampling both... under top mc-crusher's CPU is consistent but memcached's wavers. In the mc-crusher code I've forced the SSL negotiation to happen before the benchmark actually starts so it shouldn't be a state machine problem?

Don't _think_ I'll block the release on this. planning on fiddling with the perl stuff tomorrow then cutting a release noting it's pretty experimental. I'd feel a lot better with a stable benchmark though.

tharanga:
Looks like we need to do the initial handshake (and renegotiation in the future) as you suggested, in a different thread pool.
I saw your mc-crusher changes, yes, it can't be a state machine problem. What's the platform you used for testing? Which mc-crusher config was used (get/set operations, workers and payload sizes)? Unstable throughout is bit disturbing. I wonder whether this has anything to do with the size of SSL frames, SSL_read won't return the data until a complete frame is received, though the socket become ready as soon as a TCP packet arrives, we'll see EAGAIN with SSL_ERROR_WANT_READ.

dormando:
ubuntu 18.04.
`send=ascii_get,recv=blind_read,conns=50,key_prefix=foobar,key_prealloc=0,pipelines=8,key_count=10000000,live_rand=1,thread=2`
`./bench-warmer 127.0.0.1:11212 foobar 256 10000000`
`./memcached --enable-ssl -o ssl_chain_cert=t/server_crt.pem -o ssl_key=t/server_key.pem -p 11212 -m 4096 -l notls:127.0.0.1:11212 -l 127.0.0.1:11211 -t 16`

varied the thread= a few ways, same results. top end is higher.

printed all < 0 results from SSL_read/SSL_write in mc-crusher and no errors are happening. just rare EAGAIN's on write with WANT_WRITE, but they don't match up with the wobbles.

checked that it's not using/blocking on system entropy for some reason (just when mc starts, it seems).

I'll fiddle a bit more before giving up. mc-crusher's CPU is still static while mc wavers. Config is using a small number of pipelined gets, so records should be pretty small on both sides. Test is over localhost as well.

dormando:
hmmm... it stabilizes if I give crusher a lot of threads (8-16), but it's still relatively low performance (1-3m).

no errors anywhere on the benchmark side. will instrument mc I guess.

dormando:
just pushed a "fix".

crusher's readback code stops if it gets a short read (~2k), to avoid making extra syscalls with 0 byte results. With this benchmark SSL_read() always returns with ~290 bytes, the size of the TLS frame being sent back by memcached. Meaning it was slowly peeling off data and memcached would occasionally wake up to fulfill some prior EAGAIN's.

Kind of sucks. There's plenty of data in the buffer :\ Forgot that's how it works, but it's clearly stated in the manpage. The underlying BIO probably buffers the data so calling SSL_read() in a loop there won't make extra syscalls at least.

tharanga:
Awesome! This aligns with what I saw on the server side.
Server doesn't print SSL_read/SSL_write error messages for EAGAIN, rather treats that as a soft error and re-tries. I added logs around there and found no EAGAINs on the SSL_read (after the initial handshake is complete), but so many on the SSL_write with SSL error code SSL_ERROR_WANT_WRITE. This piles up as time goes on.

dormando:
Yeah.. I got it to spit back EAGAIN with some effort. oh well.

working on the perl tests fix now, then I'll burn down and release. were you planning on drafting a wiki page or should I do that?

tharanga:
Kudos for fixing perl tests.
I shared a draft version of the wiki. Feel free to change directly/ suggest.
There are some last minute edits to doc/tls.txt I will be pushing in soon.

dormando:
ok, just pushed a fix for the perl tests into my ssl_prototype_d branch. Since it was just a few calls some string eval works fine. If it were a lot of code I'd have made two internal modules with the connects abstracted and conditionally swapped them in. Now runs fine with/without the module installed. Dies loudly and specifically if you do ask for tests but the module is missing.

I'll do the release work as soon as you push the doc/tls.txt edits. It's been a few months :( sad to get out of the monthly habit for so long.

tharanga:
Done. Were you able to access the wiki page I shared?

dormando:
Squashed and merged, congratulations! I hope we didn't break everything :)

PR #442, opened 2018-12-07
Metaget support

anatale:
Adding support for metaget & metaget stats

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3126,8 +3126,10 @@ static void server_stats(ADD_STAT add_stats, conn *c) {
     APPEND_STAT("cmd_set", "%llu", (unsigned long long)slab_stats.set_cmds);
     APPEND_STAT("cmd_flush", "%llu", (unsigned long long)thread_stats.flush_cmds);
     APPEND_STAT("cmd_touch", "%llu", (unsigned long long)thread_stats.touch_cmds);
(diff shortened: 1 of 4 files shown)


anatale:
Hi @dormando, do you have any feedback for this PR?  Thanks Alex

tharanga:
+1 @anatale. This is a useful feature for one of our use cases. Can we extends this to include more metadata such as recent access time, size of data (very useful) and flags? May be you can create an issue and I can add more details. 
It's great to have the multi-key mget support, but that could be a follow-up PR.

dormando:
Yeah, definitely not implemented this way. I'm thinking it through:

1) ascii command, but using wbuf and faster printouts. should use a format like lru_crawler metadump, but also maybe a binary gettable version from ascii.
2) binprot, using extra fields and keeping the extras in binary.

These aren't too hard but I've always had a huge fear of them since people will end up relying on internal information. Also, if we don't provide optimized commands people will shoot themselves in the foot. basically a couple problems:

- if the internals ever change, these commands probably need to, so the format should be flexible or extensible.
- which means it probably can't be fast. if it can't be fast, I'll wake up one day with some huge user who does 100% metagets and is irate.

I'll maybe track this in a separate issue.

dormando:
closing in favor of: https://github.com/memcached/memcached/pull/484

Issue #447, opened 2019-01-04
memcached-tool dump loses exp value


johnleslie:
#395 fixes (uri-decodes) "key", but the s// call redefines $2 ("exp"), which is used later. Need to store $2 before calling the regex.

Maybe something like

```
        # key=foo exp=2147483647 la=1521046038 cas=717111 fetch=no cls=13 size=1232
        if (/^key=(\S+) exp=(-?\d+) .*/) {
            my ($k, $exp) = ($1, $2);
            # From https://metacpan.org/pod/URI::Escape
            $k =~ s/%([0-9A-Fa-f]{2})/chr(hex($1))/eg;
            if ($exp == -1) {
                $keyexp{$k} = 0;
            } else {
                $keyexp{$k} = $exp;
            }
        }
```

dormando:
ahh thanks! I'll take a look as soon as I can.

johnleslie:
Note: this is fixed in #448 (not merged)

dormando:
Sorry, I was away from the project for a while. I should be doing a merge burn-down this month.

dormando:
out in 1.5.14

PR #448, opened 2019-01-04
memcached-tool dump: regex call unescaping key reassigns $2, losing exp value. store/reference it as $exp

johnleslie:
See: #447


dormando:
out in 1.5.14

Issue #434, opened 2018-11-16
assoc.c error


vaneyli:
typedef  unsigned long  int  ub4;   /* unsigned 4-byte quantities */

on 64-bit machine the sizeof(unsigned long  int) is 8。

Fix:
typedef  uint32_t ub4;

morteryler:
I tried to suggest a fix https://github.com/memcached/memcached/compare/master...morteryler:patch-1

dormando:
apparently this doesn't actually matter (sorry I didn't ping the issue earlier).  It should probably just be hard-set to a uint32_t for code clarity, but if it's 8 bytes it doesn't actually affect anything.

dormando:
out in 1.5.14

PR #450, opened 2019-01-09
Fix error

morteryler:
Cross-platform bug fix for issue #434


dormando:
This is actually fine; even if it's 64-bit there's no effect to the code. It is confusing to read so I've pinned it to a uint32_t. No need for the conditionals. I've pushed my own fix, thanks for the report!

dormando:
out in 1.5.14

PR #453, opened 2019-01-29
Add optional feature support to RPM package building

mlmitch:
Some optional features can now be enabled for RPM building using the
rpmbuild conditional build parameters (--with and --without)

For example, sasl can be enabled and code coverage disabled in the RPM with:
rpmbuild --with sasl --without coverage -tb memcached-1.5.12.tar.gz

Note: rpmbuild doesn't allow dashes (`-`) in the parameter names. 
This is why `arm_crc32` `sasl_pwdb` and `option_checking` use underscores

Also, I was unable to get straightforward parameters working for optiions
with both enable and disable versions for configure. 
e.g. `--enable-dependency-tracking` and `--disable-dependency-tracking`

### DIFF
--- a/memcached.spec.in
+++ b/memcached.spec.in
@@ -1,3 +1,14 @@
+%bcond_with arm_crc32
+%bcond_with extstore
+%bcond_with seccomp
+%bcond_with sasl
+%bcond_with sasl_pwdb
+%bcond_with dtrace
+%bcond_with 64bit
+%bcond_without option_checking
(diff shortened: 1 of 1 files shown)


mlmitch:
Made this PR so creating a SASL enabled RPM is easier.
Using this script in the meantime:
https://gist.github.com/mlmitch/4408ad6875df589520c487f0ad658fe0

dormando:
out in 1.5.14

Issue #469, opened 2019-04-04
Issue with incr and delete running in parallel


aattuluri:
We have observed an issue with incr and delete of a key running in parallel and delete does not work (fails silently, gets a success response but the key is not deleted). As a result the counter doesn't get reset causing the counter to go up which results in an application issue. I have attached the code that can be used to reproduce. We used elastic cache cluster with 2 nodes to reproduce this issue.

Memcached server: 1.4.34 and 1.5.10 (happens on both versions)
Client: Spy memcached (elastic java cluster client) - com.amazonaws:elasticache-java-cluster-client:1.1.1

**Prerequisites to run**
Spy memcached: com.amazonaws:elasticache-java-cluster-client:1.1.1
Log4: org.apache.logging.log4j:log4j-core:2.8.2

**Steps to run**
i) Check REPLACE_ME and use the right memcached servers/cluster address and run the main function
ii) By default delete is used to reset the counter, so you should observe the message:
"after delete exceeds before value" when the issue occurs (one or twice every 10 minutes).
iii) Now flip isDelete variable to false, and rerun the test, this will use set operation instead of delete
iv) No occurrences of "after delete exceeds before value" will be found

```
package com.intuit.platform.gateway;

import net.spy.memcached.*;
import net.spy.memcached.ops.OperationStatus;
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;

import java.net.InetSocketAddress;
import java.util.*;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.atomic.AtomicReference;

/**
 * Created by aattuluri
 */

public class MemcachedCounter {

    public static Integer MEMCACHE_BUFFER = 1;

    public static final String THROTTLING_LOG_TEMPLATE = "logType=memcachedCounter, processTime=%d, counter=%d, key='%s', message='%s'";
    public static final String THROTTLING_LOG_DELETE_VALUE_TEMPLATE = "logType=memcachedCounter, processTime=%d, oldValue=%d, newValue=%d, key='%s', message='%s'";

    private final Logger logger = LogManager.getLogger(MemcachedCounter.class.getCanonicalName());

    private String SAFE_KEY = "IntuitMemcachedCounter";

    private Long INCR_VAL = 8L;

    private Long INCR_INTERVAL = 75L;

    private Long MIN_INCREMENTS = 1000 / INCR_INTERVAL;

    //if running locally change this to a lower value like 10
[...]

PR #470, opened 2019-04-04
close delete + incr item survival bug

dormando:
re #469 - delete actually locks/unlocks the item (and hashes the
key!) three times. Inbetween fetch and unlink, a fully locked
add_delta() can run, deleting the underlying item. DELETE then returns
success despite the original object hopscotching over it.

I really need to get to the frontend rewrite soon :(

This commit hasn't been fully audited for deadlocks on the stats
counters or extstore STORAGE_delete() function, but it passes tests.


dormando:
out in 1.5.14

PR #484, opened 2019-04-30
new flexible ascii meta commands (mget, mset, mdelete)

dormando:
Edit: overhauled top comment to reflect development work.

- we get asked a lot to provide a "metaget" command, for various uses, as well as random one-off commands.
- I really hate this and have been wanting to experiment with tuning of how get commands work for a long time.

Assuming that if I offer a metaget command which gives people the information they're curious about in an inefficient format, plus data they don't need, we'll just end up with a slow command with compatibility issues. No matter how you wrap warnings in documentation, people will put it into production under high load. Then I'm stuck with it forever.

Behold, mget! The metaget (+ mset, mdelete) commands are designed to give the ascii protocol parity with the binary protocol. Further, it extends the protocol to give atomic stampeding herd protection on MISS, stale-while-revalidate, and other features.

**NOTE: The rest of this comment is out of date. The mechanics are the same but examples will not match. See t/metaget.t and doc/protocol.txt for use examples and current syntax**

```
set foo 0 0 2
hi
STORED
mget foo
META foo exp=-1 la=2 cas=1 fetch=no cls=1 size=64
END
mget foo stfu
VALUE foo stfu 0 2 -1
END
mget foo st
VALUE foo st 2 -1
END
```
first VALUE returns client flags, size, TTL remaining, and doesn't bump the LRU
second VALUE returns size, and TTL remaining.

Without arguments it outputs the full
metadata in a format similar to LRU crawler... but without needing to
escape the key first. split by space, then by = for indexes 2+ for
parsing.

If you supply an argument to mget, it is interpreted as a set of option
flags. This is an optimized command, using fast itoa's to write numbers into the client write buffer.
For a single get it should rival or beat the existing (ancient) get code path.

Options enable what gets returned (numerically) on the VALUE line. The
options return int a strict order: if you say 's' (size) last, the
first value after the key is still the size. I might still make this
flexible, but it makes the parsing more difficult. As new options are
added in the future, they'll get independent flags and will be inserted
at the end of the current list. [NOTE: _probably_ doing flexible ordering]

Next, 'v' will conditionally enable returning the value of the object.
'u' will disable bumping the LRU. Allows fetch the value
of an object without bumping it to the top of the LRU.

Further, this command will supplant GETS/TOUCH/GAT/GATS (get-and-touch + cas variants):

`mget foo T 30`

... will update the TTL to 30 seconds from now, the same as a touch
command.

`mget foo Tsfv 30`

... is equivalent to the GAT command

`mget foo Tsfvc 30`

... is equivalent to the GATS command.

arguments that take tokens will consume them in a similar strict order
as returning them in the VALUE line. [NOTE: as above, probably a flexible order]

Now, we extend this into an anti-stampede system by overriding the
usage of CAS values:

`mget foo Ncfsv ttl`

... if this command results in a MISS, an item with an empty value will
be vivified, with a ttl. The CAS is also returned. For response:

```
mget foo Nfsvc ttl\r\n
VALUE foo NcfsvW [flags: 0] [size: 0] [cas: N]
\r\n\r\n (0b value)
```
The W flag is added to the return set, indicating this client has "Won" the right to recache this item. The CAS value is now used to overwrite the temporary item with the real value. A second client would see:

```
VALUE foo NcfsvZ [flags: 0] [size: 0] [cas: N]
\r\n\r\n
```

Z indicating that a token has been handed out to a client, so this value shouldn't be used. A lack of 'W' means it didn't win the "race" either. Clients can make informed decisions on sleep/waiting, fetching from a secondary data source, or giving up. They can also include the 't' flag to see how much TTL is remaining on a locked token.

In some cases "stale" indicated values can be useful:

client A: miss -> win, 15s TTL. CAS 1.
client B: (30s later) miss -> win (object expired), CAS 2.
client A: (5s later) mset CAS 1 (attempting to store but CAS is now old)

server notes that CAS value is less than the current CAS value. it can
allow a "stale" store: ignore the supplied TTL (keep remaining from
original object), and keep CAS as what it was before (the higher value). Now, with a
value rather than nothing at all. An 'X' flag is returned to indicate a value is "stale". Clients can decide on whether it's safe to use a stale value.

client B: (5s later) mset CAS 2.

... CAS now matches, "latest" value wins, stale and token markers are removed, TTL is acquired, value updated, etc.

---

Next: "invalidating to stale"

`mdelete foo T <ttl>`

sets TTL to new TTL (ideally low: 30s or so), flipping the stale bit if
not already set. Also bumps the CAS value. The next client:

```
client A:
mget foo svfc
VALUE foo svfcXW [size] [flags] [cas]
...data...

client B:
mget foo svfc
VALUE foo svfcXZ [size] [flags] [cas]
...data...
```

client A fetches a value that was set to stale, sees flag X (item is stale), and flag W (won for recache). It also gets a value back. It can return the value to the caller, and kick off an asynchronous recache.

if another client comes in and "shadow deletes" this item after client
A has already won, the CAS would reset, internal token flag would be undone.
Starting the dance again.

---

Internally: This is implemented via the pre-existing CAS value (8 bytes per item), and two internal flag bits (one for STALE, one for TOKEN_SENT). I had described another implementation which uses client flags, but this should be much simpler to integrate. The command may be extended later to allow atomic test-and-sets for client flag bits.

---

Finally: multiple valid clients. EDIT: This no longer works innately.

---

Clients are expected to create an abstraction around this, though with
some examples it shouldn't be as complicated as this writeup makes it
sound.

See the next comment for a more complete spec.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -117,6 +117,9 @@ For example, a metaget with flags of 'st' would return tokens for "size" and
 Syntax errors are handled the same as noted under 'Error strings' section
 below.
 
+For usage examples beyond basic syntax, please see the wiki:
+https://github.com/memcached/memcached/wiki/MetaCommands
+
 Expiration times
 ----------------
 
@@ -610,8 +613,6 @@ When combined with the X flag, or the client N or R flags, this extra response
 flag indicates to a client that a different client is responsible for
 recaching this item. If there is data supplied it may use it, or the client
 may decide to retry later or take some other action.
-
-[TODO: link to wiki for use case examples]
 
 Meta Set
 --------
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -59,7 +59,7 @@ key. The client sends a command line, and then a data block; after
 that the client expects one line of response, which will indicate
 success or failure.
 
-Retrieval commands (there are two: "get" and "gets") ask the server to
+Retrieval commands ("get", "gets", "gat", and "gats") ask the server to
 retrieve data corresponding to a set of keys (one or more keys in one
 request). The client sends a command line, which includes all the
 requested keys; after that for each item the server finds it sends to
@@ -75,6 +75,47 @@ ending with "END" on the last line.
 A command line always starts with the name of the command, followed by
 parameters (if any) delimited by whitespace. Command names are
 lower-case and are case-sensitive.
+
+Meta Commands [EXPERIMENTAL: MAY CHANGE]
+-------------
+
+Meta commands are a set of new ASCII-based commands. They follow the same
+structure of the basic commands but have a new flexible feature set.
+Meta commands incorporate most features available in the binary protocol, as
+well as a flag system to make the commands flexible rather than having a
+large number of high level commands. These commands completely replace the
+usage of basic Storage and Retrieval commands.
+
+Meta commands are EXPERIMENTAL and may change syntax as of this writing. They
+are documented below the basic commands.
+
+These work mixed with normal protocol commands on the same connection. All
+existing commands continue to work. The meta commands will not replace
+specialized commands that do not sit in the Storage or Retrieval categories
+noted in the 'Commands' section above.
+
+All meta commands follow a basic syntax:
+
+cm <key> <flags> <token1> <token2> <...>\r\n
+
+Where 'cm' is a 2 character command code. The number of tokens supplied is
+based off of the flags used.
+
+Responses look like:
+
+RC <flags> <token1> <token2> <...>\r\n
+
+Where 'rc' is a 2 character return code. The number of tokens returned are
+based off of the flags used.
+
+Flags are single character codes, ie 'q' or 'k' or 'I', which adjust the
+behavior of the command. The flags are reflected in the response. The order of
+which tokens are consumed or returned depend on the order of the flags given.
+For example, a metaget with flags of 'st' would return tokens for "size" and
+"TTL remaining" in that order. 'ts' would return "TTL remaining" then "size".
+
+Syntax errors are handled the same as noted under 'Error strings' section
+below.
 
 Expiration times
 ----------------
@@ -400,6 +441,186 @@ VALUE <key> <flags> <bytes> [<cas unique>]\r\n
 
 - <data block> is the data for this item.
 
+Meta Get
+--------
+
+The meta get command is the generic command for retrieving key data from
+memcached. Based on the flags supplied, it can replace all of the commands:
+"get", "gets", "gat", "gats", "touch", as well as adding new options.
+
+mg <key> <flags> <tokens>*\r\n
+
+- <key> means one key string. Unlike "get" metaget can only take a single key.
+
+- <flags> are a set of single character codes ended with a space or newline.
+
+- <tokens>* means zero or more tokens, based on the flags supplied. They are
+  consumed in order specified by the flags.
+
+After this command, the client expects an item to be returned, received as a
+text line followed by an optional data block. After the item is transferred,
+the server sends the string
+
+"EN\r\n"
+
+to indicate the end of the response.
+
+If a response line appearing in a retrieval request is not sent back
+by the server this means that the server does not
+have the item (because it was never stored, or stored
+but deleted to make space for more items, or expired, or explicitly
+deleted by a client).
+
+An item sent by the server looks like:
+
+VA <flags> <tokens>*\r\n
+<data block>\r\n
+
+- <flags> are a reflection of the flags sent to the server. Extra flags can be
+  added to the response based on the execution of the command.
+
+- <tokens>* are tokens returned by the server, based on the flags supplied.
+  They are added in order specified by the flags sent.
+
+- <data block> is the data for this item. Note that the data black is
+  optional, requiring the 'v' flag to be supplied.
+
+The flags used by the 'mg' command are:
+
+- c: return item cas token
+- f: return client flags token
+- h: return whether item has been hit before as a 0 or 1
+- k: return key as a token
+- l: return time since item was last accessed in seconds
+- O (token): consumes a token and copies back with response
+- q: use noreply semantics for return codes.
+- s: return item size token
+- t: return item TTL remaining in seconds (-1 for unlimited)
+- u: don't bump the item in the LRU
+- v: return item value in <data block>
+
+These flags can modify the item:
+- N (token): vivify on miss, takes TTL as a argument
+- R (token): if token is less than remaining TTL win for recache
+- T (token): update remaining TTL
+
+These extra flags can be added to the response:
+- W: client has "won" the recache flag
+- X: item is stale
+- Z: tem has already sent a winning flag
+
+The flags are now repeated with detailed information where useful:
+
+- h: return whether item has been hit before as a 0 or 1
+- l: return time since item was last accessed in seconds
+
+The above two flags return the value of "hit before?" and "last access time"
+before the command was processed. Otherwise this would always show a 1 for hit or
+always show an access time of "0" unless combined with the "u" flag.
+
+- O (token): consumes a token and copies back with response
+
+The O (opaque) token is used by this and other commands to allow easier
+pipelining of requests while saving bytes on the wire for responses. For
+example: if pipelining three get commands together, you may not know which
+response belongs to which without also retrieving the key. If the key is very
+long this can generate a lot of traffic, especially if the data block is very
+small. Instead, you can supply an "O" flag for each mg with tokens of "1" "2"
+and "3", to match up responses to the request.
+
+Opaque tokens may be up to 32 bytes in length, and are a string similar to
+keys.
+
+- q: use noreply semantics for return codes.
+
+Noreply is a method of reducing the amount of data sent back by memcached to
+the client for normal responses. In the case of metaget, if an item is not
+available the "VA" line is normally not sent, and the response is terminated by
+"EN\r\n".
+
+With noreply enabled, the "EN\r\n" marker is suppressed. This allows you to
+pipeline several mg's together and read all of the responses without the
+"EN\r\n" lines in the middle.
+
+Errors are always returned.
+
+- u: don't bump the item in the LRU
+
+It is possible to access an item without causing it to be "bumped" to the head
+of the LRU. This also avoids marking an item as being hit or updating its last
+access time.
+
+- v: return item value in <data block>
+
+The data block for a metaget response is optional, requiring this flag to be
+passed in.
+
+These flags can modify the item:
+- N (token): vivify on miss, takes TTL as a argument
+
+Used to help with so called "dog piling" problems with recaching of popular
+items. If supplied, and metaget does not find the item in cache, it will
+create a stub item with the key and TTL as supplied. If such an item is
+created a 'W' flag is added to the response to indicate to a client that they
+have "won" the right to recache an item.
+
+The automatically created item has 0 bytes of data.
+
+Further requests will see a 'Z' flag to indicate that another client has
+already received the win flag.
+
+Can be combined with CAS flags to gate the update further.
+
+- R (token): if token is less than remaining TTL win for recache
+
+Similar to and can be combined with 'N'. If the remaining TTL of an item is
+below the supplied token, return a 'W' flag to indicate the client has "won"
+the right to recache an item. This allows refreshing an item before it leads to
+a miss.
+
+- T (token): update remaining TTL
+
+Similar to "touch" and "gat" commands, updates the remaining TTL of an item if
+hit.
+
+These extra flags can be added to the response:
+- W: client has "won" the recache flag
+
+When combined with N or R flags, a client may be informed they have "won"
+ownership of a cache item. This allows a single client to be atomically
+notified that it should cache or update an item. Further clients requesting
+the item can use the existing data block (if valid or stale), retry, wait, or
+take some other action while the item is missing.
+
+This is used when the act of recaching an item can cause undue load on another
+system (CPU, database accesses, time, and so on).
+
+- X: item is stale
+
+Items can be marked as stale by the metadelete command. This indicates to the
+client that an object needs to be updated, and that it should make a decision
+o if potentially stale data is safe to use.
+
+This is used when you want to convince clients to recache an item, but it's
+safe to use pre-existing data while the recache happens.
+
+- Z: item has already sent a winning flag
+
+When combined with the X flag, or the client N or R flags, this extra response
+flag indicates to a client that a different client is responsible for
+recaching this item. If there is data supplied it may use it, or the client
+may decide to retry later or take some other action.
+
+[TODO: link to wiki for use case examples]
+
+Meta Set
+--------
+
+Meta Delete
+-----------
+
+Meta No-Op
+----------
 
 Slabs Reassign
 --------------
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -461,7 +461,7 @@ ME <key> <k>=<v>*\r\nEN\r\n
 Each of the keys and values are the internal data for the item.
 
 exp   = expiration time
-la    = last access time
+la    = time in seconds since last access
 cas   = CAS ID
 fetch = whether an item has been fetched before
 cls   = slab class id
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -444,6 +444,31 @@ VALUE <key> <flags> <bytes> [<cas unique>]\r\n
 
 - <data block> is the data for this item.
 
+Meta Debug
+----------
+
+The meta debug command is a human readable dump of all available internal
+metadata of an item, minus the value.
+
+me <key>\r\n
+
+- <key> means one key string.
+
+The response looks like:
+
+ME <key> <k>=<v>*\r\nEN\r\n
+
+Each of the keys and values are the internal data for the item.
+
+exp   = expiration time
+la    = last access time
+cas   = CAS ID
+fetch = whether an item has been fetched before
+cls   = slab class id
+size  = total size in bytes
+
+Others may be added.
+
 Meta Get
 --------
 
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -543,8 +543,8 @@ The flags are now repeated with detailed information where useful:
 - l: return time since item was last accessed in seconds
 
 The above two flags return the value of "hit before?" and "last access time"
-before the command was processed. Otherwise this would always show a 1 for hit or
-always show an access time of "0" unless combined with the "u" flag.
+before the command was processed. Otherwise this would always show a 1 for
+hit or always show an access time of "0" unless combined with the "u" flag.
 
 - O (token): opaque value, consumes a token and copies back with response
 
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -492,7 +492,7 @@ The flags used by the 'mg' command are:
 - h: return whether item has been hit before as a 0 or 1
 - k: return key as a token
 - l: return time since item was last accessed in seconds
-- O (token): consumes a token and copies back with response
+- O (token): opaque value, consumes a token and copies back with response
 - q: use noreply semantics for return codes.
 - s: return item size token
 - t: return item TTL remaining in seconds (-1 for unlimited)
@@ -518,7 +518,7 @@ The above two flags return the value of "hit before?" and "last access time"
 before the command was processed. Otherwise this would always show a 1 for hit or
 always show an access time of "0" unless combined with the "u" flag.
 
-- O (token): consumes a token and copies back with response
+- O (token): opaque value, consumes a token and copies back with response
 
 The O (opaque) token is used by this and other commands to allow easier
 pipelining of requests while saving bytes on the wire for responses. For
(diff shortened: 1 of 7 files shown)


dormando:
still working on clarifying the vague CAS related commands into proper fundamentals:

- [ ] add missing output lines: last access, classid, fetched.
- [ ] input: T for ttl (touch, gat, gats)
- [ ] input: vivify on miss (not too useful on its own but should be standalone)
- [ ] input (token): if bit N is 0, flip to 1 and report win (combine with above for miss->lease flow)
- [ ] input (token+token?): (B) if A is set, and B is not set, flip both and report win.
- [ ] input (token): (D) unconditionally flip bit N to 1, bump CAS (for marking object as stale; combine with T to edit the TTL remaining)

Use D, then B to do stale-while-revalidate with an external process marking items as stale (rather than them being expired)
*maybe* due to ordering, combining D with B could allow explicitly marking an item as stale + win without extra roundtrips.

To fully flesh out flow, need an 'mset' command to complement. allows bit testing + CAS testing:
- supplied CAS is lower and bit N is 1, copy existing TTL and client flags, overwrite value.
- supplied CAS is the same, overwrite all.
... "old" CAS cannot overwrite final CAS since bit N would be 0.

might need a generic "execute only if bit N is X" for both.

dormando:
Extra notes in this comment:

I originally wanted the bits to be consumed in order, ie:

`mget foo sft`

... would output `VALUE size flags ttl` in that order. then conditionals would be linked together:

`mget foo RD 30 1` (if TTL < 30 AND bit 1 is 0, flip to 1 and return win)

then conditionals could be stacked or chained... but two complicating issues:

1. some flags like 'u' need to be known before the item is fetched.
2. binprot would need to use the same flag string, and then can't use a bit field.

not 100% convinced either of these things are the show stoppers, but I'm implementing this first pass with a strict ordering. Changing it from there is a matter of cut/pasting a bunch.

(1) could be dealt with by flipping through the options twice, or forcing clients to pay attention to the order : 'u' would have to come before anything that tests, reads, or modifies the item (gross). Though most mod options will require the item lock be held... then we end up with a "we want to modify the item" flag.

Performance should also be better, since with strict ordering every potential flag has to be tested. With loose ordering having two flags would only examine those two options.

In the end that ends up being pretty complicated. With strict ordering it's a lot easier on client authors (maybe?): flag X influences flag Y and so on.

hopefully I'll understand this better after writing out the tests.

sergmour:
Hey, @dormando, this looks great but the command becomes pretty overloaded. My understanding is that the goal is to do all that is needed in one roundtrip, which is good. But client libraries will need to resort to multiple function implementations for mget with different options.
Would it be better to have 2 commands?
- mget: simple, no mutations
- mset: allows modifications and also returns data
That might simplify things. Also, the name mget conflicts with ftp's mget meaning. Not a big deal but it makes the command name confusing.
Anyway, these are just very superficial considerations. And thank you for working on that - we really need it.

dormando:
So 'mset' is going to need to exist to implement the CAS-adjusted mutations. I could potentially overload it all into one command but I don't think I want to do that for the same reasons you said. When both commands exist one could easily start piece-mealing the options between them... but I think it ultimately makes sense that any command with a payload be mset. Then mset comes with some standard tokens (size, ttl, client flags, etc). There'll probably even be some overlap in the flags. That does mean mget is the most frequently used command, which also makes sense to me. The _intent_ of the user is to fetch or adjust an item, not overwrite a value, for all of my original examples.

You're right too; for the most part you can do this with extra round-trips:
```
get foo -> miss
add fooT etc -> OK [secondary modified key]
[double-trip, fails to add]
winner does set foo
winner delete's fook [but other clients are now succeeding at the original get]
```

Except this doesn't quite work. maybe for binprot since you can't get a CAS value back from ADD in ASCII. Also, most clients actually eat the result from ADD! Even if I wanted to I couldn't implement this properly in most memcached clients. This new command gives us a chance to fix this by making the return value always important.

I do intend that client libraries create sets of API calls with it, ie: get, gets, touch, GAT, GATS and so on should be library calls that mutate into mget option sets. From there they can add a "metadata" hash/flag field, for returning the TTL alonside data, etc. "lease-get" and "lease-set" like commands for doing a few of the standard CAS dances should be done as well.

... and then clients should offer a more raw mode for advanced users.

sergmour:
This makes sense. My other maybe selfish concern was about using the smart proxies in front of memcached shards, say, mcrouter. It would become difficult to define the routing logic there. The parser there would need to parse all the flags to figure out what the command really does to determine the route being used. It shouldn't be a primary concern for memcached, though, just a consideration.

dormando:
Not much I can do about that. The key is still clear and in most cases it should just forward, so I don't see it being complicated. They don't have binprot at all so they've never had much compatibility with OSS memcached's features anyway...

sergmour:
They do a pretty good job for simple key/val use cases (no cas, etc.). And we've been using mcrouter with OSS memcached up to 1.5.10. Anyway, just wanted to mention that. Thanks for the well though explanations!

dormando:
@sergmour may I ask what your mcrouter use case looks like?

amcrn:
I'm not sergmour, but I use mcrouter with memcached and share a similar concern.

Here's an example of a mcrouter config we use in Production:

```
{
  "pools" : {
    "all" : {
      "server_timeout" : 100,
      "servers" : [
        "memcached1.us-west-1.example.com:11211",
        "memcached2.us-west-1.example.com:11211",
        "memcached3.us-east-1.example.com:11211",
        "memcached4.us-east-1.example.com:11211",
        "memcached5.eu-cent-1.example.com:11211",
        "memcached6.eu-cent-1.example.com:11211",
      ]
    },
    "memcached1.us-west-1.example.com" : {
      "servers" : [
        "memcached1.us-west-1.example.com:11211"
      ]
    },
    "memcached2.us-west-1.example.com" : {
      "servers" : [
        "memcached2.us-west-1.example.com:11211"
      ]
    }
  },
  "route" : {
    "operation_policies" : {
      "delete" : "AllFastestRoute|Pool|all",
      "get" : {
        "children" : [
          {
            "children" : [
              "PoolRoute|memcached1.us-west-1.example.com",
              "PoolRoute|memcached2.us-west-1.example.com"
            ],
            "type" : "MissFailoverRoute"
          },
          {
            "children" : [
              "PoolRoute|memcached2.us-west-1.example.com",
              "PoolRoute|memcached1.us-west-1.example.com"
            ],
            "type" : "MissFailoverRoute"
          }
        ],
        "type" : "RandomRoute"
      },
      "set" : "AllFastestRoute|Pool|all"
    },
    "type" : "OperationSelectorRoute"
  }
}
```

Route Handler Reference: https://github.com/facebook/mcrouter/wiki/List-of-Route-Handles

Simplified:

* `get` operations are round-robin'd (via RandomRoute) between memcached1 and memcached2, but if the result is a miss, the other is then consulted (via MissFailoverRoute).
* `set` and `delete` requests are immediately sent to memcached1 through memcached6 in parallel, returning the first non-error reply to come back, or if all replies are errors, then the last error reply. Other requests complete in the background and their replies are silently ignored.

OperationSelectorRoute defines the routes for a given operation. If `mget` isn't split into `mget` and `mset`, this can be problematic. Why: We'd want the non-mutating `mget` commands to share the `get` routing, whereas we'd want the mutating `mset` commands to share the `set` and `delete` routing. This won't be possible unless mcrouter is modified to do flag inspection.

Hope that helps and really excited to see that MGET is coming :)

dormando:
good to know, thanks!

Even in the original text I'd written `mset` or `cas` as used to update data. not sure where I got people lost, but it's a long post :)

in your case, when MissFailoverRoute is activated (or on a full miss), without any kind of lease structure a number of requests could end up in flight for the same key and you'll have a lot of SET's. With mget the only difference is the router would have to remember the cas result for the replay back to the original server with mset. The thundering herd would disappear. still have to change the routing handlers a bit though.

I've been combing things over and I'll probably split off the mget-invalidation into mdelete. will update the other comments with details. otherwise all the code paths I can find still work.

sergmour:
@dormando : our use case is simple - we need to handle High Availability (HA) and Disaster Recovery (DR) events. To achieve this we configure up to 6 geo-distributed replicas of cache. The cache is huge, each replica may consist of up to 90 memcached shards along with multiple mcrouters to handle the throughput and latency. Read ops are handled using MissFailoverRoute (serial) and write ops are handled using parallel routes distributing updates to all replicas.
One of the reasons for our need for metaget is to warm up cold replicas. Say, you run a get against a cold replica. On miss, if the data is retrieved from the warm replica the cold replica is updated in the background asynchronously. Mcrouter does have a couple of options for that. But the main inconvenience is in not knowing the TTL value. It needs to be a default TTL, which will lead to inconsistent expiration time between replicas. This, in turn, can lead to replicas warming up each other in a vicious cycle. Mcrouter does have a metaget command, but this is to be used with internal facebook implementation of memcached.
Interestingly, for replica warm-up use case there is no reason for metaget to be performant as it works asynchronously in the background.
Our customers also ask for metaget as they want to know the expiration time and potentially other things from metadata. This does need to perform.
I think if we had simpler separate commands for mutating or read-only mget and lease get, this would simplify the higher level implementations like proxies and clients. But I'd agree that this should not be a limiting factor for memcached protocol in any way. I see that you've already put a lot of thought into mget, thanks for that!

dormando:
@sergmour got it, thanks! Yes that sucks. Pretty sure I have this worked out by splitting of mdelete's specifically for invalidations. Else you'd always be using get or mget (lease mode) so there's no reason to still have two commands, since both would also need the full set of options for giving ttl yes/no to cover all use cases.

Even in the async case it should be performant: having a server come up from death causes a burst of traffic. if metaget were too CPU intensive on servers you'd have load spikes. I'll be writing up a bunch of test cases (and probing people to write ones that would fail)

dormando:
Ok. flattened out my last weeks' worth of notes into the top two comments. The long third comment can be ignored. Will edit further notes into this comment. :) Hopefully this is a lot less scatterbrained.

The biggest impact is deciding to kill a deprecated feature to free up some internal flag bits to use instead of the client flags, though some functionality is lost. Also win/stale/token indicators are decided, commands are mapped out, and argument flags are closer to final.

dormando:
Alright, error handling and parsing is a lot closer to where I'm happy with things. Need to run through an example client at this point, or simply write more tests.

Have some more things to fix with error handling. The key is now returned for most errors, outside of egregious parsing problems, which should be exclusively client bugs rather than user or internal error. mset in noreply/quiet mode still eats the response too often.

One of the things binprot fixed with quiet mode was still allowing "interesting" error responses, ie OOM's, CAS failures, and similar. ASCII noreply normally eats them all with impunity. I'm ensuring the semantics of the m* commands mirrors this, since they're dropping the "noreply maybe at the end of the string" thing. The latest commit starts doing this properly.

Some more refactoring to reduce redundant code is possible too... maybe a macro or something. I kind of need some client author opinions, or attempt to implement this in some clients and see how it goes. Originally I wanted to finish this a few weeks ago, however slowing down and chipping at it has been beneficial in making the semantics a lot cleaner. Had I rushed it I may still be using client flags which on further reflection could result in being DOA. I've also added the flag return, which gave me the semi out of band channel for reporting token status and that works pretty well.

It's all rebased on top of the latest 'next' branch.

aisk:
Sorry, not Engilish native speakingSorry, non-English-native-speaking, please ignore my reply if I have a misunderstanding on these APIs.
I have these questions:

1. Is the new APIs will replace all of the old ones?
2. I saw the `mget` API can not only query the information about a key but also update its meta info (like TTL). Can we give idempotency on the command names like the HTTP methods,  like mget methods never change the value or information on a key, for some proxies or middlewares can judgment the idempotency of the command or some future benefits?

dormando:
Thanks for taking a look!

1) Not all, but most for ASCII: get, set, delete, add, replace, touch, maybe some others I'm forgetting. All the common ones.

2) Only mset can update values, mget cannot. Only mdelete can invalidate data. mget _can_ update TTL or autovivify items for token-atomic operations, but a proxy can identify that by the flags used. I believe I have the proxy use cases worked out, but might be missing something. It's similar to some HTTP functionality (stale-while-revalidate, etc) but the protocols don't completely line up.

We can talk in Chinese if that's easier for you, or I can have some parts of this documentation translated.

dormando:
@aisk was I able to answer your questions?

Your second question has a somewhat complex answer. I believe any proxy cases only should worry about when values are being updated, vs metadata. Though it's been long enough now since I did the initial work that I should walk through it again..

aisk:
Sorry late for reply. I'm totally fine with this commands and will try to implemented it if it's done.

memcached.c
```diff
@@ -4181,6 +4263,740 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+                rtokens++;
+                break;
+            case 'C':
+                if (!safe_strtoull(tokens[rtokens].value, &req_cas_id)) {
+                    goto error;
+                }
+                comm = NREAD_CAS;
+                rtokens++;
+                break;
+            case 'S':
```
tharanga: Is it useful to have the mset without `S` ? Looks like we are truncating (`vlen=0`) the item in that case.

memcached.c
```diff
@@ -4181,6 +4263,740 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+            if ((it->it_flags & ITEM_TOKEN_SENT) == 0) {
+                // If we're stale but no token already sent, now send one.
+                won_token = true;
+            }
+        }
+
+        for (i = 0; i < olen; i++) {
+            switch (opts[i]) {
+                case 'T':
+                    // FIXME: I broke this.
```
tharanga: Looks like we need to fix this one to get the `mget foo T <ttl>` working.

memcached.c
```diff
@@ -4181,6 +4263,740 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+                case 'h':
+                    *p = ' ';
+                    if (it->it_flags & ITEM_FETCHED) {
+                        *(p+1) = '1';
+                    } else {
+                        *(p+1) = '0';
+                    }
+                    break;
+                default:
+                    fprintf(stderr, "Unknown option: %c\n", opts[i]);
```
tharanga: What about `u` ? Or you didn't implement it until finding a good way to test it?

tharanga: Shouldn't we send an error message to the client for invalid options instead of printing to the `stderr` (probably based on the verbosity level)

memcached.c
dormando: People set zero byte items sometimes. It could be a shortcut or some other thing; typically not useful no.

memcached.c
dormando: As the comment says, yes :)

memcached.c
dormando: re: u, there's just no way to "fetch the item with a bump and get the last_access time before the bump" in the item API. Need to make that happen first. It's just a TODO.

re: printf. yes, that's a development/debug line right now. I wrote that before out_errstr existed.

memcached.c
```diff
@@ -2961,6 +3008,24 @@ enum store_item_type do_store_item(item *it, int comm, conn *c, const uint32_t h
             c->thread->stats.slab_stats[ITEM_clsid(old_it)].cas_hits++;
             pthread_mutex_unlock(&c->thread->stats.mutex);
 
+            STORAGE_delete(c->thread->storage, old_it);
+            item_replace(old_it, it, hv);
+            stored = STORED;
+        } else if (c->set_stale && ITEM_get_cas(it) < ITEM_get_cas(old_it)) {
```
tharanga: Don't we need to clear the `set_stale` on failures? Otherwise the flag is get carried over to subsequent commands.

memcached.c
```diff
@@ -3237,6 +3303,7 @@ static void server_stats(ADD_STAT add_stats, conn *c) {
     APPEND_STAT("cmd_set", "%llu", (unsigned long long)slab_stats.set_cmds);
     APPEND_STAT("cmd_flush", "%llu", (unsigned long long)thread_stats.flush_cmds);
     APPEND_STAT("cmd_touch", "%llu", (unsigned long long)thread_stats.touch_cmds);
+    APPEND_STAT("cmd_mget", "%llu", (unsigned long long)thread_stats.mget_cmds);
```
tharanga: Stats for mset and mdelete?

memcached.c
dormando: I'll check, good catch.

dormando:
Took another read-through on this over the weekend. It's rebasing fine.

- There're a lot of small things left undone. Half are documented in the test file, half here, with the third half in FIXME's, TODO's, and debug printf's.
- Overall these are minor; structurally not much should change. Just a good amount of typing and testing.
- Tests aren't bad, but more rounds will help a lot; specifically testing proxied scenarios is a must.

Feedback I'm looking for at the moment is higher level. Do these commands have major conflicts or distsys-level races/bugs? I initially held off on the cleanup efforts to not waste time on it if I had to make major changes to the overall protocol first... but it hasn't happened yet, so it may be safe enough to just clean it up and try.

tharanga:
Is the API still evolving? When finalized, It would be great if `protocol.txt` can be updated by summarizing all the details.

dormando:
I mean; it might evolve if someone tells me there's a bug :) I've been nearly fruitlessly asking for feedback for months, so as usual I'll probably just Do The Thing and hope for the best. Hopefully if there is anything wrong I'll find it as I write out more tests, but distributed systems protocols like this are just hard in general.

tharanga:
@dormando can we add another option to `mget` to get the expiration time in addition to the ttl?
We are using the expiration time given by the `lru_crawler` for cache warming and would like to use the same semantics with `mget`

dormando:
@tharanga That may have been a mistake to be honest. Memcached's monotonic clock can drift over time, so if you get "TTL remaining" from memcached it's accurate, but if it projects onto its clock to give you "expiration time" it can actually be wrong :/ I went back and forth on this a bit.

For the metadump use case you're collecting a ton of data all at once then remaking it elsewhere, so the workflow makes more sense. For the metaget use case you should be using the value shortly after fetching it, so using the TTL it hands you back as the TTL you supply to a set should be simplest? Or are you still buffering everything?

tharanga:
Yes we are thinking of buffering. With TTL, I believe error margins will be dominated by network round trips/ connectivity/ hops. We should be able to workaround these with an acceptable upper bound.

dormando:
Sounds about right; if you really need the absolute time but won't be reusing the object for a while you can add it to current time with the process reading the data back... since there's only a second granularity in TTL to begin with even a few hundred MS of worst-case network latency won't be more than the natural drift of cutting everything to 1s boundaries.

memcached.c
dormando: I have it resetting to false at the end of do_store_item... that's probably not right since it could error out before calling do_store_item. it's also cleared on conn_new, so that takes care of any case where the connection closes...

Alright, moved it around; forced false in complete_nread_ascii (which called store_etc) and the switch to conn_swallow during conn_nread. Can't switch between bin/ascii without reconnecting.

will keep an eye out for a better way of doing this. Don't want to kill it on every conn_new to avoid the extra memory twiddle for every non-store command :/

memcached.c
dormando: oh I misread you; the 'u' is handled in the preparser above. some common and idempotent flags are done up there in the common code and passed back with the of bits. 'u' already works.

dormando: I'm a liar; 'u' wasn't properly implemented yet. it is now.

dormando:
Added a bunch of small fixes, new flags, tests, etc. Been circling towards finalized return values.

noreply should match binrpot: non-errors get eaten (except for GET results, then that eats miss which is the same as before). IE: for mdelete "DELETED" gets eaten, but "NOT_FOUND" gets returned.

To do this properly, with 'mget foo stkvq' (quiet/noreply), it should eat the END after each fetch.
Then add an "mnop" command, which simply returns "END". Pipelined requests stack everything together with either the mnop at the end or the final mget without the 'q'. Swap k for small values of o when keys are long and you get optimal bytes on wire for multigets.

One thing ASCII get does is always returns the key with its data:
```
get foo
VALUE foo 0 2
hi
END
```

When you have very long keys (perhaps relative to short values) this can waste a lot of bytes on the wire. So the "opaque" was added, which is a small number that's blindly copied from request to response. Unfortunately the opaque is always sent/retrieved, wasting bytes when unused (as well as CAS, key length).

There's now an opaque flag:
```
mget foo tco 500
VALUE foo tco -1 1 500
END
```

Great, but we still have the key. The next step is to make the key optional. Then we'll have:
```
mget foo tco 500
VALUE tco -1 1 500
END
```
or:
```
mget foo tck
VALUE tck -1 1 foo
END
```

Another option here is adding 'o' silences the key and puts the opaque where the key was before. But then the parser is more complex?

This doesn't exist for mdelete or mset yet. mset is a bit harder since the return line is far from the request and now has to be conditional based on mset vs set. In binprot every command has opaque, but only the get commands can return a key. This still makes sense to me. I was trying to add the key to responses for mset/mdelete, but supporting opaque properly is probably the right idea.

Set opaque limit to 32 bytes, but want to shrink that to 8-16 to make it easier to stuff into the conn structure without being wasteful.

t/metaget.t
```diff
@@ -0,0 +1,450 @@
...
+use Test::More;
+use FindBin qw($Bin);
+use lib "$Bin/lib";
+use MemcachedTest;
+
+my $server = new_memcached();
+my $sock = $server->sock;
+
+# command syntax:
+# mget [key] [flags] [tokens]\r\n
```
tharanga: Nit : update commands reflect short versions

dormando: fixed

dormando:
Ok doc/protocol.txt first pass is done. Most of the TODO items from the roundup are done or punted.

Got a handful of FIXME's/TODO's in the code that I'm following up on and need to re-test extstore. Commands and response codes are all standardized now. It's pretty hard to tell what the use cases are from doc/protocol.txt, so I'll clean up the examples I've been writing out in here into a wiki page to link to.

Also I'm going to move the meta response (mg with no arguments) to its own command `me` since it feels wrong as-is. I may then have 'mg <key>' (and no flags) be the same as 'fstv', meaning client flags, size, ttl, value. Otherwise it'll be an error.

'md' also needs a no-flags mode to just delete as a typical 'delete' command.

Pretty sure we'll see this out in a release sometime tomorrow. Maybe that'll help get more interest into getting any major or minor issues fixed sooner :)

memcached.c
```diff
@@ -4191,6 +4287,837 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+        total += 3;
+        memcpy(c->wbuf + total, ITEM_key(it), it->nkey);
+        total += it->nkey;
+        c->wbuf[total] = ' ';
+        total++;
+
+        ret = snprintf(c->wbuf + total, c->wsize - (it->nkey + 12),
+                "exp=%d la=%llu cas=%llu fetch=%s cls=%u size=%lu\r\nEN\r\n",
+                (it->exptime == 0) ? -1 : (current_time - it->exptime),
+                (unsigned long long)(current_time - it->time),
```
tharanga: Looks like `la` is `seconds elapsed since the last access`, not the last access time.

dormando: That's correct. It suffers from the same drift issue as exptime otherwise, so I changed it in the same way :(

doc/protocol.txt
```diff
@@ -400,6 +444,366 @@ VALUE <key> <flags> <bytes> [<cas unique>]\r\n
...
+cls   = slab class id
+size  = total size in bytes
+
+Others may be added.
+
+Meta Get
+--------
+
+The meta get command is the generic command for retrieving key data from
+memcached. Based on the flags supplied, it can replace all of the commands:
```
tharanga: Are we planning to support multiple keys with meta commands in the future?

dormando: I don't think so; based on dealing with ascii-multiget (which only works with get!) it makes the network code horrible. In my next batch of work re-doing the network code it should be possible to batch syscalls for any stack of pipelined commands. So "me key1\r\nme key2\r\n" and then you'd get one syscall for all the responses, except it'll also work with set/delete/etc.

For now it'll be equivalent to the binary protocol which batches nothing ever :)

memcached.c
tharanga: +1, update the protocol.txt?

dormando: fixed

memcached.c
```diff
@@ -4191,6 +4287,837 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+        item_remove(it);
+        c->wbytes = total + ret;
+        c->wcurr = c->wbuf;
+        conn_set_state(c, conn_write);
+        c->write_and_go = conn_new_cmd;
+    } else {
+        out_string(c, "EN");
+    }
+    pthread_mutex_lock(&c->thread->stats.mutex);
+    c->thread->stats.mget_cmds++;
```
tharanga: stats only get updated for the `me` command, shouldn't this happen in the `process_mget_command` method?

memcached.c
```diff
@@ -4191,6 +4287,837 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+                break;
+            // mset-related.
+            case 'F':
+            case 'S':
+            case 'C':
+                tokens++;
+                break;
+            case 'I':
+                of->set_stale = 1;
+                break;
```
tharanga: Shouldn't we error out if user specifies at least one wrong option?

memcached.c
```diff
@@ -4191,6 +4287,837 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+    // we can tag plus the initial space.
+    // This could be simpler by adding two iov's for the header line, but this
+    // is a hot path so trying to keep those to a minimum.
+    fp = c->wbuf + 6;
+    memcpy(fp, opts, olen);
+    p = fp + olen;
+    fp--; // next token, or final space, goes here.
+
+    // scrubs duplicated options and sets flags for how to load the item.
+    rtokens -= _meta_flag_preparse(opts, olen, &of);
```
tharanga: Not sure whether we should handle the case where user supplies no valid flags?

memcached.c
```diff
@@ -4191,6 +4287,837 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+    }
+
+    // Copy opts into wbuf and advance pointer.
+    // We return the initial options + extra indicator flags.
+    // we reserve 4 bytes in front of the buffer, for up to three extra flags
+    // we can tag plus the initial space.
+    // This could be simpler by adding two iov's for the header line, but this
+    // is a hot path so trying to keep those to a minimum.
+    fp = c->wbuf + 6;
+    memcpy(fp, opts, olen);
```
tharanga: May be we can error if there are no valid options and move this after that check?

memcached.c
dormando: I should change it from mget -> meta. mg/ms/md count as get/set/delete commands. I believe that's the right approach because otherwise adopting the commands will screw everyone's metrics. I left a note about needing new stats for specific sub-actions like autoviv or mark-stale or etc.

memcached.c
dormando: Not honestly sure. I'd rather have the parser shared between all of the code, so bad flags are just ignored. Else you'd have to map everything out which is a lot more code.

memcached.c
dormando: ms is the only one currently enforcing that (needing at least 'S' flag) since mg and md can both operate with zero supplied flags. I'll think about it a bit.

memcached.c
dormando: fixed this instance. adding rest of counters is low prio atm.

dormando:
Apparently I wrote a comment in here then never clicked comment...

I did the next mergedown prep and finished a first pass at the wiki page yesterday. Think a few minor items left for a sanity check then I'll release it. Today or tomorrow doesn't really matter I think.

Will move the punted TODO's and thoughts into an issue similar to the restart code so we can continue to iterate.

doc/protocol.txt
```diff
@@ -400,6 +444,366 @@ VALUE <key> <flags> <bytes> [<cas unique>]\r\n
...
+- O (token): opaque to copy back.
+
+See description under 'Meta Get'
+
+- q: noreply
+
+See description under 'Meta Set'. In the case of meta delete, this will hide
+response lines with the code "DE". It will still return any other responses.
+
+- T (token): updates TTL
```
tharanga: May be it's good to mentioned this option is only valid with `I` for `md`

memcached.c
tharanga: `tokens` should be always >=0, can we use `-1` as the return code to indicate a parser error?

memcached.c
tharanga: `mg foo bar` returns `VA bar\r\nEND\r\n` (if foo exists), not an error, we silently gives back the garbage user provided :)

memcached.c
```diff
@@ -4191,6 +4287,837 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
...
+    } else {
+        failed = true;
+    }
+
+    if (of.locked) {
+        // Delayed bump so we could get fetched/last access time/etc.
+        // TODO: before segmented LRU, last-access time would only update
+        // every 60 seconds. Currently it doesn't update at all if an item is
+        // marked as ACTIVE. I believe this is a bug... in segmented mode
+        // there's no reason to avoid bumping la on every access.
```
tharanga: +1. So `mg <key> uv` is the only way to get the value without bumping the LRU position? If the ascending order of `la` values indicates the LRU order, we can use that to test the `u` flag.

dormando: The hit marker is accurate, so I used that to verify u in the tests. That's another way to check it too.

memcached.c
dormando: Not sure I follow? You still need to know which flags pertain to the sub-command in the parser here. That might make the code too obtuse.

memcached.c
dormando: `mg foo bar baz qux\r\n`
doesn't error either. that might be worth checking for too many tokens supplied
if the keys are long you'll get a client error though....
```
mg foo baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
CLIENT_ERROR options flags are too long
```

most people have really long keys.

what I was thinking of doing is filling the buffer with any valid flags, deduped, during the flag preparse phase. Though it could include flags that don't pertain to that object, the result would be pretty obviously different from what the user sent. That feels right along the "be liberal of what you accept and strict in what you send" mantra.

Else we should error on the first dupe or invalid char. I've been going back and forth on this for weeks but haven't been able to engage enough client authors to get an outside opinion, so I was thinking of punting this until the first release revision.

doc/protocol.txt
dormando: It does below, but I'll make that more explicit.

memcached.c
dormando: flag checking is now strict

memcached.c
dormando: decided to just make it strict. meh.

dormando:
Squashed it mostly down and merged from another branch.

dormando:
finally out in 1.5.19. marked experimental. hoping not a ton of time before it finalizes.

PR #489, opened 2019-05-12
-Y [filename] for ascii authentication mode

dormando:
Loads "username:password\n" tokens (up to 8) out of a supplied authfile.

If enabled, disables binary protocol (though may be able to enable both
if sasl is also used?).

authentication is done via the "set" command. A separate handler is
used to avoid some hot path conditionals and narrow the code
executed in an unauthenticated state.

ie:
```
set foo 0 0 7\r\n
foo bar\r\n
```
returns "STORED" on success. Else returns CLIENT_ERROR with some
information.

Any key is accepted: if using a client that doesn't try to authenticate
when connecting to a pool of servers, the authentication set can be
tried with the same key as one that failed to coerce the client to
routing to the correct server. Else an "auth" or similar key would
always go to the same server.

kill -HUP'ing memcached causes it to reload the authentication file. Haven't decided on a command or not yet.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,7 +23,8 @@ memcached_SOURCES = memcached.c memcached.h \
                     logger.c logger.h \
                     crawler.c crawler.h \
                     itoa_ljust.c itoa_ljust.h \
-                    slab_automove.c slab_automove.h
+                    slab_automove.c slab_automove.h \
+                    authfile.c authfile.h
 
 if BUILD_CACHE
 memcached_SOURCES += cache.c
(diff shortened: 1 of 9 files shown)


dormando:
TODO:

- [ ] disable UDP when auth enabled (should do this for SASL as well)
- [ ] doc/protocol.txt update
- [ ] tests
- [ ] stats counters? for auth fail/success
- [ ] some warning when the passfile is too large?

dormando:
whytf is this test failing... can't repro.

PR #493, opened 2019-05-16
Little mem leak fix proposal during unit tests.

devnexen:


dormando:
thanks!

dormando:
out in 1.5.15, thanks!

Issue #463, opened 2019-03-19
_logger_thread_parse_xx: LOGGER_PARSE_SCRATCH is not the length of keybuf


tnclong:
I find the `uriencode(le->key, keybuf, le->nkey, LOGGER_PARSE_SCRATCH);` is strange when studying from your code.

https://github.com/memcached/memcached/blob/master/logger.c#L201

dormando:
LOGGER_PARSE_SCRATCH is big enough to buffer an entire log line, I probably just didn't want to have separate constants for the same buffer.

When URI encoding a key, it's possible to have the resultant line be 3x longer than the real key (250 bytes -> 750), which is why it's not equal to key maxlen either.

dormando:
fixed in 1.5.15

PR #495, opened 2019-05-20
Using the correct buffer length when URI encoding keys

tharanga:
Issue : #463


dormando:
looks good, thanks!

dormando:
out in 1.5.15, thanks!

Issue #390, opened 2018-06-12
list defaults in --help output


dormando:
memcached has a bunch of default options: would be nice to list them in the output of --help alongside their options. keep the alignment of the output in mind.

sindhugudivada:
@dormando  could you please direct me to the docs which explain how I can set-up the dev environment and build the code etc..

PR #496, opened 2019-05-20
Adding missing defaults to the --help output

tharanga:
Issue: #390

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -7543,13 +7543,25 @@ static void clock_handler(const int fd, const short which, void *arg) {
     }
 }
 
+static const char* flag_enabled_disabled(bool flag) {
+    return (flag ? "enabled" : "disabled");
+}
+
+static void verify_default(const char* param, bool condition) {
+    if (!condition) {
+        printf("Default value of [%s] has changed."
+            " Modify the help text and default value check.\n", param);
+        exit(EXIT_FAILURE);
+    }
+}
+
 static void usage(void) {
     printf(PACKAGE " " VERSION "\n");
-    printf("-p, --port=<num>          TCP port to listen on (default: 11211)\n"
-           "-U, --udp-port=<num>      UDP port to listen on (default: 0, off)\n"
+    printf("-p, --port=<num>          TCP port to listen on (default: %d)\n"
+           "-U, --udp-port=<num>      UDP port to listen on (default: %d, off)\n"
            "-s, --unix-socket=<file>  UNIX socket to listen on (disables network support)\n"
            "-A, --enable-shutdown     enable ascii \"shutdown\" command\n"
-           "-a, --unix-mask=<mask>    access mask for UNIX socket, in octal (default: 0700)\n"
+           "-a, --unix-mask=<mask>    access mask for UNIX socket, in octal (default: %o)\n"
            "-l, --listen=<addr>       interface to listen on (default: INADDR_ANY)\n"
 #ifdef TLS
            "                          if TLS/SSL is enabled, 'notls' prefix can be used to\n"
@@ -7558,9 +7570,9 @@ static void usage(void) {
            "-d, --daemon              run as a daemon\n"
            "-r, --enable-coredumps    maximize core file limit\n"
            "-u, --user=<user>         assume identity of <username> (only when run as root)\n"
-           "-m, --memory-limit=<num>  item memory in megabytes (default: 64 MB)\n"
+           "-m, --memory-limit=<num>  item memory in megabytes (default: %ld)\n"
            "-M, --disable-evictions   return error on memory exhausted instead of evicting\n"
-           "-c, --conn-limit=<num>    max simultaneous connections (default: 1024)\n"
+           "-c, --conn-limit=<num>    max simultaneous connections (default: %d)\n"
            "-k, --lock-memory         lock down all paged memory\n"
            "-v, --verbose             verbose (print errors/warnings while in event loop)\n"
            "-vv                       very verbose (also print client commands/responses)\n"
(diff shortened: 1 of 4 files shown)


dormando:
The defaults shouldn't be in two places; need to move to define's and then be used in the help section, else they're either going to drift or just be more work to maintain.

A few exceptions, such as ext_item_age, should be listed in a human readable format, ie "no age limit" vs INT_MAX

tharanga:
> The defaults shouldn't be in two places; need to move to define's and then be used in the help section, else they're either going to drift or just be more work to maintain.
I started with replacing all defaults in the help message with their default settings values, so changing values will not make the help message out of sync. However, there are some instances more help text has added based on the default value; hence abandoned that approach. Not sure adding asserts along with the help message is a good idea for those instances. We could also do the best effort and leave the rest like ext_item_age. 
> 
> A few exceptions, such as ext_item_age, should be listed in a human readable format, ie "no age limit" vs INT_MAX

dormando:
Not sure I follow what you mean by "more help text has added based on the default value"

tharanga:
For example:
- lru_crawler_tocrawl: max items to crawl per slab per run
                          default is 0 (unlimited)
Default value has a special meaning in this context. I initially modified this with an assert, so anyone who changes the default value will be forced to change the description, not sure whether that's a good idea though.

memcached.c
```diff
@@ -6419,78 +6419,91 @@ static void usage(void) {
...
+           "   - ext_threads:         number of IO threads to run. (default: 1)\n"
+           "   - ext_item_size:       store items larger than this (bytes; default: 512)\n"
+           "   - ext_item_age:        store items idle at least this long (default: 4294967295)\n"
+           "   - ext_low_ttl:         consider TTLs lower than this specially (default: 0)\n"
            "   - ext_drop_unread:     don't re-write unread values during compaction\n"
-           "   - ext_recache_rate:    recache an item every N accesses\n"
+           "                          (default: disabled)\n"
+           "   - ext_recache_rate:    recache an item every N accesses (default: 2000)\n"
            "   - ext_compact_under:   compact when fewer than this many free pages\n"
+           "                          (default: total storage / (ext_page_size * 4) )\n"
```
tharanga: Is it alright to be descriptive for a default like this? There's a danger that this could go out of sync with the actual logic.

dormando:
> For example:
> 
> * lru_crawler_tocrawl: max items to crawl per slab per run
>   default is 0 (unlimited)
>   Default value has a special meaning in this context. I initially modified this with an assert, so anyone who changes the default value will be forced to change the description, not sure whether that's a good idea though.

No that's like the item_age one. Should just be a handful of exceptions like that though?

memcached.c
dormando: I think so, though maybe "1/4th of the assigned storage" is more clear.

tharanga:
Yes, it is like item_age and there's only a few of them. Let me change others.

tharanga:
I changed the approach and almost finished changes. Let me revive the branch and update the PR.

dormando:
This might get nuked by an upcoming code reorg, in case you can find some time to wrap it up

memcached.c
```diff
@@ -7601,76 +7608,104 @@ static void usage(void) {
...
-           "                          table should be. normally grows at runtime.\n"
+           "                          table should be. normally grows at runtime. (default starts at: %d)\n"
            "                          set based on \"STAT hash_power_level\"\n"
            "   - tail_repair_time:    time in seconds for how long to wait before\n"
            "                          forcefully killing LRU tail item.\n"
            "                          disabled by default; very dangerous option.\n"
            "   - hash_algorithm:      the hash table algorithm\n"
            "                          default is murmur3 hash. options: jenkins, murmur3\n"
-           "   - lru_crawler:         enable LRU Crawler background thread\n"
+           "   - lru_crawler:         enable LRU Crawler background thread \n"
```
tharanga: lru crawler and maintainer threads are enabled by default.

dormando:
not sure if you mean for me to check this yet :P please say so clearly when ready.

tharanga:
It's ready to review. I might be missing a couple things, but appreciate if you can verify the approach (e.g. asserts for "explained defaults").

memcached.c
dormando: Most (all?) boolean options like this have a "no_" counterpart. could just change this to "no_lru_crawler: disable etc"

memcached.c
```diff
@@ -8422,6 +8457,8 @@ int main (int argc, char **argv) {
 
     /* init settings */
     settings_init();
+    assert(hash_type == MURMUR3_HASH);// if this assertion fails, change the default hash_algorithm
```
dormando: why is this here when the other asserts are up in the help text?

related; do any tests actually call -h explicitly? A bunch of the supports_* stuff does, but perhaps this could use a singular test for "do the help asserts fire"?

dormando:
looks pretty good I think. added a couple comments, and I think there's an errand formatting issue here or there.

I think the asserts are okay, just make sure they're close to each other and maybe add the explicit test. probably worth doing it in 00-startup so we don't get mysterious test failures later on.

memcached.c
tharanga: `hash_type` is local to the main method, it's not a part of the settings unless we refactor it.
+1 on a test, I will add one.

dormando:
Neat. looks like it makes sense. anything else you want to do here?

tharanga:
Nope. I will squash if all is good.

dormando:
go ahead and squash. I'll make a final review later and fix anything I've missed myself.

dormando:
shoot. was doing one last review and started fixing some things, but there's one big pattern that I'd like you to fix:

finally realized a bunch of spots have text like: `(default: 64mb)` or `(default: 8kb)`, etc. but the actual commandline arguments don't take numbers like that, which will confuse users. There were some existing cases but it was separated by a space so if a user input `-m 64 MB` they'd get an error or it wouldn't do anything, but they might try to input `-m 8g` which currently starts (ugh) without any warning (ugh), but gives you 8 megabytes of RAM.

`-I` can take suffixes of 'm', 'k', etc, but not kb/mb/etc. looks like a one character type in the original text too.

in all cases the text says "size in kilobytes/megabytes/etc where etc", so just change the `(default: 8kb)` and similar to `(default: 8)`.

Thanks!

dormando:
@tharanga wanna give this one last set of edits? :) going to be ripping apart memcached.c shortly. want to get this in ahead of it.

tharanga:
@dormando we can squash if this looks good.

dormando:
@tharanga looks good! go ahead and squash. I'll merge it soon.

dormando:
out in 1.5.21, thanks!

PR #498, opened 2019-05-25
Fix segfault when an item's flags are 0

matthewshafer:
If flags are 0 ```item_make_header``` sets nsuffix to 0 (to save space), otherwise it sets it to ```sizeof(flags)```. In the case where nsuffix was 0 flags were still being memcpy'd which in our case was leading to segfaults.

I've changed it so the memcpy of flags should only happen when the space for them has been accounted for. This should be like the behavior in 1.5.14.

### DIFF
--- a/items.c
+++ b/items.c
@@ -330,7 +330,9 @@ item *do_item_alloc(char *key, const size_t nkey, const unsigned int flags,
     it->nbytes = nbytes;
     memcpy(ITEM_key(it), key, nkey);
     it->exptime = exptime;
-    memcpy(ITEM_suffix(it), &flags, sizeof(flags));
+    if (nsuffix > 0) {
+        memcpy(ITEM_suffix(it), &flags, sizeof(flags));
+    }
 
     /* Initialize internal chunk. */
     if (it->it_flags & ITEM_CHUNKED) {


dormando:
hmmmm... Trying to figure out how I missed that. How/where were you seeing segfaults? I need to update the tests with this. Most my testing for the last few weeks (with metaget code) have been with 0 sized flags, but I'm not getting segfaults or else that would've never gone into the release.

dormando:
ah I see how I screwed that up in the removal... trying to understand how this broke anything still.

ITEM_suffix() points at the area just past CAS in the item. The flags would be copied over data space, so the only way this would cause an issue is ... if you're using 0-3 byte sized items which align exactly with the slab class.

Which should be testable. I'll cut another release ASAP, I apologize for the gaffe :(

dormando:
Well, thanks for the patch! Guess when you don't have code review and hack while sick sometimes shit happens. I just re-skimmed the related commits and that seems to be the only bug.

matthewshafer:
We noticed it in the test suite for one of our applications. When running through memcached-debug we saw this assertion failure.

```
Assertion failed: (p->sl_curr == 0 || ((item *)p->slots)->slabs_clsid == 0), function do_slabs_alloc, file slabs.c, line 345.
Abort trap: 6
```

Sorry I wasn't able to come up with an easy way to reproduce it.

Thanks for responding so quickly!

dormando:
Yeah I can think of ways but it's painful. With some reworking I'm doing soon it'll be easier to test redline issues... but that's a few months out still.

Regardless, I try pretty damn hard to not break anything so this is sad but... is what it is.

1.5.16 is up now with this fix, so closing this out.

matthewshafer:
Thanks for the new release, that was super quick (wasn't expecting it so quickly)! Also thanks for all the amazing work you've put into memcached.

PR #499, opened 2019-05-27
memory leak and possible ptr dereference

devnexen:


dormando:
Thanks! Fails whitespace test though. Mind fixing?

dormando:
Released in 1.5.17!

PR #507, opened 2019-07-12
drop privileges, FreeBSD.

devnexen:
Expand on sandboxing support, assuming memcached
does not support end of life versions (available
since FreeBSD 10.x ~2014).

### DIFF
--- a/freebsd_priv.c
+++ b/freebsd_priv.c
@@ -11,6 +11,33 @@
  * in FreeBSD vocabulary.
  */
 void drop_privileges() {
+    cap_rights_t wd, rd;
+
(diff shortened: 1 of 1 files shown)


devnexen:
So after rebasing and tried a bit with extstore it seems to be able to work. What would cause an issue ?

dormando:
Just so long as you can build it with extstore and all of the extstore tests run and pass that's good enough for me. just hard for me to test freebsd right now.

dormando:
(it looked like some of those privs would ban opening/reading files which extstore does)

dormando:
out in 1.5.19! There're some other FreeBSD Issues that could use attention if you're feeling up to it :)

PR #509, opened 2019-07-18
extstore: emulate pread(v) for macOS

gjasny:


dormando:
nice, thanks! I don't have a mac anymore so it's hard test things :(

dormando:
Released in 1.5.17!

PR #510, opened 2019-07-26
move mem_requested from slabs.c to items.c

dormando:
mem_requested is an oddball counter: it's the total number of bytes
"actually requested" from the slab's caller. It's mainly used for a
stats counter, alerting the user that the slab factor may not be
efficient if the gap between total_chunks * chunk_size - mem_requested
is large.

However, since chunked items were added it's _also_ used to help the
LRU balance itself. The total number of bytes used in the class vs the
total number of bytes in a sub-LRU is used to judge whether to move
items between sub-LRU's.

This is a layer violation; forcing slabs.c to know more about how items
work, as well as EXTSTORE for calculating item sizes from headers.
Further, it turns out it wasn't necessary for item allocation: if we
need to evict an item we _always_ pull from COLD_LRU or force a move
from HOT_LRU. So the total doesn't matter.

The total does matter in the LRU maintainer background thread. However,
this thread caches mem_requested to avoid hitting the slab lock too
frequently. Since sizes_bytes[] within items.c is generally redundant
with mem_requested, we now total sizes_bytes[] from each sub-LRU before
starting a batch of LRU juggles.

This simplifies the code a bit, reduces the layer violations in slabs.c
slightly, and actually speeds up some hot paths as a number of branches
and operations are removed completely.

This also fixes an issue I was having with the restartable memory
branch :) recalculating p->requested and keeping a clean API is painful
and slow.

---

Leaving this self-PR up overnight before merging in case I think of some reason why this doesn't work. This does have an unusual move of changing the existing stats output... The counter moves from `stats slabs` to `stats items`... preserving the slabs output is obnoxious so I'm thinking of just doing this relatively small change.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -900,6 +900,7 @@ number_temp            Number of items presently stored in the TEMPORARY LRU.
 age_hot                Age of the oldest item in HOT LRU.
 age_warm               Age of the oldest item in WARM LRU.
 age                    Age of the oldest item in the LRU.
+mem_requested          Number of bytes requested to be stored in this LRU[*]
 evicted                Number of times an item had to be evicted from the LRU
                        before it expired.
 evicted_nonzero        Number of times an item which had an explicit expire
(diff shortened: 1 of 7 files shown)


dormando:
merged

Issue #520, opened 2019-08-14
Log connection id of clients with fetchers and mutations


tharanga:
It's helpful to have client's connection id with fetchers and mutations. This way we can monitor individual clients when needed.

PR #521, opened 2019-08-14
log client connection id with fetchers and mutations

tharanga:
you can now monitor fetch and mutations of a given client
Issue : #520

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -186,7 +186,7 @@ static int _logger_thread_parse_ise(logentry *e, char *scratch) {
 
     uriencode(le->key, keybuf, le->nkey, KEY_MAX_URI_ENCODED_LENGTH);
     total = snprintf(scratch, LOGGER_PARSE_SCRATCH,
-            "ts=%d.%d gid=%llu type=item_store key=%s status=%s cmd=%s ttl=%u clsid=%u connid=%d\n",
+            "ts=%d.%d gid=%llu type=item_store key=%s status=%s cmd=%s ttl=%u clsid=%u cfd=%d\n",
             (int)e->tv.tv_sec, (int)e->tv.tv_usec, (unsigned long long) e->gid,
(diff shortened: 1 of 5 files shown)


dormando:
Thanks! No need to open a separate issue unless you expect a ton of discussion.

Sad that you had to modify so many different parts of the logger. I tried pretty hard to get a good API for it :(

logger.c
```diff
@@ -653,17 +653,18 @@ static void _logger_log_ext_write(logentry *e, item *it, uint8_t bucket) {
  * TODO: This and below should track and reprint the client fd.
```
tharanga: @dormando we can remove this TODO if it means this change.

tharanga:
Ack on not opening a separate issue for straightforward fixes.

logger.c
dormando: Yeah think so.

logger.c
```diff
@@ -186,9 +186,9 @@ static int _logger_thread_parse_ise(logentry *e, char *scratch) {
 
     uriencode(le->key, keybuf, le->nkey, KEY_MAX_URI_ENCODED_LENGTH);
     total = snprintf(scratch, LOGGER_PARSE_SCRATCH,
-            "ts=%d.%d gid=%llu type=item_store key=%s status=%s cmd=%s ttl=%u clsid=%u\n",
+            "ts=%d.%d gid=%llu type=item_store key=%s status=%s cmd=%s ttl=%u clsid=%u connid=%d\n",
```
dormando: s/connid/cfd/ or cid? it's the actual file descriptor; either way connid is a bit long considering 'gid' and 'clsid' are shortened.

dormando:
this is clear enough on a read. I was just too lazy to do it before :) thanks for the tests too.

Remove the comment and shorten connid to cfd and I'll merge it.

dormando:
Released in 1.5.17!

PR #523, opened 2019-08-15
add server address to the "stats conns" output

tharanga:
this is helpful when it's required to identify which clients are
connected to which server address when memcached listens on
multiple addresses


dormando:
Thanks! Think this is reasonable. I'll give it a closer look when I can.

dormando:
Ah shoot; could you rebase this? One of the changes in 'next' caused some conflicts (the PR here is showing it as well, I just noticed it now). I can fix if I don't hear back in time for the cut tomorrow.

tharanga:
On it.

dormando:
Released in 1.5.17!

PR #524, opened 2019-08-16
Faster slab reassignment

byrnedj:
Sometimes the slab reassignment mechanism gets stuck on busy items. 

We have found as workloads scale and
applications use increasing amounts of memory, the current
reassignment mechanism in memcached is inefficient. In fact,
we measure that reassignments can take millions of requests
to complete a single reassignment.

Motivated by these findings, we introduce a faster slab re-
assignment mechanism. In our experiments, we show that the
time needed to reassign a slab reduces by over 99% resulting
in the ability to reach workloads’ steady state miss ratio by
53% to 75% faster. 

See the attached figures for a sample experiment using a workload that introduces a change to the item size distribution. 
[p2-2app-100-25.384.alloc.speed.both.pdf](https://github.com/memcached/memcached/files/3510984/p2-2app-100-25.384.alloc.speed.both.pdf)
[p2-2app-100-25.384.speed.mr.both.pdf](https://github.com/memcached/memcached/files/3510985/p2-2app-100-25.384.speed.mr.both.pdf)


dormando:
Neat, thanks! Looks like it doesn't compile though.

Been meaning to optimize this for a long time but never had a complaint/time/motivation :) Fair warning though; validating these sorts of changes for merge is pretty effort intensive so the PR could sit for a while before being merged.

dormando:
(and if you're curious; the torture suites are here: https://github.com/memcached/mc-crusher) under conf/ + slab-tosser scripts. I'll probably just write a test suite similar to the files under test-suites and let it run for a week.

While developing the feature initially those suites found a handful of obscure crashes/corruptions, sometimes after running for hours.

byrnedj:
Fixed the compilation issue. Currently, this PR doesn't include chunked objects. I am running the tests on my setup here too.

dormando:
Ok. Please let me know when it's feature complete, I guess :) More tests are appreciated. Looks like you're already developing against the 'next' branch, which is perfect since some slab reassign changes are queued in there.

byrnedj:
Okay, I added chunk support. Still running tests on my end no issues so far.

byrnedj:
Have been running the mc-crusher all last week. No issues on my end and it
passes all tests.

Take a closer look, I’d love to see this in the next release.

byrnedj:
Thanks for taking a closer look at this.

>I see you only changed code for the cases of the rebalancer wanting to
move memory but without saving the item

This was this case that was slowing down the slab reassignment the most,
since the *wasBusy *counter would be incremented causing the slab
rebalancer to sleep (1000ms).

>I believe both of those cases are atypical for most workloads, or they're
supposed to be anyway. Do you have more context for the test you made?

In our environment, memcached reassigns slabs across multiple classes. When
we are removing slabs from one class to another, it is almost always the
case that the class which we are removing slabs from is already full. We
are often under tight memory bounds (~4GB). The workload used to test this
is a synthetic workload based on access patterns from Facebook and other
web caches. Zipfian (alpha=1-0.25) and about 10 million unique objects.

>(1) Your change seems to guard against refcount not actually reaching zero

I introduced this to make sure there were no outstanding references to the
item in other threads before
we unlink and free. If this is mistaken, let me know and I'll take care of
it.

>(2) it looks like I could just (after unlink) lock the slabber, call
do_slabs_free() and then immediately _cut_free() and unlock

That is all that I am trying to do :)

>(3) I figured the big speedup here would be using a bit vector to mark
completed chunk offsets so we stop flipping the locks around and re-looking
at every item when revisiting. Was that also part of your change?

Once we cut it from the freelist, we could mark the offset, skip it and
move on to the next item? I believe that would speed things up too. Again,
the biggest problem we ran into was having to wait after every *busy item *in
case (2) that you mentioned.

>(4) Unfortunately there was an actual reason why I let the slab lock go
for the duration of that MOVE_FROM_LRU code

I did not have any measurable drop in latency or throughput in my test
setup. We use mutilate and run the workload through there. I'll try running
through mc-crusher too.

Thinking about it now, we should only need to lock from do_slabs_free to
cut_from_freelist since reference count would go from >1 to >=1.

If you think its worth the time, I'll take some time to implement the bit
vector to check for chunks that have already been cut from the freelist.
Second, I can tighten the locks by being more precise with the reference
count and possibly clean up re: your Q1

byrnedj:
Thanks. Was on vacation will get back to this as I catch up this week. As an note, I confused the usleep with 1000ms rather than 1000us. 

>At every step I've written this thing super conservatively; don't hammer the locks, sleep, retries, etc. The best generic speedup you'll get in the short term is a bit vector + changing the usleep to be a retry backoff (maxing out at 1ms). Usually the vast majority of a slab will quickly free, and in a pessimistic case where it has to unlink everything without a rescue it shouldn't sleep for the first try. A bit vector will protect from any hard-looping cases. (use it to count which chunk it's on, rather than remembering offsets for safety).

The retry backoff is a good solution, but it still fails to move the most amount of slabs/second. See the table below from our tests. CPU usage was measured when the slab rebalancer thread was active. The higher usage is a result of not waiting in `usleep`.

| algorithm | sleep interval | cpu usage \% | slabs/s |
|-----------|----------------|--------------|---------|
| base      | 1              | 13.8         | 36.93   |
| base      | 10             | 10.6         | 35.5    |
| base      | 100            | 8.0          | 26.55   |
| base      | 1000           | 3.1          | 4.12    |
| fast      | 1000           | 97.0         | 252.31  |

Also, we still need to remove items from the slab's freelist right away if we are not going to rescue them. If we don't do this then we run into the problem of those slots getting realloc'd. 

What I propose, is the following:
1. Immediately cut non-rescued items from the slab's freelist (I know this results in essentially random eviction, but a better solution may take more time). To get around the lock issue, only unlock the slabs lock if we pass the `if (save_item)`. That avoids unlocking and relocking the slabs lock.
2. Add the retry backoff timer for the items which are busy.

dormando:
Thanks for the update; I'll respond in detail when I can. You're right that your particular benchmark won't see as much improvement with my initial suggestions.. in most cases if memory needs to move from A to B there's a lessening pressure on A. Your test isn't like that.

dormando:
So I might have a simplification... I think the entire slabs_free() bit in the original unlink code might be pointless after my recent patch to remove the mem_requested counter. All cut_free() does is make sure it's not linked in the freelist. slabs_free() used to maintain mem_requested, and the slab mover also had to adjust it a bit... but that's all gone now.

So in the non-rescue case after the unlink, the slab chunk is owned by the mover thread, not the linked list or the slabber...  so you can do the MOVE_FROM_SLAB routine and set refcount/flags and not tick was_busy. Just deleting the busy_items++, was_busy++, slabs_free() calls and adding an it_flags = ITEM_SLABBED|ITEM_FETCHED line should give you your original speedup?

From there we still need the vector and backoff to speed it up for the general use case.

... seem sane? unless I've misread something?

dormando:
Oh, lost half my thought; *if the item wasn't chunked*, elide that code. If it was chunked we have a much more complicated free routine, so it should I'd be much happier just letting it loop as-is. The chunked item cases typically work with very large data (512k+, in the future cutting down to 128k or 256k or so) so having a fast retry time will cover that case pretty well as there're only 2-4 chunks in the entire slab.

byrnedj:
Hi, I finally got around to making the changes. See #542 for the updated PR (I had to switch from my master branch to your next branch to start clean).

dormando:
For posterity: #608 fixes a critical bug in this PR + a bug with the completed routine.

PR #525, opened 2019-08-16
add close_conn command to close a given client connection

tharanga:
helps to evict zombie or malicious clients in real time.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -502,9 +502,8 @@ static const char *prot_text(enum protocol prot) {
     return rv;
 }
 
-void close_connection(conn *c, bool is_idle_connection) {
-    if (is_idle_connection &&
-        settings.idle_timeout > 0 &&
+void close_idle_connection(conn *c) {
+    if (settings.idle_timeout > 0 &&
         (current_time - c->last_cmd_time) > settings.idle_timeout) {
         if (c->state != conn_new_cmd && c->state != conn_read) {
             if (settings.verbose > 1)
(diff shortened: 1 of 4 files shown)


tharanga:
We can put this behind a config parameter if this is not widely applicable (or not desired in some situations).

memcached.c
```diff
@@ -3633,6 +3633,31 @@ static void process_stat(conn *c, token_t *tokens, const size_t ntokens) {
...
+    if (conn == NULL) {
+        out_string(c, "ERROR Connection not found");
+    } else if (conn->state == conn_listening) {
+        out_string(c, "ERROR Listening connection");
+    } else if (IS_UDP(conn->transport) && conn->state == conn_read) {
+        out_string(c, "ERROR UDP connection");
+    } else if (conn->sfd == c->sfd) {
+        out_string(c, "ERROR Current connection");
+    } else {
+        conn_close(conn);
```
tharanga: Should we close the connections asynchronously?

dormando: haven't looked closely yet; if you do it the same way the idle connection kicker works it should be safe. IIRC that was a signal back to the owner thread to close it out?

tharanga: Yes it is. Let me do it that way.

dormando:
Kind of raises a question I should've asked already; what's the use case for this that the idle thread doesn't do for you? Are you finding conns stuck in other states?

tharanga:
This is expected to help on evicting zombie or malicious/buggy clients in real time. Idle timeout helps zombies when the specified timeout is reached.

dormando:
what're the connections doing though? Stuck in a loop fetching an item, stuck halfway through an upload, idle, etc?

tharanga:
Some connections are idle, clients keep on re-connecting and leave connections behind. These will eventually get cleaned up by the idle timeout, but until then they pile up. Some clients are abusive, they are active and we want to abort them.
I looked at the state machine and realized we should implement a way to interrupt the state machine and let it close connections. Closing a connection while a worker is serving could be ugly, specially on TLS.
For the moment we'll use iptables to block such clients. I will close this PR and re-open it if there's a clean way to close connections.

PR #527, opened 2019-08-19
add error handling when calling dup function

minkikim89:
The dup function can return errors.
So I added error handling code.


dormando:
Released in 1.5.17!

PR #529, opened 2019-09-03
Remove a wrong function comment

minkikim89:
The function doesn't use 'key' parameter anymore.


dormando:
thanks, nice catch! trivial documentation fixes like these are committed via the maintainers (I should make this policy explicit in the github PR form thingie...). Fix is upstream now.

PR #532, opened 2019-09-15
Make get_cas_id use an atomic counter instead of a lock.

en4bz:


dormando:
Thanks!

This fix is actually a bit more complex. You can see what I had to do in the logger.c file; I wanted to make those bits generic and reuse them for logger and cas.

I'm holding off on that change right now until I get time to solve a PA-RISC build problem, since that may be related to the use of atomics.

en4bz:
Is there a support matrix available? Based on what I read it seems like x86,arm64 are the primary targets with OS support across Linux,BSD,Solaris on gcc/clang? Also seems bizarre to support hardware that can no longer be purchased from a first party and was supplanted by another architecture (ia64) which is also EOL.

dormando:
I don't really have a support matrix, but if I get requests about a platform (and it's definitely not a one off), and fixing it isn't a huge deal, I'm not bothered by the fix. I have a buildbot network that does most of the supported distros/platforms but I haven't been able to expose that publicly since the last major buildbot revision.

Weirdly, recently, someone reached out about PA-RISC.

That said, Linux/BSD are primary. OS X is in there. Solaris because we have support and I try to not break it, x86/x86-64/arm32/arm64 are the most common platforms. POWER tends to be covered by arm64 but is also supported.

Also, when a platform bug shows up it's often a legitimate oversight. I haven't had one show up that turned into a major problem.

PR #533, opened 2019-09-18
Make memcached setgroups failure non-fatal

jwbee:
A call to setgroups can fail on Linux if the user lacks cap_setgid or if setgroups has been disallowed in a user namespace. The latter happens when the namespace has no GID map or if /proc/pid/setgroups has been changed to "deny". Treating this failure as fatal means that memcached cannot start in a bazel fakeroot. Ignoring failure seems harmless.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -8475,7 +8475,6 @@ int main (int argc, char **argv) {
         }
         if (setgroups(0, NULL) < 0) {
             fprintf(stderr, "failed to drop supplementary groups\n");
-            exit(EX_OSERR);
         }
         if (setgid(pw->pw_gid) < 0 || setuid(pw->pw_uid) < 0) {
             fprintf(stderr, "failed to assume identity of user %s\n", username);
--- a/memcached.c
+++ b/memcached.c
@@ -8474,7 +8474,18 @@ int main (int argc, char **argv) {
             exit(EX_NOUSER);
         }
         if (setgroups(0, NULL) < 0) {
-            fprintf(stderr, "failed to drop supplementary groups\n");
+            /* setgroups may fail with EPERM, indicating we are already in a
+             * minimally-privileged state. In that case we continue. For all
+             * other failure codes we exit.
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! will take a look at this.

dormando:
Trying to work this through: it's always been a fatal error because this may have meant we were unable to drop root fully... If you don't have the setgid cap, does mc still need to start as root for you? I thought most people just started it with a user account in that case.

What might be best here is to handle an EPERM error and warn but not exit, since there're other cases where it should be fatal (invalid group ids/etc)?

jwbee:
Currently memcached exits because it lacks adequate privileges to drop its privileges, which is amusing up to a point. The only error we can get from setgroups(NULL, 0) is EPERM because 0 is not greater than NGROUPS_MAX and an empty list is not likely to trigger ENOMEM, either.  I'd be happy to handle those cases if you like, and only continue in case of EPERM.

You're right this can only happen if you started as UID 0, but this can happen in a user namespace without UID 0 really being root. You're also right about cap_setgid and if you don't have it memcached will still exit on the next line of code unless you pass -u root.

Sorry if that was confusing.  The short version is that today if you are root and you use -u root memcached may still fail due to setgroups()=EPERM.

jwbee:
By the way this hasn't always been a fatal condition for memcache because the call in question was not added until about a year ago.

dormando:
It was added for good reason though... I'd be happier if it exited if the errno wasn't EPERM, so we're at least being explicit here. Should just be a simple check, warn, fail on all else.

Still not super comfortable with how that came to be.

jwbee:
PTAL.

dormando:
Looks great, thanks! Bonus points if you squash it but I can also do that when I swing around to merge.

dormando:
I did a squash/merge/push. This is now in the next branch. thanks!

dormando:
out in 1.5.19!

PR #537, opened 2019-09-19
restart: add missing msync at close time

pbalcer:
To make sure that the backing pages actually make it to the backing
device prior to shutdown, munmap must be preceded by an msync call.
Additionally, since the metadata file is used as a flag to check if
the shutdown was graceful, the data needs to be sync'd prior to that
file being written.

### DIFF
--- a/restart.c
+++ b/restart.c
@@ -303,6 +303,8 @@ bool restart_mmap_open(const size_t limit, const char *file, void **mem_base) {
 
 /* Gracefully stop/close the shared memory segment */
 void restart_mmap_close(void) {
+    msync(mmap_base, slabmem_limit, MS_SYNC);
+
     if (restart_save(memory_file) != 0) {
         fprintf(stderr, "[restart] failed to save metadata");
     }


dormando:
Thanks, I'll take a look; this isn't typically needed since you're supposed to run this out of a ramdisk or DAX mount, which has no buffer to flush. Or at least, I was under that impression from the documentation :)

Should still add this since people will invariably run it off of a disk and expect it to work, I guess.

dormando:
Alright merged for next release. I had to change the unit tests to run out of /tmp so this is good to have to ensure they're reliable. So far as I recall the code paths for shared mem and mmap'ed tmpfs files is exactly the same, but this patch has been open since 2017 and I no longer have a clear memory of what I'd looked up... So either way, this is good.

Thanks again! I usually leave PR's open until the release cut.

andyrudoff:
A clarification (just for the record, since you're already accepting the change):  you need msync (or fsync) to make things persistent even with DAX.  For traditional file systems, msync will flush the page cache.  For DAX, there is no page cache but dirty data might still be sitting in the CPU caches.  msync detects the DAX case and uses flush instructions like CLWB on x86 to make sure the stores are persistent.

dormando:
Makes sense, thanks! I completely missed that.

dormando:
(I don't actually have any DAX hardware so all the restartable work has been in tmpfs, but the access-items-via-mmap was extensively tested on an optane machine... so that at least worked)

pbalcer:
The fact that there's no page cache buffer to flush is an implementation detail of the underlying file system. POSIX (or more specifically, Linux kernel) still requires you to call an msync if you expect the data to be actually present in the file at a defined time.
Even if that weren't true, you still need to take care of ordering to ensure that you can distinguish between graceful and ungraceful shutdowns. If I understand correctly, your mechanism uses the metadata file consistency to decide if the cache should be resumed or started from scratch, and since your current code first writes the metadata and then lets the kernel flush the file whenever it wants, you might end up in a situation with a valid meta file and invalid data file.

dormando:
out in 1.5.19!

PR #539, opened 2019-09-20
TLS leak

dormando:
frees said context. Don't believe SSL_shutdown is required? Should
double check.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -5805,6 +5805,7 @@ static void drive_machine(conn *c) {
                             if (settings.verbose) {
                                 fprintf(stderr, "SSL connection failed with error code : %d : %s\n", err, strerror(errno));
                             }
+                            SSL_free(ssl);
                             close(sfd);
                             break;
                         }
--- a/thread.c
+++ b/thread.c
@@ -500,6 +500,12 @@ static void thread_libevent_process(int fd, short which, void *arg) {
                             fprintf(stderr, "Can't listen for events on fd %d\n",
                                 item->sfd);
                         }
+#ifdef TLS
+                        if (item->ssl) {
+                            SSL_shutdown(item->ssl);
+                            SSL_free(item->ssl);
+                        }
+#endif
                         close(item->sfd);
                     }
                 } else {


memcached.c
```diff
@@ -5805,6 +5805,7 @@ static void drive_machine(conn *c) {
                             if (settings.verbose) {
                                 fprintf(stderr, "SSL connection failed with error code : %d : %s\n", err, strerror(errno));
                             }
+                            SSL_free(ssl);
```
tharanga: +1 @dormando.
This should be :+1: 
`
if (ssl) {
                                SSL_shutdown(ssl);
                                SSL_free(ssl);
                            }
`
Also in the `conn_free`, probably better as one static method in the `tls.c` and used in all 3 places.

dormando: if (ssl == NULL) check is directly above so there shouldn't need to be another check. Looking at the docs I'm not sure SSL_shutdown is needed either, but maybe it can't hurt? Also a function call won't help as there's no connection structure in this case :)

tharanga: Yes, null check is repetitive. We have to do this in the `conn_free` as well. A method call can be something like `SSL_close(SSL *ssl)` as we always have a `SSL` pointer.

tharanga: Right, `SSL_shutdown` is not needed here.
From the man pages:
`Note that SSL_shutdown() must not be called if a previous fatal error has occurred on a connection i.e. if SSL_get_error() has returned SSL_ERROR_SYSCALL or SSL_ERROR_SSL.`, which is the case here. Only needed in `conn_free`. No need of another method.

dormando: Cool yeah. I think this is good as is then. Will wait for the reporter to test it before merging, but not planning on making a release right this moment anyway.

tharanga: Yes. Don't we need the same in `conn_free` method? If so we need `SSL_shutdown` there.

LINKIWI:
Thanks all for looking into this. I'll deploy this patch to our dark traffic cluster and will report back with results.

memcached.c
dormando: I think so yeah. It does call conn_free() with an SSL context. I'll update this patch to cover it. Since it's a malloc failure I'm dubious that code's ever been called though...

conn_new() for a client connection is only ever called in thread.c, which checks for a null response and closes the client socket. Check either goes there or in conn_free(). I'm leaning toward the dispatch call since that already owns closing the socket, though that feels a bit off.

dormando: Yeah I'll add it in the dispatch handler. We also get a null return if event_add() fails for some reason, which doesn't call conn_free().

LINKIWI:
I can confirm that, with this patch, I'm not observing the same pattern of unbounded memory increase on SSL connection churning. Thanks @tharanga @dormando for your help in tracing down this issue! I'll let this continue serving shadow traffic over the weekend and will report back if there's any anomalous behavior over a couple more days.

dormando:
@LINKIWI Thanks for testing it out! I'll probably do another release cut next week, giving a few more days for bug reports to roll in, and for you to test over the weekend.

dormando:
@LINKIWI how goes? Hoping we're clear to merge this :)

LINKIWI:
Yep! The cluster has been serving dark traffic for the entire weekend with this patch and total memory usage has been constant, and no other problems have emerged. I think we are good to merge.

dormando:
out in 1.5.19!

PR #540, opened 2019-09-21
TLS: Reload CA cert in certificate refresh routine

LINKIWI:
With TLS enabled, the current implementation only reloads the certificate and key on a `refresh_certs` line from a client. This patch additionally attempts to reload the CA certificate, if supplied.

Also swap `SSL_LOCK` for `STATS_LOCK` in what I believe was a typo in the original TLS implementation PR.

cc @tharanga

### DIFF
--- a/tls.c
+++ b/tls.c
@@ -107,11 +107,20 @@ static bool load_server_certificates(char **errmsg) {
     } else if (!SSL_CTX_check_private_key(settings.ssl_ctx)) {
         snprintf(error_msg, errmax, "Error validating the certificate\r\n");
         success = false;
-    } else {
-        settings.ssl_last_cert_refresh_time = current_time;
+    } else if (settings.ssl_ca_cert) {
+        if (!SSL_CTX_load_verify_locations(settings.ssl_ctx,
+          settings.ssl_ca_cert, NULL)) {
+            snprintf(error_msg, errmax,
+              "Error loading the CA certificate: %s\r\n", settings.ssl_ca_cert);
+            success = false;
+        } else {
+            SSL_CTX_set_client_CA_list(settings.ssl_ctx,
+              SSL_load_client_CA_file(settings.ssl_ca_cert));
+        }
     }
     SSL_UNLOCK();
     if (success) {
+        settings.ssl_last_cert_refresh_time = current_time;
         free(error_msg);
     } else {
         *errmsg = error_msg;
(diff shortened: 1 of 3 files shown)


dormando:
Thanks! I'll let tharanga deal with this :) Though I think that stats lock should probably just be removed entirely, unless the lock is first added to the places where that value is set, else there's no point in locking it just for that one read.

tls.c
```diff
@@ -107,11 +107,19 @@ static bool load_server_certificates(char **errmsg) {
     } else if (!SSL_CTX_check_private_key(settings.ssl_ctx)) {
         snprintf(error_msg, errmax, "Error validating the certificate\r\n");
         success = false;
-    } else {
-        settings.ssl_last_cert_refresh_time = current_time;
+    } else if (settings.ssl_ca_cert) {
```
tharanga: This is repetitive from https://github.com/memcached/memcached/blob/master/tls.c#L155. Let's remove loading the  CA cert from there and keep it only here, because this method is used in the `ssl_init` as well.

tls.c
```diff
@@ -107,11 +107,19 @@ static bool load_server_certificates(char **errmsg) {
...
+        SSL_CTX_set_client_CA_list(settings.ssl_ctx,
+          SSL_load_client_CA_file(settings.ssl_ca_cert));
+        if (!SSL_CTX_load_verify_locations(settings.ssl_ctx,
```
tharanga: I should have done this first and then do the `SSL_CTX_set_client_CA_list...`. Let's do it like that here.

tls.c
```diff
@@ -107,11 +107,19 @@ static bool load_server_certificates(char **errmsg) {
...
+          settings.ssl_ca_cert, NULL)) {
+            snprintf(error_msg, errmax,
+              "Error loading the CA certificate: %s\r\n", settings.ssl_ca_cert);
+            success = false;
+        }
     }
     SSL_UNLOCK();
     if (success) {
+        settings.ssl_last_cert_refresh_time = current_time;
```
tharanga: While you are at this, you may remove the line https://github.com/memcached/memcached/blob/master/tls.c#L169. Timestamp update is repetitive there.

tharanga:
@dormando is right, we don't need any locks for the `stats`.

tharanga:
A test would be handy, probably you can add one to the `t/ssl_cert_refresh.t`.
Thanks @LINKIWI for fixing this.

LINKIWI:
Thanks for the review. I'll look into updating the cert refresh test as well.

LINKIWI:
@tharanga Mind making another pass through? Thanks

t/lib/MemcachedTest.pm
```diff
@@ -236,7 +237,8 @@ sub new_memcached {
             $conn = eval qq{IO::Socket::SSL->new(PeerAddr => "$host:$port",
                                         SSL_verify_mode => SSL_VERIFY_NONE,
                                         SSL_cert_file => '$client_crt',
-                                        SSL_key_file => '$client_key');
+                                        SSL_key_file => '$client_key',
+                                        SSL_ca_file => '$ca_crt');
```
tharanga: I would not add CA_cert to defaults (when SSL is enabled) as it's an optional parameter. You may use `https://github.com/memcached/memcached/blob/master/t/ssl_verify_modes.t#L17` in `ssl_cert_refresh.t`

tharanga:
@LINKIWI there's a minor comment in tests, rest LGTM.

dormando:
cool, mind squashing?

LINKIWI:
Squashed; thanks for coordinating. Any other merge blockers?

dormando:
Think you're good. I'm going to do another release earlier than normal so it shouldn't be too long.

dormando:
out in 1.5.19!

PR #547, opened 2019-10-01
DTrace build fix

devnexen:
During config step just being "contented" by generating the header,
there is no symbols to attach for, no chance to work as is.
Changing a probe signature, on some platforms, pthread_t is an
opaque type thus casting to a type large enough to hold it
for all oses.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -172,12 +172,11 @@ if test "x$enable_dtrace" = "xyes"; then
   if test "x$DTRACE" != "xno"; then
     AC_DEFINE([ENABLE_DTRACE],1,[Set to nonzero if you want to include DTRACE])
     build_dtrace=yes
-    # DTrace on MacOSX does not use -G option
-    $DTRACE -G -o conftest.$$ -s memcached_dtrace.d 2>/dev/zero
+    $DTRACE -h -o conftest.h -s memcached_dtrace.d 2>/dev/zero
     if test $? -eq 0
     then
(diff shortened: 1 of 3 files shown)


dormando:
Thanks! I don't quite follow the comment here though. Are there any tests or config outputs that could be improved?

devnexen:
Well, in fact if you mention the configure part, the object generation could never work (tested on FreeBSD only), what happened is once the config passed, it attempted to build with the dtrace provider but got linkage issues ... with this change, the dtrace option work just fine.

WSLUser:
Just for FYI, since we have Dtrace support for Windows: https://www.microsoft.com/en-us/download/details.aspx?id=58091, can we check if build failures no longer are due to dtrace with this change? If it does, can fixes be included in this PR?

dormando:
@WSLUser if there's a build problem under (I'm assuming the windows subsystem for linux?) please open a new issue with details of the build failure. I don't have WSL and am unaware of any build issues with it.

WSLUser:
No I'm referring to Windows itself. WSL doesn't have issues.

dormando:
@WSLUser building natively for windows has (in the past) required a gigantic patch. I don't think dtrace was ever the problem for a windows build. A few people have done it as one-offs but 15+ years nobody's volunteered to maintain a windows port. I don't have any interest in doing it myself either.

dormando:
Out in 1.5.20

PR #549, opened 2019-10-04
Pluggable storage architecture

lucagiac81:
Memcached "core" and extstore are decoupled and an interface is defined between them.
- The interface supports all the functionality required by extstore, and it generalizes it to support other engines.
- There is no change in functionality or performance.

The interface gives Memcached the ability to use different storage engines, which can be tailored to specific use cases or media.

Extstore is converted to a storage engine implementing the interface. It is the default engine. No additional engines are provided at this time.

Refer to doc/pluggable-storage.txt for more details.


memcached.c
```diff
@@ -148,10 +147,7 @@ conn **conns;
 struct slab_rebalance slab_rebal;
 volatile int slab_rebalance_signal;
 #ifdef EXTSTORE
-/* hoping this is temporary; I'd prefer to cut globals, but will complete this
- * battle another day.
- */
-void *ext_storage = NULL;
+static storage_engine *engine;
```
dormando: ext_storage was only used in one place in items.c. I'd rather not regress to more globals by removing the worker thread reference and using this more

Issue #550, opened 2019-10-05
Cross-compilation is broken


bkuhls:
Hi,

this commit https://github.com/memcached/memcached/commit/9485db55edf2dc6b75b2024770ae41ad6b68de18 leads to an error during configure:

checking build system type... x86_64-pc-linux-gnu
checking host system type... x86_64-buildroot-linux-musl
...
checking for sasl/sasl.h... no
configure: error: in /home/buildroot/buildroot/output/build/memcached-1.5.19:
configure: error: cannot run test program while cross compiling
See config.log for more details

dormando:
Thanks! What was in config.log? Any more details about your setup?

bkuhls:
https://github.com/memcached/memcached/pull/552 fixed the problem, thanks!

PR #552, opened 2019-10-08
Fix cross compilation

olajep:
AC_RUN_IFELSE does not work when cross-compiling so we need to provide
fallback methods for those cases.

I tried to use constructs that work with Autoconf 2.52.
Alas, I wasn't able to generate a working build system with that version.

Autoconf 2.58 / Automake 1.7.9 is the earliest combo that I could get
to work (with and without this patch).
Perhaps it's time for a slight bump for the required version numbers?

Cross-compiles successfully against:
riscv64-unknown-linux-gnu

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -264,23 +264,42 @@ return sizeof(void*) == 8 ? 0 : 1;
     ],[
       CFLAGS="-m64 $org_cflags"
     ],[
-    AC_MSG_ERROR([Don't know how to build a 64-bit object.])
+      AC_MSG_ERROR([Don't know how to build a 64-bit object.])
+    ],[
+       dnl cross compile
+       AC_MSG_WARN([Assuming no extra CFLAGS are required for cross-compiling 64bit version.])
     ])
 fi
 
 dnl If data pointer is 64bit or not.
-AC_RUN_IFELSE(
-  [AC_LANG_PROGRAM([], [dnl
-return sizeof(void*) == 8 ? 0 : 1;
-  ])
-],[
-  have_64bit_ptr=yes
-],[
+AC_CHECK_HEADERS([stdint.h])
+AS_IF([test -z "$have_64bit_ptr"],
+  [AC_RUN_IFELSE(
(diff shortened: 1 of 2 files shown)


olajep:
Should fix #550

dormando:
neat, thanks! I admit to not being an autowhatnot expert. Looks like your bump versions are from 2003? That should be fine, heh.

I'll come back and look at this soon to make sure I understand what's going. What's your risc-v platform? All I have are mcu's.

dormando:
Also bonus points for some comment warnings about AC_RUN_IFELSE and cross compile.

olajep:
> neat, thanks! I admit to not being an autowhatnot expert. Looks like your bump versions are from 2003? That should be fine, heh.

It was this line that confused me
```m4
AC_PREREQ(2.52)
```
I've seen some projects that require support for old Autoconf versions but this is the most extreme so far :smile: 
Perhaps bump that number to at least 2.58.
After all this is only what's required for building from git.
Release tarballs would still work.

> I'll come back and look at this soon to make sure I understand what's going on. What's your risc-v platform? All I have are mcu's.

Oh, we run Rocket Chip on an FPGA as our research platform. Exploring if we should use memcached to simulate some cloud workloads.

dormando:
Hah! Yeah go ahead and bump that a bit. This project's auto* was written in 2003, which is why an prereq from 2000 is in there. I only touch the autostuff when necessary, so it hasn't exactly had a revisit for a cleanup. I just ran into a code comment about fixing something from FreeBSD 4.x, and another about ubuntu 8.04... old code is interesting sometimes.

That sounds like a fun project! None of my FPGA's are big enough to run linux and boot in the same day :)

olajep:
Ok but that should be in a separate pull.
If you bump to say:
Autoconf 2.61 2006-11-17
Automake 1.10 2006-10-15
you could get rid of m4/c99-backport.m4
No point in just bumping version numbers IMO.

bkuhls:
Tested-by: Bernd Kuhls <bernd.kuhls@t-online.de>

dormando:
Out in 1.5.20

PR #554, opened 2019-10-08
Documentation build fixes

olajep:
Tried to break this into two pulls but the second patch depends on the first.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -1,23 +1,36 @@
 man_MANS = memcached.1
 
-EXTRA_DIST = *.txt
+SPECIFICATIONS = protocol-binary.txt protocol-binary-range.txt
 
-BUILT_SOURCES=
+EXTRA_DIST = new_lru.txt  \
+	     protocol.txt \
+	     readme.txt   \
+	     storage.txt  \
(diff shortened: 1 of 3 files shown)


olajep:
(Force push was converting rfc2629.dtd from CRLF to CR, didn't show up in vimdiff)

dormando:
thanks!

dormando:
No way to automatically get the list of .txt files? There's no way I/others are going to remember to update that list, and new files do occasionally land there.

dormando:
Thanks!

dormando:
I'm releasing this anyway, but this breaks `make dist` if xml2rfc and xsltproc aren't installed. `@XML2RFC@` gets set to 'no' and it tries to run... `no`. If you have some time to fix in another PR, would be appreciated. At least give a better error than not being able to find the `no` command :)

dormando:
this also leaves some files that aren't .gitignored or cleaned up properly:
```
	doc/protocol-binary-range.full
	doc/protocol-binary.full
```

unfortunately my buildbots don't test... the release process... I guess..

dormando:
Out in 1.5.20

PR #555, opened 2019-10-09
Support running tests in out-of-tree build

olajep:
The test targets were broken for out-of-tree builds.
`make test` and
`make test_basic_tls`


dormando:
Out in 1.5.20

PR #556, opened 2019-10-09
restart: add error handling when a tag is not found in a metadata file.

minkikim89:
It seems that there is a problem that a process is executed normally even if it fails when reading the tag from the metadata file.

### DIFF
--- a/restart.c
+++ b/restart.c
@@ -82,6 +82,7 @@ static int restart_check(const char *file) {
     restart_cb_ctx ctx;
 
     ctx.f = f;
+    ctx.cb = NULL;
     ctx.line = NULL;
     ctx.done = false;
     if (restart_get_kv(&ctx, NULL, NULL) != RESTART_DONE) {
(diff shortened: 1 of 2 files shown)


dormando:
Thanks! Feel up to writing a test?

minkikim89:
@dormando 
OK. I will do that.

minkikim89:
@dormando 
I wrote a test.
Please review it.

dormando:
Hey,

So I think this avoids reading an empty file, but it doesn't guarantee a tag was found.

What would cover more bases:
in `restart_check`, init ctx.cb to NULL
then change your ctx.done check to ensuring ctx.cb is not-NULL. That guarantees something was in the file and the first line was a tag rather than a key (or corrupted value that starts with 'K' I guess)

minkikim89:
@dormando 
I fixed it!

dormando:
Looks good, thanks!

dormando:
merged for next, thanks!

dormando:
Out in 1.5.20

PR #560, opened 2019-10-14
fix potential deadlock bug

neal-zhu:
Release mutex before return


dormando:
Nice, thanks!

dormando:
Edited the commit message a bit and merged for next release, thanks again!

dormando:
Out in 1.5.20

PR #561, opened 2019-10-21
restart: remove unused restart_mmap_set function

minkikim89:
restart_mmap_set function in restart.h is not used but has not been removed.


dormando:
thanks!

dormando:
Out in 1.5.20

PR #568, opened 2019-11-09
Use a proper data type for settings.sig_hup

### LINKED ISSUE #386: Fix signal handler

danielschemmel:
This PR does not attempt to fix #386, but rather only fix `SIG_HUP` by assigning `settings.sig_hup` a proper data type (see [here](https://en.cppreference.com/w/c/program/sig_atomic_t) for further references).

### DIFF
--- a/memcached.h
+++ b/memcached.h
@@ -19,6 +19,7 @@
 #include <unistd.h>
 #include <assert.h>
 #include <grp.h>
+#include <signal.h>
 
 #include "itoa_ljust.h"
 #include "protocol_binary.h"
@@ -398,7 +399,7 @@ struct settings {
     int item_size_max;        /* Maximum item size */
     int slab_chunk_size_max;  /* Upper end for chunks within slab pages. */
     int slab_page_size;     /* Slab's page units. */
-    bool sig_hup;           /* a HUP signal was received but not yet handled */
+    volatile sig_atomic_t sig_hup;  /* a HUP signal was received but not yet handled */
     bool sasl;              /* SASL on/off */
     bool maxconns_fast;     /* Whether or not to early close connections */
     bool lru_crawler;        /* Whether or not to enable the autocrawler thread */


dormando:
was there a bug (ie; total loss of sig_hup flip) or similar? I wanted to avoid an extra memory barrier in the clock handler since 1) it's only checked once per second and 2) it's fine if it's off by multiple seconds, even.

I guess it's wrong because the clock handler is also a writer, flipping it to false... but still just a technicality given what's actually going on here.

danielschemmel:
First, `sig_atomic_t` does not need to emit a memory barrier, as it only guarantees that the write cannot be interrupted by a signal and is actually performed to memory. Basically, it does not change anything versus what the code is currently expected to do anyway. Unlike the other "atomic" types, this type is not intended to be safe for use in multi-threading, and does not guarantee inter-thread ordering of memory accesses.

Second, the potential bug waiting to happen here is that the `clock_handler` may theoretically screw up the [`if` that deals with `settings.sig_hup`](https://github.com/memcached/memcached/blob/1939cf9edc0656f7b1cfab6365f313b56b07a333/memcached.c#L7520). For example, consider that `settings.sig_hup` is either `false`, or will be set to `false` afterwards. This would enable the following transformation, which then begins swallowing "some" signals:

```
if (settings.sig_hup) {
    settings.sig_hup = false;
    authfile_load(settings.auth_file);
}
```
->
```
if (settings.sig_hup) {
    authfile_load(settings.auth_file);
}
settings.sig_hup = false;
```

With an aggressive enough optimizer, you could then follow this further, e.g., through the whole loop into the following (after all, the ordering is not enforced in any way, and the variable is never written anywhere else):

```
settings.sig_hup = false;
if (settings.sig_hup) {
    authfile_load(settings.auth_file);
}
```
The next step is obvious: drop the whole `if`.

Now, of course there is no reason to be alarmist: This will **probably** not occur anytime soon. But AFAIK there is no real reason to not do it correctly in this specific instance.

dormando:
that's fair, thanks!

dormando:
(though your example makes sense; and I've not had enough coffee to remember things properly today, if I ever saw a compiler "optimize" something outside of a loop conditional for the same variable like that I'd be throwing pitchforks. I'd buy that a computer might reorder things accidentally though)

danielschemmel:
Well, C and C++ are great areas for program testing and verification research, because they sometimes allow things that make developers throw pitchforks ;).

dormando:
Out in 1.5.20

PR #569, opened 2019-11-09
Fix data race in `assoc_start_expand`

### LINKED ISSUE #567: Data Races in `assoc_start_expand`
The `clock_handler` [calls](https://github.com/memcached/memcached/blob/1939cf9edc0656f7b1cfab6365f313b56b07a333/memcached.c#L7517) [`assoc_start_expand`](https://github.com/memcached/memcached/blob/1939cf9edc0656f7b1cfab6365f313b56b07a333/assoc.c#L143) without acquiring any specific locks. In there, `started_expanding` is both read and potentially written without locks or atomic/synch builtins.

At the same time, [`assoc_maintenance_thread`](https://github.com/memcached/memcached/blob/1939cf9edc0656f7b1cfab6365f313b56b07a333/assoc.c#L199) also [writes `started_expanding`](https://github.com/memcached/memcached/blob/1939cf9edc0656f7b1cfab6365f313b56b07a333/assoc.c#L249). This thread holds the `maintenance_lock`, but that does not help, as the other access ignores it.

Even though, this is "just" a boolean flag, there are still two properties required for it to be useful:
1. It needs to be ensured that the access is actually dispatched to main memory (or, more abstractly speaking, becomes visible to other threads) at all.
2. This memory operation may not be reorder w.r.t. certain other operations, such as those on `hashpower` (which similarly may be written in `assoc_maintenance_thread` and read in `assoc_start_expand`).

Note that the C11 standard (previous standards ignored the existence of multiple threads completely) considers data races to be UB (see e.g., $5.1.2.4/20). Therefore, although current compilers may oblige by performing these operations in a manner that does not conflict with the semantic idea, future compiler versions may cause problems.

A possible solution would be to implement the variables that need to be quickly accessed in the `clock_handler` as C11 atomics, or by way of compiler builtins ([which modern GCC versions also base on these atomics](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#g_t_005f_005fatomic-Builtins)).

danielschemmel:
Fixes #567 by performing a singular `try_lock` on the `maintenance_lock`.

### DIFF
--- a/assoc.c
+++ b/assoc.c
@@ -48,7 +48,6 @@ static item** old_hashtable = 0;
 
 /* Flag: Are we in the middle of expanding now? */
 static bool expanding = false;
-static bool started_expanding = false;
 
 /*
  * During expansion we migrate values with bucket granularity; this is how
@@ -141,13 +140,11 @@ static void assoc_expand(void) {
 }
 
 void assoc_start_expand(uint64_t curr_items) {
-    if (started_expanding)
-        return;
-
-    if (curr_items > (hashsize(hashpower) * 3) / 2 &&
-          hashpower < HASHPOWER_MAX) {
-        started_expanding = true;
-        pthread_cond_signal(&maintenance_cond);
+    if (pthread_mutex_trylock(&maintenance_lock) == 0) {
+        if (curr_items > (hashsize(hashpower) * 3) / 2 && hashpower < HASHPOWER_MAX) {
+            pthread_cond_signal(&maintenance_cond);
+        }
+        pthread_mutex_unlock(&maintenance_lock);
     }
 }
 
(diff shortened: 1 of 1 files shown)


assoc.c
```diff
@@ -141,13 +140,11 @@ static void assoc_expand(void) {
...
 
 void assoc_start_expand(uint64_t curr_items) {
-    if (started_expanding)
-        return;
-
-    if (curr_items > (hashsize(hashpower) * 3) / 2 &&
-          hashpower < HASHPOWER_MAX) {
-        started_expanding = true;
-        pthread_cond_signal(&maintenance_cond);
+    if (0 == pthread_mutex_trylock(&maintenance_lock)) {
```
dormando: 0 == ? is inconsistent with rest of codebase

danielschemmel: I apologize, an updated version is pushed that flips the comparison to the right hand side.

dormando:
Out in 1.5.20

Issue #485, opened 2019-05-03
prove t/whitespace.t fail


hernaneche:
```
# make test 
or
# prove t/whitespace.t
```

**Ends in same fail results**

```
t/whitespace.t .. 1/319
#   Failed test 'Makefile (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'Makefile.in (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'aclocal.m4 (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'cache.o (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'compile (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'config.guess (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'config.h (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'config.h.in (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'config.log (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'config.status (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'config.sub (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'configure (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'depcomp (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'doc/Makefile (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'doc/Makefile.in (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'install-sh (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'memcached (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

#   Failed test 'memcached-assoc.o (see devtools/clean-whitespace.pl)'
#   at t/whitespace.t line 42.

[...]

PR #570, opened 2019-11-12
Ensure t/whitespace.t dies with a good error outside valid git checkouts

terabyte:
Greetings!  I'm trying to familiarize myself with the memcached codebase so I thought I'd try to bang out a few easy "help wanted" issues.  Please let me know if there is anything I can change to get this PR accepted.  Thanks!

The test as written uses git to find a list of files to check, so it doesn't
work correctly outside a git checkout.

Reported in #485

### DIFF
--- a/t/whitespace.t
+++ b/t/whitespace.t
@@ -5,6 +5,11 @@ our @files;
 
 BEGIN {
     chdir "$Bin/.." or die;
+    unless (-d "$Bin/../.git") {
+        use Test::More;
+        plan skip_all => "Skipping test because this does not appear to be a memcached git working directory";
+        exit 0;
+    }
 
     my @exempted = qw(Makefile.am ChangeLog doc/Makefile.am README README.md compile_commands.json);
     push(@exempted, glob("doc/*.xml"));


dormando:
Thanks!

Close! If you put the die there, the test will simply fail when run outside of a git repo (like a tarball download). If you look a few lines below at the skip_all + exit, that's what you have to repeat there.

Normally if git isn't even installed or you're not in a git repo at all it will properly skip the tests. The bug is when people unroll a memcached tarball inside something else that has a git repo, so git ls-files works and it dies processing the wrong data. So we just need the extra skip_all if the git directory isn't local.

terabyte:
Ah, that makes sense!  Is this more like what you had in mind?  Sorry if this is more work than you just doing it =)

terabyte:
Erm, minor typo, re-uploading in a sec after I've tested more carefully locally.

terabyte:
Tested locally with and without a .git directory:

    t/whitespace.t .............. skipped: Skipping test because this does not appear to be a memcached working directory

dormando:
message should probably say something about git :)

skips properly when the git dir is one below the base directory?

terabyte:
Yes, it checks that the .git directory specifically exists in "$Bin/../.git".  We could try to be even fancier by calling something like:

     $ git rev-parse --show-toplevel
    /home/cmyers/projects/memcached

And then doing some rel2abs directory math... Also maybe we could just run git commands to see if it "looks like" a memcached git checkout somehow?  Open to ideas on how that might work.  I wrote this change assuming that if the .git directory is in the right "level", it is probably correct.

dormando:
Doesn't need to be fancy; just confirming you tested the original condition :) I'll give it a local test and queue it up for the next branch when I swing back for a merge down.

dormando:
out in 1.5.21, thanks!

Issue #544, opened 2019-09-26
Provide a config option to disable "watch" commands


robertmathews:
As a feature enhancement, I'd like to suggest a configuration option to disable the "watch" commands.

In some cases, the "watch" commands can be a privacy/security issue, particularly on shared hosting. It's common for shared hosting companies to offer a single server-wide memcached instance on localhost as a convenience to sites on that server. Some software that stores data in memcache therefore uses private hashed/random keys to attempt to provide privacy-through-obscurity. However, the "watch" command reveals the keys to anyone who can connect, allowing one user on the server to bypass the "privacy" and view and alter another's data.

(A more general problem would be memcached instances unwisely exposed to the Internet, where anyone -- not just localhost users -- can connect and view keys.)

Two examples of software that uses memcache in a way that can be compromised if someone else views the key are the PHP/PECL memcached module that stores PHP sessions using unique random keys, and the "W3 Total Cache" WordPress plugin that stores cached pages using a randomly generated key based off a secret number. In the first case, the ability to view the keys allows PHP session hijacking; in the second case, it allows modification of pages displayed by other websites on the same server.

Yes, I know and agree that people shouldn't do this kind of thing on shared servers. But they do. Given that, an option to disable it seems useful. If a patch would be accepted for this, let me know.

dormando:
This is an excellent suggestion, and a feature we already have for other ... features... there's a "don't allow dumping" and "don't allow flush_all" flag. While I hope some day we can multi-tenant more properly, for now I understand that people do this and that's alright.

I'll leave this open until I take care of it. Also note that now there's an ASCII based authentication mechanism, which would make it easier for shared hosts to at least run individual memcached's for users without anyone just being able to connect to them.

dormando:
Fixed in 1.5.21

PR #571, opened 2019-11-12
configuration option to disable watch commands

terabyte:
Suggested in #544.  This is my first stab at this feature.  Looking for
feedback about what to do when commands are disabled (print ERROR,
other?).  Examples of other feature flags I should look at?

TODO:
* test locally
* write tests
* get feedback on help option name / descriptions
* other?

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -320,6 +320,7 @@ static void settings_init(void) {
     settings.logger_watcher_buf_size = LOGGER_WATCHER_BUF_SIZE;
     settings.logger_buf_size = LOGGER_BUF_SIZE;
     settings.drop_privileges = false;
+    settings.no_watch = false;
 #ifdef MEMCACHED_DEBUG
     settings.relaxed_privileges = false;
 #endif
@@ -5577,6 +5578,12 @@ static void process_watch_command(conn *c, token_t *tokens, const size_t ntokens
     assert(c != NULL);
 
     set_noreply_maybe(c, tokens, ntokens);
+    if (settings.no_watch) {
+        // watch is not allowed
+        out_string(c, "CLIENT_ERROR watch commands not allowed");
+        return;
+    }
+
     if (ntokens > 2) {
         for (x = COMMAND_TOKEN + 1; x < ntokens - 1; x++) {
             if ((strcmp(tokens[x].value, "rawcmds") == 0)) {
@@ -7634,6 +7641,7 @@ static void usage(void) {
            "   - modern:              enables options which will be default in future.\n"
            "             currently: nothing\n"
            "   - no_modern:           uses defaults of previous major version (1.4.x)\n"
+           "   - no_watch:            disable watch commands\n"
 #ifdef HAVE_DROP_PRIVILEGES
            "   - drop_privileges:     enable dropping extra syscall privileges\n"
            "   - no_drop_privileges:  disable drop_privileges in case it causes issues with\n"
@@ -8304,6 +8312,7 @@ int main (int argc, char **argv) {
         NO_INLINE_ASCII_RESP,
         MODERN,
         NO_MODERN,
+        NO_WATCH,
         NO_CHUNKED_ITEMS,
         NO_SLAB_REASSIGN,
         NO_SLAB_AUTOMOVE,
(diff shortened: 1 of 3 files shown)


dormando:
The flush_all command has a similar disable flag, so you can model it the same way.

terabyte:
So I have a pass above including tests even, but unfortunately the tests hang.  I implemented it looking at the way flush_all is disabled (so it uses auth failure in the case where a watch command is attempted).  The current tests as written hang for 600s then time out.  I think this is because the auth failure does not include a newline at the end to flush the buffer?  I'm kinda stuck on that right now:
```
$  telnet localhost 11211
Trying ::1...
Connected to localhost.
Escape character is '^]'.
watch
Auth failure.
```
Note there is no newline after the "Auth failure." line.  I think this is why my tests are hanging?  Here is what the output currently looks like:
```
$ prove -v t/watcher.t
t/watcher.t .. 
1..13
ok 1 - watcher enabled
ok 2 - basic get works
ok 3 - skipped some lines
ok 4 - saw a real log line after a skip
ok 5 - stored the long key
ok 6 - got the correct uri encoded key length
ok 7 - new watcher enabled
ok 8 - evictions watcher enabled
ok 9 - found rawcmd log entry
ok 10 - found eviction log entry
ok 11 - mutations watcher enabled
ok 12 - correctly logged cas command
Timeout.. killing the process
Timeout.. killing the process
not ok 13 - attempted watch gave auth failure with no_watch option
#   Failed test 'attempted watch gave auth failure with no_watch option'
#   at t/watcher.t line 133.
Auth failure.': ' 
#     expected: 'Auth failure.'
# Looks like you failed 1 test of 13.
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/13 subtests 

Test Summary Report
-------------------
t/watcher.t (Wstat: 256 Tests: 13 Failed: 1)
  Failed test:  13
  Non-zero exit status: 1
Files=1, Tests=13, 603 wallclock secs ( 0.02 usr  0.01 sys +  0.16 cusr  0.44 csys =  0.63 CPU)
Result: FAIL
```

terabyte:
aaaaaaand I figured it out.  There are two protocols, the binary one and the text one.  I copied the implementation for the binary protocol that disables flush_all and used that in the text protocol code for watch.  This has been rectified now.

I didn't see any watch code for the binary protocol, does it not support watch commands?  if the answer to that is "yes", then I think this change is good to go!

memcached.c
```diff
@@ -5577,6 +5578,12 @@ static void process_watch_command(conn *c, token_t *tokens, const size_t ntokens
     assert(c != NULL);
 
     set_noreply_maybe(c, tokens, ntokens);
+    if (settings.no_watch) {
+        // watch is not allowed
```
dormando: redundant comment.

memcached.h
```diff
@@ -428,6 +428,7 @@ struct settings {
     unsigned int logger_watcher_buf_size; /* size of logger's per-watcher buffer */
     unsigned int logger_buf_size; /* size of per-thread logger buffer */
     bool drop_privileges;   /* Whether or not to drop unnecessary process privileges */
+    bool no_watch; /* allows watch commands to be dropped */
```
dormando: why "no_watch" vs "watch_enabled", like flush_enabled?

dormando:
I never bothered implementing watch for binprot. I'm deprecating the entire protocol in favor of the new meta commands anyway.

couple nits, looks fine otherwise. thanks!

memcached.h
terabyte: I think I saw a "no_foo" flag elsewhere.  Happy to reverse this if you prefer.  I also noticed that the flush feature was a regular option, not an extended option, but I had already started to write this as an extended option - would you like me to change that or is an extended option fine?

memcached.c
terabyte: will remove.

memcached.h
dormando: hmm I guess the options are kind of inconsistent. Lets make it a top level option if there's an obvious free character? Most of the extendeds are tuning or very rarely used.

dormando:
@terabyte I disappeared for longer than expected. still interested in finishing this PR?

dormando:
I flipped this around to the -W option, didn't change anything else. Merged into `next` now

dormando:
out in 1.5.21, thanks!

PR #572, opened 2019-11-18
Let's reduce the global warming bill by reducing

devnexen:
this lengthy help string also correcting
TLS feature using native API for comparison instead
as on some platforms pthread_t is an opaque type.

### DIFF
--- a/tls.c
+++ b/tls.c
@@ -53,7 +53,7 @@ ssize_t ssl_sendmsg(conn *c, struct msghdr *msg, int flags) {
     // conn_worker_readd.
     // Currntly this connection would not be served by a different thread
     // than the one it's assigned.
-    assert(c->thread->thread_id == (unsigned long)pthread_self());
+    assert(pthread_equal(c->thread->thread_id, pthread_self()) != 0);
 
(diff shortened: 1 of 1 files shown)


dormando:
first change might collide with #496 - drop it and make a comment there or add later? Also please use something more descriptive than reducing global warming for a reason :)

The pthread_t thing is a good catch. I've screwed that up a few times now.

devnexen:
I admit I was in "comical" mood yesterday :-) so yes seems the other PR mght fix it rightfully, we ll see.

dormando:
I like (and often have) jokey comments :) they just come after or in the middle of the justification note.

moisesguimaraes:
+1

This fixed the issue I was having trying to compile memcached with TLS in alpine linux:

```
In file included from memcached.h:20,
                 from tls.c:1:
tls.c: In function 'ssl_sendmsg':
tls.c:56:33: error: comparison between pointer and integer [-Werror]
     assert(c->thread->thread_id == (unsigned long)pthread_self());
                                 ^~
```

dormando:
out in 1.5.21, thanks!

PR #573, opened 2019-11-18
Fix data race on stats due to an access w/o acquiring the lock

danielschemmel:
In `drive_machine`, the `stats_state` [is accessed](https://github.com/memcached/memcached/blob/e7c7a1089c066d105bf0a6c1a812668644a4a6f0/memcached.c#L6723) without holding the `STATS_LOCK`, which causes data races against other threads that mutate the stats even while they are holding the lock.

This PR proposes to keep the time that this lock is held to a minimum by acquiring the lock before the `if` and releasing it after potentially updating stats. It would also be possible to acquire the lock twice: once to compute the condition, and once to increment the count of rejected connections.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -6719,14 +6719,20 @@ static void drive_machine(conn *c) {
                 }
             }
 
-            if (settings.maxconns_fast &&
-                stats_state.curr_conns + stats_state.reserved_fds >= settings.maxconns - 1) {
+            bool reject;
+            if (settings.maxconns_fast) {
+                STATS_LOCK();
+                reject = stats_state.curr_conns + stats_state.reserved_fds >= settings.maxconns - 1;
+                stats.rejected_conns++;
+                STATS_UNLOCK();
+            } else {
+                reject = false;
+            }
+
+            if (reject) {
                 str = "ERROR Too many open connections\r\n";
                 res = write(sfd, str, strlen(str));
                 close(sfd);
-                STATS_LOCK();
-                stats.rejected_conns++;
-                STATS_UNLOCK();
             } else {
                 void *ssl_v = NULL;
 #ifdef TLS


danielschemmel:
Note: This PR tackles a different issue than #575.

dormando:
On 64-bit x86 systems the dirty read there should be fine, but that is wrong for 32bit.

Throwing that global lock around every connect is a non-starter. At very least the maxconns_fast check should happen unlocked, then either atomics or a separate lock for curr_conns should be used.

danielschemmel:
I have changed the PR to only lock if `maxconns_fast` is false.

If you give me a few pointers how you would like the additional lock for `stats_state.curr_conns` and `stats_state.reserved_fds` to look like, I am willing to try that variant that as well. To attack this problem with atomics, we would need to combine those two variables, as I am not aware of a way to perform the addition with both variables being loaded in a single atomic step.

memcached.c
```diff
@@ -6719,14 +6719,20 @@ static void drive_machine(conn *c) {
                 }
             }
 
-            if (settings.maxconns_fast &&
-                stats_state.curr_conns + stats_state.reserved_fds >= settings.maxconns - 1) {
-                str = "ERROR Too many open connections\r\n";
-                res = write(sfd, str, strlen(str));
-                close(sfd);
+            bool reject;
+            if(settings.maxconns_fast) {
```
dormando: spacing

dormando:
reserved_fds is static before any listener sockets are opened, which should make that easier. I'm not sure off the top of my head if there's a better way to fix it. The way you've patched it is probably fine since the STATS_LOCK() was there before, and removing the branch from within the lock should help keep it short.

memcached.c
danielschemmel: There is now a space between the `if` and the condition.

dormando:
out in 1.5.21, thanks!

dormando:
For fuck's sake, how did I miss this..

This logic completely breaks the original feature; now every connection is counted as a rejection. The maxconns test doesn't cover rejections being counted properly and I was too distracted to check the tests here.

PR #574, opened 2019-11-18
Some resources / leak fixes proposals

devnexen:


dormando:
Seems sane, thanks! Are you eyeballing these or using anything mechanical to find/test for?

dormando:
out in 1.5.21, thanks!

a lot of people just run their pet scanning systems against us and report bugs randomly... I kinda care more about scanning for upcoming work than nits in the older code, but I'll take either passively :)

PR #575, opened 2019-11-18
Fix a second data race on stats

danielschemmel:
This data race (different from #573) is due to the logging thread being started before the stats are initialized. As `stats_init` only initializes memory, I simply moved the `stats_init` call up one line. The logging thread correctly locks the stats structure when updating it in [`logger_thread_sum_stats`](https://github.com/memcached/memcached/blob/e7c7a1089c066d105bf0a6c1a812668644a4a6f0/logger.c#L553), but that does not help, as the initialization of the stats data structures has not happened yet (and is performed without holding the stats lock).

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -9528,8 +9528,8 @@ int main (int argc, char **argv) {
     }
 
     /* initialize other stuff */
-    logger_init();
     stats_init();
+    logger_init();
     conn_init();
     bool reuse_mem = false;
     void *mem_base = NULL;


dormando:
what version of the source are you testing against? logger_init doesn't touch stats anymore.

dormando:
oh I misread. logger_thread_sum_stats could get called before stats_init... except without any worker threads it's going to add zero's. If !found_logs test also ran continue it'd similarly be a non-issue. In reality it's going to usleep for a millisecond before it does anything.

this is super nitpicky but technically correct. while some of the bugs you're finding have been pretty good, ones like these borderline on annoying because I have to spend time to validate/understand them and they don't fix anything other than the errors in a program I'll never have access to. It might be better to batch these up so they're less like interruptions.

danielschemmel:
Yes, this is indeed quite nitpicky, sorry about that. If we find more similar bugs, we will try to bunch them up.

In fact, [the `memset`](https://github.com/memcached/memcached/blob/e7c7a1089c066d105bf0a6c1a812668644a4a6f0/memcached.c#L236) that the [`logger_thread_sum_stats`](https://github.com/memcached/memcached/blob/e7c7a1089c066d105bf0a6c1a812668644a4a6f0/logger.c#L504) function races against is not necessary in the first place, as [`stats` is a global variable](https://github.com/memcached/memcached/blob/e7c7a1089c066d105bf0a6c1a812668644a4a6f0/memcached.c#L142) with static storage duration, it is already initialized to all zero (c98 §6.7.8/10).

However, to my eyes there is a semantic ordering between the stats and the logger: The logger uses the stats, but not vice versa.

Note also, that the [`usleep` in `logger_thread`](https://github.com/memcached/memcached/blob/e7c7a1089c066d105bf0a6c1a812668644a4a6f0/logger.c#L530) does never trigger during the first iteration of the loop, as [`to_sleep` is initialized to `MIN_LOGGER_SLEEP`](https://github.com/memcached/memcached/blob/e7c7a1089c066d105bf0a6c1a812668644a4a6f0/logger.c#L518) and the thread is only sent to sleep if it is larger than that.

P.S.: We are working on master, and also try to validate potential bugs against master when reporting them to ensure that no commits happened in the meantime that change the relevant behavior.

dormando:
Twice misread. Too groggy to read.

For what it's worth, master updates only when releases happen. Averages once or twice a month. Of course, I care the most about bugs in stuff being actively developed... but instead we'll just get those long threads where I talk to myself for a week and then years later someone's static analyzer will nitpick pseudo bugs that weren't fixed in the meantime :)

danielschemmel:
If you point us to another branch (next?) we can analyze that instead.

Sorry for all the bother though, but I do hope that this is a net benefit for this project!

dormando:
out in 1.5.21, thanks!

PR #578, opened 2019-11-22
Fix the bug sasl will load config from the wrong path when HAVE_SASL_CB_GETCONFPATH enabled

zhenggu:
From the function load_config of https://github.com/cyrusimap/cyrus-sasl/blob/master/lib/server.c, when we return the path "/etc/sasl2/memcached.conf" in function sasl_getconf of memcached, it will concat the path to "/etc/sasl2/memcached.conf/memcached.conf", by the following code:
```
        snprintf(config_filename, len, "%.*s%c%s.conf", (int)path_len, path_to_config,
	        HIER_DELIMITER, global_callbacks.appname);
```
so the correct configure file /etc/sasl2/memcached.conf will not be read.

### DIFF
--- a/sasl_defs.c
+++ b/sasl_defs.c
@@ -12,8 +12,8 @@ char my_sasl_hostname[1025];
  * specify one in the environment variable SASL_CONF_PATH
  */
 const char * const locations[] = {
-    "/etc/sasl",
-    "/etc/sasl2",
+    "/etc/sasl/memcached.conf",
+    "/etc/sasl2/memcached.conf",
     NULL
 };
 #endif
@@ -82,18 +82,51 @@ static int sasl_server_userdb_checkpass(sasl_conn_t *conn,
 }
 #endif
 
-#if defined(HAVE_SASL_CB_GETCONF) || defined(HAVE_SASL_CB_GETCONFPATH)
+#if defined(HAVE_SASL_CB_GETCONF)
 static int sasl_getconf(void *context, const char **path)
 {
     *path = getenv("SASL_CONF_PATH");
 
     if (*path == NULL) {
         for (int i = 0; locations[i] != NULL; ++i) {
-            char buff[30];
-            strcpy(buff, locations[i]);
-            strcat(buff, "/memcached.conf");
-            if (access(buff, F_OK) == 0) {
+            if (access(locations[i], F_OK) == 0) {
                 *path = locations[i];
+                break;
+            }
+        }
+    }
+
+    if (settings.verbose) {
+        if (*path != NULL) {
+            fprintf(stderr, "Reading configuration from: <%s>\n", *path);
+        } else {
+            fprintf(stderr, "Failed to locate a config path\n");
+        }
+
+    }
+
+    return (*path != NULL) ? SASL_OK : SASL_FAIL;
+}
+#elif defined(HAVE_SASL_CB_GETCONFPATH)
+static int sasl_getconf(void *context, const char **path)
+{
+    *path = getenv("SASL_CONF_PATH");
+
+    char buf[50];
+    if (*path == NULL) {
+        for (int i = 0; locations[i] != NULL; ++i) {
+            int locationlen = strlen(locations[i]);
+            strncpy(buf, locations[i], sizeof(buf) - 1);
+            strncat(buf, "/memcached.conf", sizeof(buf) - 1 - locationlen);
+            if (access(buf, F_OK) == 0) {
+                *path = locations[i];
+                break;
+            }
+            else if (access(locations[i], F_OK) == 0) {
+                char *p = strrchr(locations[i], '/');
+                memset(buf, 0, sizeof(buf));
+                memcpy(buf, locations[i], p - (char *)locations[i]);
+                *path = buf;
                 break;
             }
         }
--- a/sasl_defs.c
+++ b/sasl_defs.c
@@ -26,7 +26,7 @@ const char *const locations_dir_path[] = {
 };
 
 /* If the element of locations is directory, locations_file_path stores
- * the actual configue file which used by sasl, when GETCONFPATH is
+ * the actual configure file which used by sasl, when GETCONFPATH is
  * enabled */
 const char *const locations_file_path[] = {
     "/etc/sasl/memcached.conf/memcached.conf",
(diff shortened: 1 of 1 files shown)


dormando:
Thanks!

I was under the impression this was deliberate, though. You're supposed to give a directory path and then have the appname set correctly for it to find memcached.conf? I'll have to look at this more closely when I get a chance.

zhenggu:
thanks @dormando for the information, but it is really strange that we have a configure file /etc/sasl2/memcached.conf, after https://github.com/memcached/memcached/commit/7f2c52d38c35d7b7a9a408bc706fddbaa3540f36#diff-cd426aac0d29d07a8472266482403cb5L14 applied, our sasl password will be verified failed for the pwcheck changed from saslauthd to auxprop, when we debug sasl, there is no config loaded.

dormando:
Long delay :(

Started looking into this a bit. Looks like during that patch you referenced I didn't research what "getconfpath" meant. We end up defining the callback for getconf and getconfpath the same, but they expect different inputs :(

I've seen getconfpath referred to as a vendor patch in a few places. I (or anyone) need to do some further research to see if GETCONF is actually used anywhere, and what its input is. Else this PR would be fixing the GETCONFPATH cases but potentially break any GETCONF cases.

I hate SASL :( Sorry.

zhenggu:
Thanks @dormando for your response. So can I change sasl_getconf into to functions, maybe #ifdef HAVE_SASL_CB_GETCONF one function, #ifdef HAVE_SASL_CB_GETCONFPATH, another function like this?

dormando:
I guess so yes. I would first do some research to see if you can find if GETCONF exists anywhere so you can test it.

dormando:
We also have to keep in mind that any existing users may have adjusted their paths to be the long form :( so we might have to check ie: `memcached.conf/memcached.conf` along with the "correct" path before giving up.

zhenggu:
Thanks very much for your help

dormando:
Thanks for your patience and willingness to work on it :)

zhenggu:
HAVE_SASL_CB_GETCONF is added on commit [7a221d95a870bd01ba1c93650c63abd583b00d39](https://github.com/memcached/memcached/commit/7a221d95a870bd01ba1c93650c63abd583b00d39), seems from centos or some redhat related version it is SASL_CB_GETCONF, I will check other platform to see whether it is the same

zhenggu:
I have checked macosx/ubuntu/centos, SASL_CB_GETCONFPATH is there, but no SASL_CB_GETCONF there, not sure about windows.

zhenggu:
have checked the source code of https://github.com/cyrusimap/cyrus-sasl, only from branch v2.1 SASL_CB_GETCONFPATH is added, but no SASL_CB_GETCONF there, is there any compatiable library which use SASL_CB_GETCONF?
Anyway, I will working on the code to splite SASL_CB_GETCONFPATH and SASL_CB_GETCONF first.

zhenggu:
have added the source code

sasl_defs.c
```diff
@@ -95,6 +96,27 @@ static int sasl_getconf(void *context, const char **path)
...
+    char buf[50];
+    if (*path == NULL) {
+        for (int i = 0; locations[i] != NULL; ++i) {
+            int locationlen = strlen(locations[i]);
+            strncpy(buf, locations[i], sizeof(buf) - 1);
+            strncat(buf, "/memcached.conf", sizeof(buf) - 1 - locationlen);
+            if (access(buf, F_OK) == 0) {
+                *path = locations[i];
+                break;
+            }
```
dormando: code style is `} else `

sasl_defs.c
```diff
@@ -95,6 +96,27 @@ static int sasl_getconf(void *context, const char **path)
             }
         }
     }
+#elif defined(HAVE_SASL_CB_GETCONFPATH)
+    char buf[50];
```
dormando: trying to avoid adding too many magic buffers like this. perhaps just make a "locations_path[]" with it pre-cut above? then we don't need the strncpy/strncat/etc stuff either.

sasl_defs.c
```diff
@@ -87,6 +87,7 @@ static int sasl_getconf(void *context, const char **path)
 {
     *path = getenv("SASL_CONF_PATH");
 
+#if defined(HAVE_SASL_CB_GETCONF)
     if (*path == NULL) {
```
dormando: SASL_CONF_PATH environment variable should always override for both GETCONF and GETCONFPATH. This disables the environment variable if GETCONF is set.

zhenggu: I have picked up if (*path == NULL) out of the macro check to avoid this

sasl_defs.c
```diff
@@ -16,6 +16,23 @@ const char * const locations[] = {
...
+ * directory path of these elements */
+const char *const locations_dir_path[] = {
+    "/etc/sasl",
+    "/etc/sasl2",
+    NULL
+};
+
+/* If the element of locations is directory, locations_file_path stores
+ * the actual configue file which used by sasl, when GETCONFPATH is
+ * enabld */
```
dormando: enabled

zhenggu:
@dormando have fixed the typo

sasl_defs.c
```diff
@@ -16,6 +16,23 @@ const char * const locations[] = {
...
+/* If the element of locations is file, locations_dir_path stores the
+ * directory path of these elements */
+const char *const locations_dir_path[] = {
+    "/etc/sasl",
+    "/etc/sasl2",
+    NULL
+};
+
+/* If the element of locations is directory, locations_file_path stores
+ * the actual configue file which used by sasl, when GETCONFPATH is
```
dormando: lol doh. configue -> configure.

dormando:
ok I think I can merge this. Would you like to squash the commits? if not I can do that for you

zhenggu:
Please help me to do this, thanks @dormando

dormando:
Great! Thanks for sticking around on the long cycle for this one. I've squashed, edited a bit, and pushed for the next release.

zhenggu:
thanks @dormando for your great help

sasl_defs.c
zhenggu: Please help me to fix this

zhenggu: Please help me to fix it :)

dormando:
out in 1.6.0!

zhenggu:
thanks @dormando

PR #585, opened 2019-12-06
auth file, using alternative bcmp implementation instead to

devnexen:
check the token. less optimised than the usual memcmp especially
it goes through the whole buffers but more resilient against possible
attacks.

While at it, constifying a var which should have been.

### DIFF
--- a/authfile.c
+++ b/authfile.c
@@ -1,12 +1,15 @@
 /* -*- Mode: C; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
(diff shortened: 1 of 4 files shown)


dormando:
thanks!

avoiding side channels but not timing?

devnexen:
well, this should be time constant flavour.

dormando:
Yeah; it checks each byte it's still checking the plen instead of the bufferlen right? think you're supposed to zerofill and use constant size so you can't guess the length of the password as well

devnexen:
true. here it is more bcmp than anything else as well. just reasonable approach for a start :-)

dormando:
out in 1.5.21, thanks!

PR #589, opened 2019-12-11
Replace "return ;" with "return;"

iqr4m:
These lines are exception. Everywhere else return is followed by an immediate semicolon. This change will help future contributors to keep the style consistent.


dormando:
thanks!

dormando:
out in 1.5.21, thanks!

PR #591, opened 2019-12-13
stats: Fix unit tests

kanak:
* Update build instructions.
* Fix compilation issues in tests caused by not passing length of key.
* Fix tests broken by the change to ignore keys that do not have a
prefix ( commit b8ff91831829221d57944eafca28cf6420485061 ).


kanak:
@dormando :  I think the following might be good improvements too, please let me know if I should work on them (as part of this same pull request):

* [ ] Move the tests into testapp.c so that there isn't a different way to run these tests.
* [ ] Move documentation comments in stats.c into stats.h 
* [ ] Rename stats.c to stats_prefix.c (or something similar) so that it is not confused with memcached's stats subsystem.

dormando:
Hey,

Thanks! I've been thinking about deprecating the stats prefix code to be removed later, but it might not be bad to fix it up a bit in the meantime? I believe the logging code should be used so a system like this could be external to the daemon. It's a bit of a foot-gun as-is and has bitrotted for quite a while.

Do you make use of it at all?

kanak:
I have the concept of "prefix" in my key names because I have McRouter in front of my memcached servers, and I had just started investigating this feature to better understand my cache hit rates & perf. Unfortunately, my keys look like `<app>:<obj>:<id>`, so the code would give me `<app>` level stats since it stops at the first occurrence of the prefix, not `<app>:<obj>` which would be more useful.  I can fix this on the client side, but to your point, this might be an argument in favor of just a more general purpose logging mechanism that logged entire key.

Until this code is deprecated and removed, I'd love to clean things up so that the memcached code is easier to understand. Does my list of action items above sound good to you? I'm happy to add and subtract tasks as you see fit.

dormando:
Totally. I think no open patches are touching stats.c so it should be fine to do some work on it, though the code doesn't currently document in .h files. or at least, minimally if so, so you can skip it.

You should also try the logger feature (`watch`) and see how well that lines up with what you want. if you want to help improve that I can come up with some stuff to work on. I'd say it's about halfway done, but the hard parts should be done already.

kanak:
I've implemented the proposed changes, please let me know if you'd like me to change anything. Thank you!

dormando:
Thanks! I should be doing a merge-down review soon. Been stuck in bughunt hell for a while.

dormando:
out in 1.5.21, thanks!

dormando:
If you feel like working on anything else, hit me up ;)

PR #593, opened 2020-01-08
Build and test on amd64 and arm64

martin-g:
Hello,

With this Pull Request I'd like to request to add arm64 architecture to the build matrix at TravisCI.

ARM64 is used more and more for servers too and it would be good if Memcached is tested regularly.
I've ran the build and tests both on one of our arm64 machines and at [TravisCI](https://travis-ci.org/martin-g/memcached) and at the moment both are fine!

`amd64` is the default arch at TravisCI so it is preserved!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 sudo: required
 dist: xenial
 language: cpp
+arch:
+    - amd64
+    - arm64
 
 compiler:
     - gcc


dormando:
Thanks! I'll give this a quick think.

We already do build on arm64 (via an rpi 3). I actually have 5+ buildbots that do various platforms and architectures, but it's not exposed to the community right now. I'll _probably_ merge this, but need to revisit the buildbot setup soon to see if it can be integrated better.

martin-g:
It is good to hear that you already test on different architectures!

I've noticed BuildBot at https://github.com/memcached/memcached/wiki/DevelopmentRepos#development-work but I didn't see any builders at http://build.memcached.org:8010/#/builders

Thanks for your work on Memcached!

dormando:
Yeah it looks empty but there are a bunch there. After an upgrade a while back buildbot defaulted to hiding everything behind an account. I have to upgrade it again and see what my options are for exposing them publicly.

I also don't have a good way of letting random PR's run against the buildbots. I might just make it easy to do a quick review and kick them off... bit backlogged right now so not sure when I'll get to that. I run the buildbots before every release since most issues with that level of portability are minor/rare.

martin-g:
Thank you, @dormando !

PR #602, opened 2020-01-25
htonll macro checking fix proposal

devnexen:


dormando:
ended up fixing this slightly differently myself. thanks for the PR though!

devnexen:
have still the same issue with master

dormando:
the fix is in the next branch still. the last release was a crashfix so didn't include everything in next.

PR #603, opened 2020-01-25
Adding build information to README.md

pythonmobile:
Just adding the build information to README.md - Should we remove BUILD file?


dormando:
please update this to be against the 'next' branch instead of 'master' and squash some of your commits.

dormando:
just double checked: your README instructions will be wrong for the tarball distributions.

BUILD file isn't included in the tarballs, just the git repository. README.md is included in the tarballs. ./autogen.sh step is only needed for builds from git.

Merging the files doesn't make sense then. including build instructions in README.md is fine, so long as they're actually correct.

PR #609, opened 2020-02-01
stats_prefix: Remove test failure due to non-determinism

kanak:
If you have multiple keys, the order in which they are dumped is
non-deterministic because it is iterating over a hash table used by
prefix stats.

Fix the assertions by searching for expected lines in output instead
of checking that the strings are exactly the same.


dormando:
Shoot. This doesn't apply to `next`, can you fix?

kanak:
Rebased against next and fixed the conflict. Also changed the target of this pull request to "next" instead of "master" so that it doesn't show that there are 1500 lines changed :)

dormando:
all merged up, thanks!

dormando:
out in 1.6.0!

PR #610, opened 2020-02-04
NAPI_ID based thread selection as optional feature

kapatil:
This patch provides another thread selection option in the
"dispatch_conn_new" code path. The default mechanism for thread selection
is round-robin. This patch uses NAPI_ID based thread selection that
creates an association between socket(s) traffic that arrives on a given
NIC (hardware) queue and the memcached thread that consumes this traffic.
This mapping between a memcached thread and a HW NIC queue streamlines
the flow of data from the NIC to the application. In addition, an optimal
path with reduced context switches is possible, if epoll based busy
polling (sysctl -w net.core.busy_poll = <non-zero value>) is also enabled

This feature is enabled by specifying the command line parameter
"--napi_ids=<num>", where <num> is the number of available/assigned
hardware queues.  The patch also checks that the value of the <num>
parameter matches the <num> parameter specified for the "--threads=<num>"
option to ensure this mapping. If the option is unspecified, or the
conditions unmet, the code defaults to round robin thread selection.

v2: rebase on tip of tree (1.5.22).

Signed-off-by: Kiran Patil <kiran.patil@intel.com>

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -7724,6 +7724,16 @@ static void usage(void) {
     verify_default("ssl_keyformat", settings.ssl_keyformat == SSL_FILETYPE_PEM);
     verify_default("ssl_verify_mode", settings.ssl_verify_mode == SSL_VERIFY_NONE);
 #endif
+    printf("-N, --napi-ids=<num>      number of napi_ids for NAPI-ID based thread selection\n"
+           "                          Why to use this feature:                             \n"
+           "                          The Linux Kernel natively supports busy_polling feature\n"
+           "                          which enables polling of NIC HW queue directly from application\n"
+           "                          thread context. It is also extended for epoll:epoll_wait syscall.\n"
+           "                          Using busy_polling feature has direct and visible impact on\n"
+           "                          application performance w.r.t request to response time. Busy_polling\n"
+           "                          feature provides lower latency and improved predicatablity by\n"
+           "                          providing consistent result (consistent and lower response time\n"
+           "                          even at 99.9 percent\n");
     return;
 }
 
@@ -8539,6 +8549,7 @@ int main (int argc, char **argv) {
           "Y:"   /* Enable token auth */
           "e:"  /* mmap path for external item memory */
           "o:"  /* Extended generic options */
+          "N:"  /* NAPI_ID based thread selection */
           ;
 
     /* process arguments */
@@ -8579,6 +8590,7 @@ int main (int argc, char **argv) {
         {"auth-file", required_argument, 0, 'Y'},
         {"memory-file", required_argument, 0, 'e'},
         {"extended", required_argument, 0, 'o'},
+        {"napi-ids", required_argument, 0, 'N'},
         {0, 0, 0, 0}
     };
     int optindex;
@@ -9294,10 +9306,26 @@ int main (int argc, char **argv) {
             }
             free(subopts_orig);
             break;
+    case 'N':
+        settings.num_napi_ids = atoi(optarg);
+        if (settings.num_napi_ids <= 0) {
+	    fprintf(stderr, "Number of NAPI_IDs must be greater than 0\n");
+	    return 1;
+        }
+        break;
         default:
             fprintf(stderr, "Illegal argument \"%c\"\n", c);
             return 1;
         }
+    }
+
+    if (settings.num_napi_ids && settings.num_threads) {
+        if (settings.num_napi_ids > settings.num_threads) {
+            fprintf(stderr, "Number of NAPI_IDs(%d) cannot be more than num_threads (%d)\n",
+                    settings.num_napi_ids, settings.num_threads);
+            exit(EX_USAGE);
+        }
+        settings.napi_id_based_thread_selection = true;
     }
 
     if (settings.item_size_max < ITEM_SIZE_MAX_LOWER_LIMIT) {
--- a/thread.c
+++ b/thread.c
@@ -545,8 +545,100 @@ static void thread_libevent_process(int fd, short which, void *arg) {
     }
 }
 
-/* Which thread we assigned a connection to most recently. */
+/* The thread we assigned a connection to most recently. */
 static int last_thread = -1;
+
+/* This function simply returns next_thread based on round-robin */
+static LIBEVENT_THREAD *select_thread(void)
+{
+    int tid = (last_thread + 1) % settings.num_threads;
+    last_thread = tid;
+    return threads + tid;
+}
+
+/* This function selects a thread to process requests from a given connection.
+ * Thread selection is either based on the (existing) default mechanism,
+ * round-robin based thread selection, or based on NAPI_ID. NAPI_ID based
+ * selection creates an association between a (hardware) queue on a NIC and an
+ * application thread. This mapping streamlines the flow of data from the NIC
+ * to an application thread by having an application thread source data from
+ * the same hardware queue which serves up traffic for many connections
+ */
+static LIBEVENT_THREAD *select_thread_using_napi_id(int sfd)
+{
+    LIBEVENT_THREAD *thread = NULL;
+    int start_tid, i;
+    socklen_t len;
+    int napi_id;
+
+    /* first check if there is pre-existing NAPI_ID => Thread
+     * association, if so - use that thread, otherwise select unassigned
+     * or unassociated thread 
+     */
+#ifndef SO_INCOMING_NAPI_ID
+#define SO_INCOMING_NAPI_ID 56
+#endif /* SO_INCOMING_NAPI_ID */
+    len = sizeof(napi_id);
+    if (getsockopt(sfd, SOL_SOCKET, SO_INCOMING_NAPI_ID,
+		   &napi_id, &len) == -1) {
+        perror("getsockopt SO_INCOMING_NAPI_ID");
+        /* Disable napi_id based thread selection since socket option returned
+         * error, may be running on older linux kernel
+         */
+        settings.napi_id_based_thread_selection = false;
+        return select_thread();
+    }
+    /* getsockopt didn't fail but napi_id is still unassigned, This is expected if
+     * connection are setup on localhost (means not going thru' underlying
+     * device, hence for those local connection, napi_id is not expected
+     * Non NAPI-IDs are aggregated down to zero by getsockopt call
+     * for socket option SO_INCOMING_NAPI_ID
+     */
+	if (!napi_id) {
+        return select_thread();
+	}
+
+    /* find thread which has pre-existing association with napi_id,
+     * if there is pre-existing association of thread => napi_id,
+     * use that thread
+     */
+    for (int i = 0; i < settings.num_threads; i++) {
+        thread = threads + i;
+        if (thread->napi_id == napi_id) {
+            /* found associated thread based on NAPI_ID */
+            return thread;
+        }
+    }
+    /* if thread with pre-existing association of thread => napi_id is not
+     * found, select next available thread based that is not associated
+     * with any NAPI-ID 
+     */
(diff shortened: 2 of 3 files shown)


dormando:
Thanks! I probably won't get a chance to review for a few days; on a skim I think there are some indentation issues and some style stuff.

kapatil:
> Thanks! I probably won't get a chance to review for a few days; on a skim I think there are some indentation issues and some style stuff.

I guess, will wait for your comments after you get a chance to review the patch in details and then will resubmit the patch. Meanwhile, I will fix obvious indentation issue. Can you please point to the style issue as an example so that I know - what style to follow.

kapatil:
Found some regression with this patch and in process of fixing it. Will update new rev in next day or so.

kapatil:
Found small regression with the patch. Will upload new revision of this patch in next few days

kapatil:
Submitted new revision of the same patch which is v3 and addresses some regression found during our stress testing. Please let me know if you want me to change anything in patch, happy to address code review comments.

dormando:
What was the regression?

kapatil:
Regression was , after all the attempt to select thread based on NAPI_ID fails (which can happen and perfectly valid scenario) fails and “selected_thread” remains NULL.

Hence latest patch “v3: addresses that condition – to make sure thread is never NULL” If above described condition happens, V3 version disable feature “NAPI_ID bases thread selection logic” and fallbacks to default thread selection which is round robin.

Thanks,

  *   Kiran P.

kapatil:
Can I please get some review on latest revision of this patch? Happy to address comments. If patch looks good, can you please merge it.

dormando:
That's the plan :) I'll get to it when I get to it.

kapatil:
Thanks  a lot for responding and will stay tuned.

memcached.c
```diff
@@ -9294,12 +9297,28 @@ int main (int argc, char **argv) {
...
+        break;
         default:
             fprintf(stderr, "Illegal argument \"%c\"\n", c);
             return 1;
         }
     }
 
+    if (settings.num_napi_ids && settings.num_threads) {
+        if (settings.num_napi_ids > settings.num_threads) {
+            fprintf(stderr, "Number of NAPI_IDs(%d) cannot be more than num_threads (%d)\n",
```
dormando: The actual option is called --threads (-t). should also say something like "number supplied to --napi-ids"

thread.c
```diff
@@ -545,9 +545,125 @@ static void thread_libevent_process(int fd, short which, void *arg) {
...
+
+/* This function selects a thread to process requests from a given connection.
+ * Thread selection is either based on the (existing) default mechanism,
+ * round-robin based thread selection, or based on NAPI_ID. NAPI_ID based
+ * selection creates an association between a (hardware) queue on a NIC and an
+ * application thread. This mapping streamlines the flow of data from the NIC
+ * to an application thread by having an application thread source data from
+ * the same hardware queue which serves up traffic for many connections
+ */
+static void select_thread(int sfd, LIBEVENT_THREAD **thread_selected)
```
dormando: Why does `select_next_thread` return a LIBEVENT_THREAD, but `select_thread` is void with an input pointer?

thread.c
```diff
@@ -545,9 +545,125 @@ static void thread_libevent_process(int fd, short which, void *arg) {
...
+    LIBEVENT_THREAD *thread, *selected_thread = NULL;
+    socklen_t len;
+    int napi_id;
+
+    /* if napi-id based thread selection is not enabled, use
+     * default round-robin based thread selection
+     */
+	if (!settings.napi_id_based_thread_selection) {
+        selected_thread = select_next_thread();
+        goto out;
```
dormando: I might be wrong but the only uses of goto in the code are for error handling; or should be almost all of it.

might be okay with this case though..

thread.c
```diff
@@ -545,9 +545,125 @@ static void thread_libevent_process(int fd, short which, void *arg) {
...
+     */
+#ifndef SO_INCOMING_NAPI_ID
+#define SO_INCOMING_NAPI_ID 56
+#endif /* SO_INCOMING_NAPI_ID */
+    len = sizeof(napi_id);
+    if (getsockopt(sfd, SOL_SOCKET, SO_INCOMING_NAPI_ID,
+		   &napi_id, &len) == -1) {
+        perror("getsockopt SO_INCOMING_NAPI_ID");
+        /* Disable napi_id  based thread selection */
+        settings.napi_id_based_thread_selection = false;
```
dormando: is there no way to detect this earlier? can it fail permanently during runtime?

thread.c
```diff
@@ -564,11 +680,9 @@ void dispatch_conn_new(int sfd, enum conn_states init_state, int event_flags,
         return;
     }
 
-    int tid = (last_thread + 1) % settings.num_threads;
-
-    LIBEVENT_THREAD *thread = threads + tid;
+    LIBEVENT_THREAD *thread = NULL;
 
-    last_thread = tid;
+    select_thread(sfd, &thread);
```
dormando: This is a little nitpicky but this routine makes the common case convoluted and with too much overhead:
- function call, which then checks for a flag, which then makes a function call (common case! most users will not be using this feature).
- select_thread() has an out: label with a handler that checks and runs `select_next_thread`, but that's only used once? another time `select_next_thread` is called explicitly, then sent to out:. Should at least be consistent.

thread.c
kapatil: I can make that change, Absolutely no problem.

thread.c
kapatil: I will rewrite this function and eliminate goto

thread.c
kapatil: This will fails only if you arr running on older linux kernel 4.12 since SO_INCOMING_NAPI_ID socket option was added in 4.12 upstream linux kernel.

Once it fails, it will always fail during runtime, hence disabling napi_id based thread selection mechanism forever.

To detect it earlier, at least we have to make one call on accepted connection, hence it was done here and if fails - disable the napi_id based selection logic.

Please let me know if I am not answering to your question or if I misunderstood your comment

thread.c
kapatil: Agree, Once I rewrite that function, hopefully this comment will get addressed.

kapatil:
I will add documentation regarding when to use this optional feature, in terms of when it make sense to use this feature. I can provide the data w.r.t showing the performance delta with and without using this feature to prove its value.

Will also provide some doc pointers about linux kernel busy_polling feature. memcached already uses epoll per thread as a mechanism to see requests and then reads them from socket. Intel has extended epoll_wait to use kernel busy_polling internally. So epoll_wait doing busy_polling of network queue along with aligning sockets based on napi_ids improves memcached performance
(reduces latency, improved predictability by providing consistent request-response time,  improves throughput).

kapatil:
Can I get some review on  latest patch>

Added justification why someone should use this feature in help text and rewrite thread_selection function based on review comment and tries to fix minor nitpicks, etc..

dormando:
Hey,

Sorry; you should ask right when you need a review. otherwise I can't tell just by commits being pushed in. The description needs work. I'll give more feedback as soon as I can.

dormando:
Also according to travis the whitespace test fails. Do you run the test suite?

kapatil:
Thanks a bunch. Will stay tuned for your feedback.

thread.c
dormando: Can this test be done on the listening socket during startup?

thread.c
```diff
@@ -545,9 +545,101 @@ static void thread_libevent_process(int fd, short which, void *arg) {
...
+
+    /* If here means likely that number of unique NAPI_ID are
+     * more than the number of threads and that is fine.
+     * From this point onwards, disable "napi_id_based_thread_selection"
+     * and fallback to default thread selection mechanism
+     * This is needed so that at least application can still continue to work
+     * and apply it's own load-balancing technique
+     */
+    settings.napi_id_based_thread_selection = false;
+    return select_thread();
```
dormando: I feel like this will give users a mystery. The stat should be added to the stats_settings output, so users can tell if it's actually on or not. Further, we need to document or somehow log/warn/error since it looks like unless you set the worker threads to exactly match the napi_ids it'll just shut itself off immediately.

dormando:
Also, if you can please rebase if you haven't recently.

memcached.c
```diff
@@ -7724,6 +7724,16 @@ static void usage(void) {
...
+           "                          Why to use this feature:                             \n"
+           "                          The Linux Kernel natively supports busy_polling feature\n"
+           "                          which enables polling of NIC HW queue directly from application\n"
+           "                          thread context. It is also extended for epoll:epoll_wait syscall.\n"
+           "                          Using busy_polling feature has direct and visible impact on\n"
+           "                          application performance w.r.t request to response time. Busy_polling\n"
+           "                          feature provides lower latency and improved predicatablity by\n"
+           "                          providing consistent result (consistent and lower response time\n"
+           "                          even at 99.9 percent\n");
     return;
```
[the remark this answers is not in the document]
kapatil: I see. May be I can try to answer this way. Do you really want all the description in help?
I can certainly point to link about NAPI, what is busy_polling and epoll:busy_poll.
This is OS feature and it is not specific a network card.
This napi_id based busy_polling is linux kernel feature, kernel 4.12 and onwards supports epoll:busy_poll.

Other option is, can we put this info in commit message.

Let me think and get back to you.

thread.c
kapatil: Sure, let me see about stats and find ways to log/warn/errors about threads and napi_ids match

kapatil: Sure, let me look into that.

memcached.c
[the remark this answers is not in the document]
kapatil: DO you really want all this in help or in patch commit desc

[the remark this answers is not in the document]
dormando: I thought I was clear in that it's not good for the help option; I want a wiki page and a link or a doc file under doc.

To be absolutely clear: I need you to stop trying to explain this to me and explain it to the end users. A commit message doesn't come with the tarball and doesn't do any user any good.

memcached.c
kapatil: Agree, will fix it in next rev.

thread.c
kapatil: Yes. So you want to perform this test on listening socket and if fails, disable the feature completely and don;t do it for every accepted.child socket.

Sure. will make that change.

memcached.c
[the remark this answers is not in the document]
kapatil: Aah. I can put all this description in file under "doc" directory. Should I add in one of the existing file under doc or create brand new file e.g. napi_ids.txt? 

I am new to this memcached git tree, so please suggest.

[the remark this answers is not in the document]
dormando: To give even more point here: 100,000+ people could look at this help option, and if they don't understand it they'll be coming to me, not you :) So every inch we put here to reduce support burden goes a long way.

thread.c
dormando: I think it should try to bail out the entire application or loudly print at least. Unless there's a reason not to, a user trying to enable an option that then doesn't work should be communicated as early as possible. If we can check during the listen stage that should be early enough to bail out.

memcached.c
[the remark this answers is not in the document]
dormando: You can put a new file under doc/ and then refer to it in the help output. That's pretty common and any file under doc/ gets sent out with the tarball. I might adjust it to also have a URL but will do that myself.

dormando:
Just a head's up: I'm going to be tearing apart memcached.c into multiple files shortly. You may have to do significant rewiring to update your patch after a few days.

dormando:
closing this in favor of the other pr

PR #613, opened 2020-02-20
Optional SSL server session cache

LINKIWI:
Optionally enable server-only session caching with `SSL_CTX_set_session_cache_mode` if the binary is invoked with extended option `ssl_session_cache`. Useful for clients that support TLS session resumption (immediate use case is [mcrouter](https://github.com/facebook/mcrouter/blob/1c4b4838d7758d396fab7fa541c8f908d0b008ee/mcrouter/mcrouter_options_list.h#L379)).

This change also introduces a change in behavior to default the session cache to off. It also introduces a new gauge to keep track of new successful handshakes, `ssl_new_sessions`, enabled only when session caching is enabled.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -1634,6 +1634,9 @@ following additional statistics are available via the "stats" command.
 | ssl_handshake_errors           | 64u      | Number of times the server has |
 |                                |          | encountered an OpenSSL error   |
 |                                |          | during handshake (SSL_accept). |
+| ssl_new_sessions               | 64u      | When SSL session caching is    |
+|                                |          | enabled, the number of newly   |
+|                                |          | created server-side sessions.  |
 | time_since_server_cert_refresh | 32u      | Number of seconds that have    |
 |                                |          | elapsed since the last time    |
 |                                |          | certs were reloaded from disk. |
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -1637,6 +1637,8 @@ following additional statistics are available via the "stats" command.
 | ssl_new_sessions               | 64u      | When SSL session caching is    |
 |                                |          | enabled, the number of newly   |
 |                                |          | created server-side sessions.  |
+|                                |          | Available only when compiled   |
+|                                |          | with OpenSSL 1.1.1 or newer.   |
 | time_since_server_cert_refresh | 32u      | Number of seconds that have    |
 |                                |          | elapsed since the last time    |
 |                                |          | certs were reloaded from disk. |
--- a/memcached.c
+++ b/memcached.c
@@ -260,6 +260,7 @@ static void settings_init(void) {
     settings.udpport = 0;
 #ifdef TLS
     settings.ssl_enabled = false;
+    settings.ssl_session_cache = false;
     settings.ssl_ctx = NULL;
     settings.ssl_chain_cert = NULL;
     settings.ssl_key = NULL;
@@ -269,7 +270,6 @@ static void settings_init(void) {
     settings.ssl_ca_cert = NULL;
     settings.ssl_last_cert_refresh_time = current_time;
     settings.ssl_wbuf_size = 16 * 1024; // default is 16KB (SSL max frame size is 17KB)
-    settings.ssl_session_cache = false;
 #endif
     /* By default this string should be NULL for getaddrinfo() */
     settings.inter = NULL;
--- a/memcached.c
+++ b/memcached.c
@@ -3233,6 +3233,9 @@ static void server_stats(ADD_STAT add_stats, conn *c) {
 #endif
 #ifdef TLS
     if (settings.ssl_enabled) {
+        if (settings.ssl_session_cache) {
+            APPEND_STAT("ssl_new_sessions", "%llu", (unsigned long long)stats.ssl_new_sessions);
+        }
         APPEND_STAT("ssl_handshake_errors", "%llu", (unsigned long long)stats.ssl_handshake_errors);
         APPEND_STAT("time_since_server_cert_refresh", "%u", now - settings.ssl_last_cert_refresh_time);
     }
--- a/memcached.c
+++ b/memcached.c
@@ -269,6 +269,7 @@ static void settings_init(void) {
     settings.ssl_ca_cert = NULL;
     settings.ssl_last_cert_refresh_time = current_time;
     settings.ssl_wbuf_size = 16 * 1024; // default is 16KB (SSL max frame size is 17KB)
+    settings.ssl_session_cache = false;
 #endif
     /* By default this string should be NULL for getaddrinfo() */
     settings.inter = NULL;
@@ -3312,6 +3313,7 @@ static void process_stat_settings(ADD_STAT add_stats, void *c) {
 #endif
 #ifdef TLS
     APPEND_STAT("ssl_enabled", "%s", settings.ssl_enabled ? "yes" : "no");
+    APPEND_STAT("ssl_session_cache", "%s", settings.ssl_session_cache ? "yes" : "no");
     APPEND_STAT("ssl_chain_cert", "%s", settings.ssl_chain_cert);
     APPEND_STAT("ssl_key", "%s", settings.ssl_key);
     APPEND_STAT("ssl_verify_mode", "%d", settings.ssl_verify_mode);
@@ -8020,6 +8022,8 @@ static void usage(void) {
            "                          enables restartable cache (stop with SIGUSR1)\n");
 #ifdef TLS
     printf("-Z, --enable-ssl          enable TLS/SSL\n");
+    printf("-z, --enable-ssl-session-cache enable SSL server session cache to support session\n"
+           "                          resumption; requires SSL to be enabled\n");
 #endif
     printf("-o, --extended            comma separated list of extended options\n"
            "                          most options have a 'no_' prefix to disable\n"
@@ -8911,8 +8915,9 @@ int main (int argc, char **argv) {
 
     char *shortopts =
           "a:"  /* access mask for unix socket */
-          "A"  /* enable admin shutdown command */
+          "A"   /* enable admin shutdown command */
           "Z"   /* enable SSL */
+          "z"   /* enable SSL server session cache */
           "p:"  /* TCP port number to listen on */
           "s:"  /* unix socket path to listen on */
           "U:"  /* UDP port number to listen on */
@@ -8952,6 +8957,7 @@ int main (int argc, char **argv) {
         {"unix-mask", required_argument, 0, 'a'},
         {"enable-shutdown", no_argument, 0, 'A'},
         {"enable-ssl", no_argument, 0, 'Z'},
+        {"enable-ssl-session-cache", no_argument, 0, 'z'},
         {"port", required_argument, 0, 'p'},
         {"unix-socket", required_argument, 0, 's'},
         {"udp-port", required_argument, 0, 'U'},
@@ -8997,15 +9003,20 @@ int main (int argc, char **argv) {
             /* enables "shutdown" command */
             settings.shutdown_command = true;
             break;
+#ifdef TLS
         case 'Z':
             /* enable secure communication*/
-#ifdef TLS
             settings.ssl_enabled = true;
+            break;
+        case 'z':
+            settings.ssl_session_cache = true;
+            break;
 #else
+        case 'Z':
+        case 'z':
             fprintf(stderr, "This server is not built with TLS support.\n");
             exit(EX_USAGE);
 #endif
-            break;
         case 'a':
             /* access for unix domain socket, as octal mask (like chmod)*/
             settings.access= strtol(optarg,NULL,8);
@@ -9858,6 +9869,14 @@ int main (int argc, char **argv) {
 
 
 #ifdef TLS
+    /*
+     * SSL parameter validation
+     */
+    if (settings.ssl_session_cache && !settings.ssl_enabled) {
+        fprintf(stderr, "-z (ssl_session_cache) requires -Z (ssl_enabled)\n");
+        exit(EX_USAGE);
+    }
+
     /*
      * Setup SSL if enabled
      */
--- a/memcached.c
+++ b/memcached.c
@@ -9873,7 +9873,7 @@ int main (int argc, char **argv) {
      * SSL parameter validation
      */
     if (settings.ssl_session_cache && !settings.ssl_enabled) {
-        fprintf(stderr, "-z (ssl_session_cache) requires -Z (ssl_enabled)\n");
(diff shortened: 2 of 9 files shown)


dormando:
nice, thanks! Will take a look soon. Have you tested TLS tickets at all?

LINKIWI:
Thanks! I've only tested session ID based resumption for now in an end-to-end environment with mcrouter as the client. Seems like supporting ticket based resumption is a bit more complicated, and in truth I'm probably not familiar enough with OpenSSL APIs to implement it here.

ankita25:
Based on https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_session_id_context.html, 

`If the session id context is not set on an SSL/TLS server and client certificates are used, stored sessions will not be reused ` . So we also need to set SSL_CTX_set_session_id_context

dormando:
I'm a bit busy to validate this right now, might be another week or two. Super bonus points to anyone who adds tests + fixes the context issue :) I'm pretty sure I've written TLS resumption tests ten years ago but my own mind's full of cobwebs.

LINKIWI:
I had some time to take a closer look at this, seems like this is actually a noop change in behavior since server-side session cache is already enabled by default. I confirmed this with a call to `SSL_CTX_get_session_cache_mode`, and documentation confirms this as well:

> The default mode is SSL_SESS_CACHE_SERVER.

So there might not actually be too much value in merging this change. We could offer the ability to explicitly disable the session cache if desired, if you think there is value in providing such an option.

I've been able to verify session resumption behavior with the openssl binary itself and checking output for new and reused sessions, though I'm not sure how you feel about a unit test invoking a binary and screen-scraping the output for assertions.

```
$ openssl s_client -key t/client_key.pem -cert t/client_crt.pem -CAfile t/cacert.pem -connect localhost:11211 -no_ticket -sess_out /tmp/sess
$ openssl s_client -key t/client_key.pem -cert t/client_crt.pem -CAfile t/cacert.pem -connect localhost:11211 -no_ticket -sess_in /tmp/sess
```

tharanga:
@LINKIWI thanks for catching this. I checked my notes on the initial implementation, the goal was not to enable session resumption by default (I wasn't sure about vulnerabilities and how to invalidate sessions with automatic cert refresh without flushing them all), but looks like I missed it. 

I propose to disable the session resumption by default and keep this changeset to enable if needed.

LINKIWI:
Thanks @tharanga. Sounds reasonable to me, I've updated this patch to explicitly set `SSL_SESS_CACHE_OFF` by default but otherwise set `SSL_SESS_CACHE_SERVER` (OpenSSL default) when the option is passed.

tharanga:
@LINKIWI LGTM. re: unit tests, looks like you can use `IO::Socket::SSL->new(..` with `SSL_session_cache` created from `IO::Socket::SSL::Session_Cache->new` with `SSL_session_key`. More details are here : http://manpages.ubuntu.com/manpages/trusty/man3/IO::Socket::SSL.3pm.html

dormando:
This looks better, but I still want some tests before merging it. Would one of you folks mind doing that soon? :) I'm about to do a lot of work in memcached.c which would cause a conflict.

LINKIWI:
Sure, I'll try to get something up for tests in the next few days. Thanks!

LINKIWI:
I've spent some time with this, but unfortunately I'm not familiar with Perl so I would appreciate some help on the necessary modifications to `lib/MemcachedTest.pm`. In particular, it seems like `new_sock` is decoupled from the initial `$conn = ...` so I'm not sure of the best way to set a global `Session_Cache` object that can be shared among all invocations.

My proposal for the actual unit test is something like this, which basically just makes sure total number of TCP connections goes up but new SSL sessions is unchanged (this takes advantage of a new gauge `ssl_new_sessions` introduced by this PR):

```perl
my $server;
my $sock;
my $stats;

$server = new_memcached();
$stats = mem_stats($server->sock);
is($stats->{ssl_new_sessions}, undef, "new SSL sessions not recorded when session cache is disabled");

$server = new_memcached("-o ssl_session_cache");
$stats = mem_stats($server->sock);
my $initial_ssl_sessions = $stats->{ssl_new_sessions};
my $initial_total_conns = $stats->{total_connections};
$stats = mem_stats($server->new_sock);
cmp_ok($stats->{total_connections}, '>', $initial_total_conns, "new connection is established");
is($stats->{ssl_new_sessions}, $initial_ssl_sessions, "SSL session is reused");
```

LINKIWI:
Added `ssl_session_resumption.t`. I have no experience with Perl so please let me know if needs additional work. Thanks @tharanga @dormando 

```
$ SSL_TEST=1 perl t/ssl_session_resumption.t 
ok 1 - new SSL sessions not recorded when session cache is disabled
ok 2 - client-side session cache is noop in establishing a new connection
ok 3 - client-side session cache is unused
Signal handled: Interrupt.
ok 4 - initial connection is established
ok 5 - successful new SSL session
ok 6 - new connection is established
ok 7 - no new SSL sessions are created on the server
ok 8 - client-persisted session is reused
1..8
Signal handled: Interrupt.
```

dormando:
Thanks! will look as soon as I can. For an added bonus; please base branches off of 'next' (and target merge to 'next'). I tried to fix the defaults a few weeks ago but failed, sorry. I fixed a lot of flaky tests in the next branch.

tls.c
```diff
@@ -162,6 +162,17 @@ int ssl_init(void) {
         exit(EX_USAGE);
     }
 
+    // Optional session caching; default disabled.
+    if (settings.ssl_session_cache) {
+        SSL_CTX_sess_set_new_cb(settings.ssl_ctx, ssl_new_session_callback);
```
tharanga: Is there a reason why we don't need/can't have the `ssl_new_sessions` counter when the session cache is not enabled? I guess this will be same as the number of connections made in that case.

LINKIWI: The set-session callback isn't actually registered internally in OpenSSL unless session caching is enabled. So when the counter is enabled when session caching is off, it stays at 0. This behavior seems consistent with the documentation: https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_sess_set_new_cb.html

LINKIWI:
Changed base pointer to `next`. Thanks

This change actually conflicts with #627 since both PRs add a new TLS stat; if we plan to merge both it might make more sense to merge #627 first and I can resolve the conflict + add a line to doc/protocol.txt in this PR after rebasing.

memcached.c
```diff
@@ -7984,6 +7989,8 @@ static void usage(void) {
            "   - ssl_ca_cert:         PEM format file of acceptable client CA's\n"
            "   - ssl_wbuf_size:       size in kilobytes of per-connection SSL output buffer\n"
            "                          (default: %u)\n", settings.ssl_wbuf_size / (1 << 10));
+    printf("   - ssl_session_cache:   enable server-side SSL session cache, to support session\n"
```
tharanga: @dormando do we have to prefix this flag with `enable`?

tharanga:
@LINKIWI test looks good to me as that counter is a direct function of whether session cache enabled or not. You have to install a newer version of `libio-socket-ssl-perl` or disable tests that are checking the client side session caching as `get_session_reused` is only available in newer versions of `IO-Socket-SSL`.

memcached.c
dormando: I don't think so actually. Double checking the help output, convention is to just name the feature. Disabling it gets a `no_` prefix though.

dormando:
that makes sense. I think @tharanga's checking the counter PR shortly, then I'll finalize and merge if we all agree on it.

dormando:
think we'll miss 1.6.1 on this (too much stuff queued, few things to fix/review here still) but it'll hit `next` soon.

doc/protocol.txt
```diff
@@ -1634,6 +1634,9 @@ following additional statistics are available via the "stats" command.
 | ssl_handshake_errors           | 64u      | Number of times the server has |
 |                                |          | encountered an OpenSSL error   |
 |                                |          | during handshake (SSL_accept). |
+| ssl_new_sessions               | 64u      | When SSL session caching is    |
+|                                |          | enabled, the number of newly   |
+|                                |          | created server-side sessions.  |
 | time_since_server_cert_refresh | 32u      | Number of seconds that have    |
```
tharanga: Thanks @LINKIWI for adding `time_since_server_cert_refresh`

LINKIWI:
Hm, I'm seeing different behavior between OpenSSL 1.1.0 (CI environment) and OpenSSL 1.1.1 (my local environment). I'll dig deeper this week.

dormando:
Thanks for putting the work in! I'll check in on this too. what's the difference you're seeing?

LINKIWI:
Thanks! Seems like the combination of OpenSSL 1.1.0 + latest IO::Socket::SSL properly reports that the session is reused on the client side, but doesn't increment ssl_new_sessions on the server (i.e., the new session callback isn't called). However, when testing manually with `openssl s_client`, I can see the server-side stat incremented properly.

The combination of OpenSSL 1.1.1 + latest IO::Socket::SSL behaves as I expect, per the unit test (ssl_new_sessions is incremented on the server, and client-side reports that the session is reused). It also continues to behave correctly with `openssl s_client`.

There's a possibility that OpenSSL 1.1.1 introduces some bug fixes around this. Though seeing as session resumption seems to work fine in both versions, one option forward is to skip the assertion `ssl_new_sessions > 0` in the unit test.

tharanga:
That's strange. Latest IO::Socket::SSL shouldn't impact the memcached, it should continue to use OpenSSL 1.1.0, where that callback is used.
Do you need those client side checks? I believe we can rely on the server session count for the correctness.

LINKIWI:
The client-side check is not strictly necessary, but the issue is that the server-side session count seems to only be incremented properly under OpenSSL 1.1.1. See for example the test failure here: https://travis-ci.com/github/memcached/memcached/builds/153608008

LINKIWI:
So it seems like the behavior is actually different between OpenSSL 1.1.0 and 1.1.1.

1.1.0: https://github.com/openssl/openssl/blob/b2758a2292aceda93e9f44c219b94fe21bb9a650/ssl/ssl_lib.c#L2977
1.1.1: https://github.com/openssl/openssl/blob/8658feddea6aef5cf5cbb1cfbf6b1817fa432051/ssl/ssl_lib.c#L3641

From what I gather, 1.1.1 always invokes the callback (see inlined comment in the source above) while 1.1.0 only invokes it if external session cache storage is enabled (`SSL_SESS_CACHE_NO_INTERNAL_STORE`) or internal caching failed.

Given this, I'd like to propose skip the server-side counter assertion in the unit test (`assert ssl_new_sessions > 0 after handshake`) and adding a note in the docs that the counter will only be incremented for new sessions in OpenSSL versions 1.1.1 and above. We can consider re-enabling the assertion if we upgrade the base CI image to Bionic or upgrade OpenSSL. Any thoughts on this?

t/lib/MemcachedTest.pm
```diff
@@ -384,11 +384,15 @@ sub new_sock {
     if ($self->{domainsocket}) {
         return IO::Socket::UNIX->new(Peer => $self->{domainsocket});
     } elsif (MemcachedTest::enabled_tls_testing()) {
-        return eval qq{ IO::Socket::SSL->new(PeerAddr => "$self->{host}:$self->{port}",
-                                    SSL_verify_mode => IO::Socket::SSL::SSL_VERIFY_NONE,
-                                    SSL_cert_file => '$client_crt',
-                                    SSL_key_file => '$client_key');
-                                    };
+        my $ssl_session_cache = shift;
```
dormando: the eval was here because, if missing, the test suite would blow up with a compile error if you're missing the TLS related perl packages, even if you're not building memcached with `--enable-tls`. You can see how this works in a few other sections that call the ::SSL stuff.

You can still shift in these variables and then use the original return eval; it had $client_key and stuff in it already.

t/ssl_session_resumption.t
```diff
@@ -0,0 +1,59 @@
...
+if (!enabled_tls_testing()) {
+    plan skip_all => 'SSL testing is not enabled';
+    exit 0;
+}
+
+my $server;
+my $sock;
+my $stats;
+
+my $session_cache = IO::Socket::SSL::Session_Cache->new(1);
```
dormando: Think this needs to wrap with an eval as well, or it will fail to compile on machines missing SSL perl modules.

sorry this is so complicated :/ I might just do this myself if you can't soon. Need to spin up a VM missing the modules or something.

dormando:
sorry I've been so slow to poke at this :(

t/ssl_session_resumption.t
LINKIWI: No worries! Thanks for the review, I've updated to wrap this in eval as well. Please feel free to push additional commits to this branch as needed.

I was able to run this in a Docker container without the SSL Perl module installed and it failed properly without crashing:

```
1..0 # SKIP SSL testing is not enabled
```

t/ssl_session_resumption.t
```diff
@@ -0,0 +1,59 @@
...
+
+### Enabled SSL session cache
+
+$server = new_memcached("-o ssl_session_cache");
+# Support for session caching in IO::Socket::SSL for TLS v1.3 is incomplete.
+# Here, we will deliberately force TLS v1.2 to test session caching.
+$sock = $server->new_sock($session_cache, 'TLSv1_2');
+$stats = mem_stats($sock);
+cmp_ok($stats->{total_connections}, '>', 0, "initial connection is established");
+SKIP: {
```
tharanga: (optional) : as `ssl_new_sessions` is the only indicator of server side sessions, we may enable the test with the wrong output (documents the 1.1.0 behavior in the test) and a comment. In this way, as we move to OpenSSL 1.1.1, we'll see this test is failing and whoever moves will fix it.

tharanga: NM. This will cause user builds to fail if their env has the OpenSSL 1.1.1. I guess we can leave this as it is and fix it at sometime in the future.

dormando:
I bumped travis to bionic to simplify this a bit (and cut several minutes off the build/test run time...).

Think I'm out of complaints. Do you want to rebase/squash this down or should I do this for you?

LINKIWI:
Great! I believe Bionic ships with OpenSSL 1.1.1 already so I think I can remove the skipped tests now and the extra Perl installation in .travis.yml as well. I'll rebase against `next` and make those changes.

dormando:
You should keep the skipped test unless you want to add version detection in there; as users may be on older versions, the test would become spurious.

LINKIWI:
Rebased against `next`. I reverted all my changes to `.travis.yml` from this branch. Let me know if you have any remaining concerns

dormando:
released in 1.6.3 (I did a squash down for you)

PR #615, opened 2020-02-22
timedrun proper signal handler initialization.

devnexen:
being 'nicer' when forgetting arguments.


dormando:
thanks!

dormando:
out in 1.6.0!

Issue #546, opened 2019-10-01
meta command follow ups


dormando:
Meta features (#484) are out in 1.5.19. There'll be a trickle of changes hopefully with every release to smooth things out. Keeping track of them with this issue.

- [ ] LOGGER_LOG() entries specific to meta commands
- [ ] mget input: return value if CAS match or not match (maybe invert `v` if supplied cas matches?)
- [x] meta get should work without flags supplied
- [x] meta delete has extraneous newline (meta get probably has the same bug)
- [x] O and k should be reflected on EN/miss.
- [x] remove the "meta_response_old" option.

edit: removed a bunch of completed items/notes.
edit2: (final?) todo list.
edit3: updated TODO list again.

dormando:
Trying to implement a simplified client with this in a few more languages. Just ran into something backwards.

For sending flags to the server, capital letter flags always consume a token.
However for flags received by the client, there's no blind way of determining which flags associate with tokens. You have to know the full list of flags ahead of time; which means adding a new flag that consumes a token could be dangerous.

Up until now I was writing the clients with tokens being passed around blindly... but this has accidentally made parsing the value part of the response impossible. In all my other tests I'd been looking for `EN\r\n` and overlooked this when reviewing. As did everyone else :)

Clients either need to test for "was there a v flag?", and if so, "which token position is the size?", or blind parse by looking for `EN\r\n`. Options:

1. Box capital letter flags for any flag which _returns_ a token. If 'v' and no 's', look for `EN\r\n`. If 's', pull by length. Could also box by which half of alphabet.
2. If 'v', return code is: `VA <size> <returnflags> <tokens>`. If no 'v', return code is: `HD <returnflags> <tokens>`, VA for value, HD for header.
3. Do 2, but also box token'ed flags so tokens can be assigned to flags without the client knowing future flags. This is less useful with the VA/HD change, but would still allow for building map/hash/dict of flag -> token. The caller should usually know what it's expecting and shouldn't need to look at the flags at all, but it should be possible to build this list without thinking about it.

Hmm. Another possibility would be to always have capital means "contains a token". So response flags get capitalized when they have a token. l -> L and so on.

dormando:
```
a: none
A:
b: none
B:
c: return CAS
C: set/use supplied CAS
d: none
D:
e: none
E:
f: get flags
F: set flags
g: none
G:
h: return if item was hit
H: none
i: none
I: invalidation (FIX)
j: none
J:
k: return key as token
K: none
[...]

PR #618, opened 2020-03-02
Meta protocol updates

dormando:
Commits fleshing out the TODO's from #546 

- fixes a refcount leak (oops!)
- some missing tests/cleanup
- refactors some core code
- adds the mode switch to metaset: set, add, cas, replace, append, prepend are now all just the `ms` command.
- adds meta arithmetic (`ma`) command.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -655,9 +655,9 @@ may decide to retry later or take some other action.
 Meta Set
 --------
 
-The meta set command a generic command for storing data to memcached. Based on
-the flags supplied, it can replace the commands: "set", "cas". as well as
-adding new options.
+The meta set command a generic command for storing data to memcached. Based
+on the flags supplied, it can replace all storage commands (see token M) as
+well as adds new options.
 
 ms <key> <flags>*\r\n
 
@@ -702,6 +702,7 @@ The flags used by the 'ms' command are:
 - q: use noreply semantics for return codes
 - S(token): size of <data block> to store
 - T(token): Time-To-Live for item, see "Expiration" above.
+- M(token): mode switch to change behavior to add, replace, append, prepend
 
 The flags are now repeated with detailed information where useful:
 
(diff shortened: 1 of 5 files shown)


dormando:
@shivnagarajan you might be interested in this

memcached.c
```diff
@@ -4136,18 +4209,62 @@ static int _meta_flag_preparse(token_t *tokens, const size_t ntokens, struct _me
...
+                }
+                break;
+            case 'J': // marithmetic initial value
+                if (!safe_strtoull(tokens[i].value+1, &of->initial)) {
+                    *errstr = "CLIENT_ERROR invalid numeric initial value";
+                    of->has_error = 1;
+                }
+                break;
+            case 'D': // marithmetic delta value
+                if (!safe_strtoull(tokens[i].value+1, &of->delta)) {
```
shivnagarajan: So the `M` mode option for `ma` accepts `I, `D`, `+` or `-`,  and assumes implicitly that the `D` option will not accept negative values (which `safe_strtoull` handles and appropriately fails).

But the protocol description does say that the argument for `-D` is an `unsigned int`, but `safestrtoull` will happily accept a string value > an `unsigned int`. Maybe this is not that important, but we could easily adapt the protocol documentation to just say `will accept an argument that is a positive unsigned long long` maybe? 🤷‍♂

memcached.c
```diff
@@ -4563,29 +4642,51 @@ static void process_mset_command(conn *c, token_t *tokens, const size_t ntokens)
...
+    // APPEND/PREPEND allow a simplified CAS check.
+    if (of.has_cas && (comm == NREAD_SET || comm == NREAD_REPLACE)) {
+        comm = NREAD_CAS;
+    }
 
-    // TODO: can we treat vlen as unsigned? :(
-    if (vlen < 0 || vlen > (INT_MAX - 2)) {
-        // TODO: specific error.
+    // We attempt to process as much as we can in hopes of getting a valid and
+    // adjusted vlen, or else the data swallow after error will be for 0b.
```
shivnagarajan: typo `swallow` -> `swallowed`

doc/protocol.txt
```diff
@@ -727,12 +728,23 @@ See description under 'Meta Get'
...
 
 Errors are always returned.
 
+- M(token): mode switch. Takes a single character for the mode.
+
+E: "add" command. LRU bump and return NS if item exists. Else
+add.
+A: "append" command. If item exists, append the new value to its data.
+P: "prepend" command. If item exists, prepend the new value to its data.
+S: "set" command. The default mode, added for completeness.
```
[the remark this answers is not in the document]
dormando: Fixed, thanks!

doc/protocol.txt
```diff
@@ -795,6 +807,88 @@ When marking an item as stale with 'I', the 'T' flag can be used to update the
...
+- T(token): update TTL
+
+On success, sets the remaining TTL to the supplied value.
+
+-M(token): mode switch. Takes a single character for the mode.
+
+I: Increment mode (default)
++: Alias for increment
+D: Decrement mode
+-: Alias for decrement
```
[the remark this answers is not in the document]
dormando: It does have one, I missed the doc. In meta commands "noreply" is a separate q flag. There's even a test for it in metaget.t :) Updated locally now.

memcached.c
dormando: Nice catch! This was supposed to match the incr/decr documentation as it handles arguments the same way. I've updated a few words to explicitly say 64-bit unsigned integer.

memcached.c
dormando: fixed

dormando:
Found another bug this morning + doc fixes from review comments. Think I'm done with this and ready to merge into `next`.

dormando:
Thanks :) I'm excited to get this out there.

PR #620, opened 2020-03-08
Fix to handle negative exptime as common function

minkikim89:
Every way to handle negative expiretime is different. 
I united as one.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -199,6 +199,13 @@ static void maxconns_handler(const int fd, const short which, void *arg) {
 
 #define REALTIME_MAXDELTA 60*60*24*30
 
+/* Negative exptimes can underflow and end up immortal. realtime() will
+   immediately expire values that are greater than REALTIME_MAXDELTA, but less
+   than process_started, so lets aim for that. */
+#define EXPTIME_TO_POSITIVE_TIME(exptime) (exptime < 0) ? \
+        REALTIME_MAXDELTA + 1 : exptime
+
+
 /*
  * given time value that's either unix time or delta from current unix time, return
  * unix time. Use the fact that delta can't exceed one month (and real time value can't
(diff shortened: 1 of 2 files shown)


dormando:
Thanks! Unfortunately I just modified some of this same code in another branch that I just merged. You'll have to update before I can review. It might be better as a macro since it can be a hot spot, too.

minkikim89:
@dormando 
I fixed. Please review it

memcached.c
```diff
@@ -3911,7 +3918,7 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
                 goto stop;
             }
 
-            it = limited_get(key, nkey, c, exptime, should_touch, DO_UPDATE, &overflow);
+            it = limited_get(key, nkey, c, realtime(exptime), should_touch, DO_UPDATE, &overflow);
```
dormando: move the realtime() conversion back to the if (should_touch) section. no reason to keep converting it.

memcached.c
```diff
@@ -4152,10 +4160,8 @@ static int _meta_flag_preparse(token_t *tokens, const size_t ntokens,
                     *errstr = "CLIENT_ERROR bad token in command line format";
                     of->has_error = 1;
                 } else {
-                    if (tmp_int < 0) {
-                        tmp_int = REALTIME_MAXDELTA + 1;
-                    }
-                    of->autoviv_exptime = realtime(tmp_int);
+                    tmp_time = EXPTIME_TO_POSITIVE_TIME(tmp_int);
+                    of->autoviv_exptime = realtime(tmp_time);
```
dormando: of->autoviv_exptime = realtime(EXPTIME_TO_POSITIVE_TIME(tmp_time)) ?

minkikim89:
@dormando 
I fixed it. And I added test case.

dormando:
Looks like the tests are failing on travis. Did you run them locally?

minkikim89:
@dormando 
I made a mistake. It works well now.

dormando:
merged into next, thanks!

dormando:
Fwiw though, I would accept even more tests (make another PR): you added a test for the `set` code path, which was functionally equivalent here (good to test), you also changed touch and gat so those should have specific tests so you can see the before/after impact.

minkikim89:
OK. I'll do it later

dormando:
out in 1.6.1!

PR #622, opened 2020-03-09
fix multiple definition of 'crc32c' (GCC 10)

remicollet:


dormando:
Thanks! I need to get a buildbot with gcc-10..

dormando:
out in 1.6.1!

PR #624, opened 2020-03-10
Needs to allow this syscall as well, caught on ARM.

devnexen:


dormando:
Thanks!

dormando:
out in 1.6.1!

PR #625, opened 2020-03-13
proto: clean process_command()

dormando:
Pulled almost all of the inlined commands out of the function. Also
abstract out the token count checks.

Meta commands have a special fast handler

also reordered/restructured the rest of the commands. cleaner code and
should be same/similar performance for the common commands.

Need to do a quick verification bench and wait a bit to see if I still like the token macros later.

edit: also need to finish TLS mode

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -2932,10 +2932,7 @@ typedef struct token_s {
 
 #define WANT_TOKENS(ntokens, min, max) \
     do { \
-        if (min != -1 && ntokens < min) { \
-            out_string(c, "ERROR"); \
-            return; \
-        } else if (max != -1 && ntokens > max) { \
+        if ((min != -1 && ntokens < min) || (max != -1 && ntokens > max)) { \
             out_string(c, "ERROR"); \
             return; \
         } \
@@ -6099,7 +6096,6 @@ static void process_command(conn *c, char *command) {
             process_stat(c, tokens, ntokens);
         } else if (strcmp(tokens[COMMAND_TOKEN].value, "shutdown") == 0) {
 
-            WANT_TOKENS_OR(ntokens, 2, 2);
             process_shutdown_command(c);
         } else if (strcmp(tokens[COMMAND_TOKEN].value, "slabs") == 0) {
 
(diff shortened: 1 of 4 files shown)


memcached.c
```diff
@@ -2914,6 +2914,33 @@ typedef struct token_s {
...
+        if (min != -1 && ntokens < min) { \
+            out_string(c, "CLIENT_ERROR wrong number of command tokens"); \
+            return; \
+        } else if (max != -1 && ntokens > max) { \
+            out_string(c, "CLIENT_ERROR wrong number of command tokens"); \
+            return; \
+        } \
+    } while (0)
+
+#define WANT_TOKENS_OR(ntokens, a, b) \
```
minkikim89: we need this macro now?
Anything implemented with this macro can be replaced with WANT_TOKENS.
For example, WANT_TOKENS_OR(ntokens, 6, 7) == WANT_TOKENS(ntokens, 6, 7)

dormando: those're all new. in my first pass I had everything with just WANT_TOKENS(), but most of the paths were originally just two ||'s or a single >= so I added the extra two macros to not add significant slowdown.

memcached.c
```diff
@@ -5750,321 +6019,162 @@ static void process_command(conn *c, char *command) {
...
+                conn_set_state(c, conn_mwrite);
+                break;
+            case 'a':
+                process_marithmetic_command(c, tokens, ntokens);
+                break;
+            case 'e':
+                process_meta_command(c, tokens, ntokens);
+                break;
+        }
+    } else if (tokens[COMMAND_TOKEN].value[0] == 'g') {
```
minkikim89: if input string is just a space(" "), segfault occurs here.
Please check it.

dormando: good catch! I left out the default: there. missing coverage in t/metaget.t

dormando:
purposefully not putting this into 1.6.1. been glancing at it in case I want to rip out more of the token checks or something.

dormando:
performance sucks, apparently. The ntokens shortcutting did a good job of keeping the strcmp's down in the old code.

Going to restructure it a bit more aggressively since we've got an "it might add up in latency" level slowdown for the non-get commands. Should also find a solution for the resp memset being too large since that's showing up under perf significantly.

dormando:
to test I took the "touch" command and ran mc-crusher against an empty cache. Touch is halfway down the if/else tree. I compared `next` and this branch. For `next` I also moved the metaget commands to the bottom of the tree, since adding them had slowed things down a bit already.

Two mc threads vs 4 memcached worker threads. misses so I could see the overhead of the parse tree more clearly. ~600k misses/sec for modified `next` and ~700k for this branch after the fix I just pushed. So now it should be faster than it ever was for the common commands.

dormando:
happy enough with this for now. going to merge it into `next` soon.

dormando:
released in 1.6.3

PR #627, opened 2020-03-15
ssl_errors stat

LINKIWI:
This PR proposes the addition of new stats key `ssl_errors` to keep track of the number of times the server encounters an OpenSSL error during `SSL_accept`.

Since we only permit TLS v1.2+ clients, I tested by forcing TLS 1.1 on client-side and observing `ssl_errors` increment on each accept failure:

```
$ openssl s_client -tls1_1 -key t/client_key.pem -cert t/client_crt.pem -CAfile t/cacert.pem -connect localhost:11211
```

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -1621,6 +1621,23 @@ The value of the "state" stat may be one of the following:
 |                | sending back multiple lines of response data).            |
 |----------------+-----------------------------------------------------------|
 
+TLS statistics
+--------------
+
+TLS is a compile-time opt-in feature available in versions 1.5.13 and later.
+When compiled with TLS support and TLS termination is enabled at runtime, the
+following additional statistics are available via the "stats" command.
+
+|--------------------------------+----------+--------------------------------|
+| Name                           | Type     | Meaning                        |
+|--------------------------------+----------+--------------------------------|
+| ssl_handshake_errors           | 64u      | Number of times the server has |
+|                                |          | encountered an OpenSSL error   |
+|                                |          | during handshake (SSL_accept). |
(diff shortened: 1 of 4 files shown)


dormando:
thanks for this!

Stats also need to be noted in doc/protocol.txt - I wonder if it should be named tls_handshake_errors instead, given where the stat is?

LINKIWI:
Sounds good to me! I think ssl_handshake_errors would be more consistent with the other user-facing TLS terminology like the extended options ssl_key, ssl_ca_cert, etc.

Do you prefer a separate section in protocol.txt for TLS-related stats since it is a compile time opt-in feature? Or would you prefer adding it to the existing table at https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L1179 ?

dormando:
can you base this branch against 'next' ?

dormando:
@tharanga any thoughts on this?

dormando:
Looks like your `next` branch is a bit behind upstream, which is causing the travis errors.

tharanga:
LGTM. Looks like you can add a better test once #613 is merged as you've added the ability to specify the protocol version to a SSL client.
Looks like there's a bug in the original code I added at https://github.com/memcached/memcached/blob/76419985d44dcdb8bd309c4c2589e7dc6f1d69b5/memcached.c#L7024. It should be `if (ret <= 0) {` as according to https://www.openssl.org/docs/man1.1.0/man3/SSL_accept.html `0` is an error condition as well.

dormando:
whoo. you caught my hand right over the release button. should I merge this up and fix the bug you noted here?

tharanga:
+1 @dormando, go for it. Thanks!

dormando:
out in 1.6.1!

Issue #633, opened 2020-03-24
duplicate symbols found when building with LTO


darix:
```
 memcached.h:482:24: error: type of 'settings' does not match original declaration [-Werror=lto-type-mismatch]
  482 | extern struct settings settings;
      |                        ^
testapp.c:48:17: note: 'settings' was previously declared here
   48 | struct settings settings;
      |   
```

hash.c some how uses the settings struct defined in testapp.c

dormando:
@kanak - looks like hash.c pulls in memcached.h and testapp.c has a struct settings... and LTO is a bit insane. any chance you'd want to fiddle with this?

darix:
not sure if it pulls the header ... but hash.c is in the source list for memcached ... that's why it sees both symbols when trying to link memcached.

kanak:
I'd be happy to take a look. I adapted the trick we were using for `struct conn` in testapp.c to get the hash settings in here. 

What compiler flags do i pass to reproduce the warnings? `-flto`?

darix:
-flto=auto is what we use.

kanak:
Was able to repro it, here's the problem I'm running into:

* `stats_prefix` test needs to do `hashpower_init` because it uses the hash table from `assoc`. `hashpower_init` sets a field in `struct settings`. `struct settings` is defined in `memcached.h`.

* lto is complaining that we have a settings defined with fewer fields in testapp.c than in memcached.h .

* OK, so why don't we just include memcached.h ? 

* That breaks `struct conn` which is also defined with fewer fields and `STATS_LOCK` function which is re-implemented.

@dormando, what do you think we should do? Maybe pull out the definition of the settings struct into its own header so we can pull that in without pulling everything else in?

kanak:
@darix and @dormando : I have a possible fix, in https://github.com/memcached/memcached/pull/634 . 

Tested locally with GCC 9.2.1 with `-flto=auto`.

@dormando: Are you open to breaking memcached.h up into several header files? The settings struct needs a few #define and typedefs, I just moved them into the file which is ugly.

dormando:
I'm thinking about it. Will take a closer look at that file.

dormando:
Wondering if it'd be less dramatic to just import jenkins hash and call that directly? Since it's hard-coding the jenkins hash anyway, the only point to pulling hash.c is to get the selector. I see calls to hash() in testapp.c but nothing from assoc.c itself?

kanak:
[...]

PR #634, opened 2020-03-25
Proof of Concept fix for LTO warning

kanak:
https://github.com/memcached/memcached/issues/633

* Pulled out settings.h into its own header.
* settings struct relies on a few definitions that were in memcached,
for the purposes of this proof of concept, moved them to settings.h .
I think want to define more headers?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ noinst_PROGRAMS = memcached-debug sizes testapp timedrun
 
 BUILT_SOURCES=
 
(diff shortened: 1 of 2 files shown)


dormando:
looks like this broke compile for TLS mode

kanak:
Sorry about that I did `make test` but did not try with TLS. will fix and push an update.

darix:
works for me. with TLS

darix:
JFYI: this patch also fixes the build problems with -fno-common

dormando:
released in 1.6.4

PR #638, opened 2020-03-30
Fix undefined behavior when compiling with -D_FORTIFY_SOURCE=2

tstellar:
When compiling with -D_FORTIFY_SOURCE=2, printf is defined as a macro,
which makes the #ifdef TLS block in the usage() function of memcached.c
undefined behavior.

This warning was caught by compiling with clang:

memcached.c:7829:2: error: embedding a directive within macro arguments has undefined behavior [-Werror,-Wembedded-directive]
 ^
memcached.c:7832:2: error: embedding a directive within macro arguments has undefined behavior [-Werror,-Wembedded-directive]
 ^

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -7825,12 +7825,13 @@ static void usage(void) {
            "-s, --unix-socket=<file>  UNIX socket to listen on (disables network support)\n"
            "-A, --enable-shutdown     enable ascii \"shutdown\" command\n"
            "-a, --unix-mask=<mask>    access mask for UNIX socket, in octal (default: %o)\n"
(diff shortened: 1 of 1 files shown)


dormando:
Huh, ok. Thanks!

tstellar:
Just noticed a typo in the patch, will updated.

dormando:
had to write my own patch just now. had a conflict in this code and I reordered something to cut down on the ifdefs that were just added. It's now uploaded.

Thanks!

dormando:
released in 1.6.4

PR #641, opened 2020-04-02
sasl build fix proposal.

devnexen:
on some systems, shared libraries do not contain symbols
thus build failing for wrong reasons.
here just trying to compile dumb code to check it instead.


dormando:
Thanks!

Mind doing me a favor? I've been editing the commit messages from a lot of your pulls, mind taking a look and switching your format?

Also a hint in the message as to what "some systems" means is useful to people reading/reviewing.

dormando:
Pushed to next; commit message is closer! :)

Needs an empty line between the "Title" and the message body. title should be a descriptive summary.

thanks!

dormando:
released in 1.6.4

PR #643, opened 2020-04-06
Add build option to disable unix socket

jefyt:
1. **configure.ac** - Add **_--disable-unix-socket_** to disable and define **DISABLE_UNIX_SOCKET**.
2. **memcached.*** - Guard all unix socket-related codes with **DISABLE_UNIX_SOCKET**. Take note of negative checking (**_#ifndef DISABLE_UNIX_SOCKET_** instead of **_#ifdef UNIX_SOCKET_**). This is just to make sure that current code even without a config file is the default or supports unix socket.
3. **t/** - Check first if unix socket is supported before executing unix socket-related tests.

**Tests:**
1. [test_default_build_(unix_socket_supported).log](https://github.com/memcached/memcached/files/4436455/test_default_build_.unix_socket_supported.log) - **_t/unixsocket.t .............. ok_. All tests successful. Files=78, Tests=104868,**
2. [test_disable-unix-socket_build.log](https://github.com/memcached/memcached/files/4436456/test_disable-unix-socket_build.log) - **_t/unixsocket.t .............. skipped: Unix domain socket is not supported._. All tests successful. Files=78, Tests=104425,**

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3371,7 +3371,6 @@ static inline void get_conn_text(const conn *c, const int af,
     addr_text[0] = '\0';
     const char *protoname = "?";
     unsigned short port = 0;
-    size_t pathlen = 0;
 
     switch (af) {
         case AF_INET:
@@ -3396,7 +3395,10 @@ static inline void get_conn_text(const conn *c, const int af,
             protoname = IS_UDP(c->transport) ? "udp6" : "tcp6";
             break;
 
+#ifndef DISABLE_UNIX_SOCKET
         case AF_UNIX:
+        {
+            size_t pathlen = 0;
             // this strncpy call originally could piss off an address
             // sanitizer; we supplied the size of the dest buf as a limiter,
             // but optimized versions of strncpy could read past the end of
@@ -3419,7 +3421,9 @@ static inline void get_conn_text(const conn *c, const int af,
                     pathlen);
             addr_text[pathlen] = '\0';
             protoname = "unix";
+        }
             break;
+#endif /* #ifndef DISABLE_UNIX_SOCKET */
     }
 
     if (strlen(addr_text) < 2) {
@@ -7812,6 +7816,7 @@ static int server_sockets(int port, enum network_transport transport,
     }
 }
 
+#ifndef DISABLE_UNIX_SOCKET
 static int new_socket_unix(void) {
     int sfd;
     int flags;
@@ -7889,6 +7894,9 @@ static int server_socket_unix(const char *path, int access_mask) {
 
     return 0;
 }
+#else
+#define server_socket_unix(path, access_mask)   -1
+#endif /* #ifndef DISABLE_UNIX_SOCKET */
 
 /*
  * We keep the current time of day in a global variable that's updated by a
@@ -7967,9 +7975,13 @@ static void usage(void) {
     printf(PACKAGE " " VERSION "\n");
     printf("-p, --port=<num>          TCP port to listen on (default: %d)\n"
            "-U, --udp-port=<num>      UDP port to listen on (default: %d, off)\n"
+#ifndef DISABLE_UNIX_SOCKET
            "-s, --unix-socket=<file>  UNIX socket to listen on (disables network support)\n"
+#endif /* #ifndef DISABLE_UNIX_SOCKET */
            "-A, --enable-shutdown     enable ascii \"shutdown\" command\n"
+#ifndef DISABLE_UNIX_SOCKET
            "-a, --unix-mask=<mask>    access mask for UNIX socket, in octal (default: %o)\n"
+#endif /* #ifndef DISABLE_UNIX_SOCKET */
            "-l, --listen=<addr>       interface to listen on (default: INADDR_ANY)\n"
 #ifdef TLS
            "                          if TLS/SSL is enabled, 'notls' prefix can be used to\n"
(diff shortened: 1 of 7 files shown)


memcached.c
```diff
@@ -255,7 +255,9 @@ static void stats_reset(void) {
 
 static void settings_init(void) {
     settings.use_cas = true;
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 258: In hopes of getting more of these windows things upstreamed, lets find out how to minimize the ifdef soup that's going to happen.

Here, I'd suggest leaving the struct's unchanged

memcached.c
```diff
@@ -279,7 +281,9 @@ static void settings_init(void) {
     settings.oldest_live = 0;
     settings.oldest_cas = 0;          /* supplements accuracy of oldest_live */
     settings.evict_to_free = 1;       /* push old items out of cache when memory runs out */
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 284: same here, just remove the ifdef.

memcached.c
```diff
@@ -631,6 +635,7 @@ conn *conn_new(const int sfd, enum conn_states init_state,
     c->transport = transport;
     c->protocol = settings.binding_protocol;
 
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 638: with other changes below socketpath will always be 0, so this ifndef/else/endif should be removable as well.

memcached.c
```diff
@@ -3252,9 +3260,11 @@ static void process_stat_settings(ADD_STAT add_stats, void *c) {
     APPEND_STAT("verbosity", "%d", settings.verbose);
     APPEND_STAT("oldest", "%lu", (unsigned long)settings.oldest_live);
     APPEND_STAT("evictions", "%s", settings.evict_to_free ? "on" : "off");
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 3263: same. it'll just always be NULL.

memcached.c
```diff
@@ -3396,6 +3408,7 @@ static inline void get_conn_text(const conn *c, const int af,
             protoname = IS_UDP(c->transport) ? "udp6" : "tcp6";
             break;
 
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 3411: think I'd like to add a scope (`{ }`)  around the code under case AF_UNIX: so pathlen can be moved in there and reduce two ifdef's to one.

memcached.c
```diff
@@ -7812,6 +7826,7 @@ static int server_sockets(int port, enum network_transport transport,
     }
 }
 
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 7829: not much to be done here :) Maybe when I'm splitting out the code next week I can consolidate some of this into its own file.

memcached.c
```diff
@@ -7967,9 +7983,13 @@ static void usage(void) {
     printf(PACKAGE " " VERSION "\n");
     printf("-p, --port=<num>          TCP port to listen on (default: %d)\n"
            "-U, --udp-port=<num>      UDP port to listen on (default: %d, off)\n"
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 7986: this'll have to stay I guess.

memcached.c
```diff
@@ -8919,11 +8943,15 @@ int main (int argc, char **argv) {
     setbuf(stderr, NULL);
 
     char *shortopts =
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 8946: ... but leave the option parsers in there. remove all the ifdefs from getopt stuff.

memcached.h
```diff
@@ -397,9 +397,11 @@ struct settings {
     rel_time_t oldest_live; /* ignore existing items older than this */
     uint64_t oldest_cas; /* ignore existing items with CAS values lower than this */
     int evict_to_free;
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 400: don't need this.

memcached.c
```diff
@@ -10187,6 +10223,7 @@ int main (int argc, char **argv) {
 #endif
     clock_handler(0, 0, 0);
 
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 10226: Think if an `#else` is added above to define an empty server_socket_unix() if disabled, this could be removed, since socketpath will always be null.

memcached.c
```diff
@@ -9015,10 +9047,12 @@ int main (int argc, char **argv) {
             exit(EX_USAGE);
 #endif
             break;
+#ifndef DISABLE_UNIX_SOCKET
```
dormando, line 9050: compilers may complain about switch coverage... could move the ifdef's inside the parser and print a text error about it being disabled.

jefyt:
@dormando Thanks for the review! I believe I understood your preferred way of disabling unix socket. In my changes, I completely disabled in preprocessing and this includes everything. Your preferred way is to reduce #ifdefs and just use socketpath. I'll revise based on your comments.

jefyt:
@dormando I've made changes based on your comments. For your reference, I've tested both with and without unix socket build. Please see attached. Thanks!

[test_disable-unix-socket_build.log](https://github.com/memcached/memcached/files/4464647/test_disable-unix-socket_build.log)
[test_default_build.log](https://github.com/memcached/memcached/files/4464648/test_default_build.log)

dormando:
better, looks much shorter, thanks!

dormando:
there's definite risk of bitrot with this until I get extra buildbots to run with/without so that sort of sucks, but I'll try it for now.

next tip: make pull requests against the `next` branch.

dormando:
released in 1.6.4

PR #647, opened 2020-04-11
Fix build warnings in Windows.

jefyt:
1. **_warning: passing argument 4 of 'event_set' from incompatible pointer type [-Wincompatible-pointer-types]_**
   --> libevent expects callbacks to use **evutil_socket_t** for fd instead of int. Current code works fine in *nix because it's just **int** but not in Windows because it's defined as **intptr_t**.
2. **_warning: passing argument 4 of 'getsockopt' from incompatible pointer type [-Wincompatible-pointer-types]_**
   --> Apply casting for Windows' **getsockopt** call
3. **_warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]_**
   --> Replace **%lx** with more portable **%p**. **long** is just 32-bit in 64-bit Windows.

[fix_windows_warnings_test.log](https://github.com/memcached/memcached/files/4464890/fix_windows_warnings_test.log) is the test log of this change.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -103,7 +103,7 @@ static bool get_stats(const char *stat_type, int nkey, ADD_STAT add_stats, void 
 static void settings_init(void);
 
 /* event handling, network IO */
-static void event_handler(const int fd, const short which, void *arg);
+static void event_handler(const evutil_socket_t fd, const short which, void *arg);
 static void conn_close(conn *c);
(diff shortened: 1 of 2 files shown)


dormando:
No complaints, I think.

dormando:
released in 1.6.4

had to adjust a bit for a clang warning in the %p bits

PR #648, opened 2020-04-11
Add BoringSSL support.

jefyt:
This won't really matter in *nix since OpenSSL/LibreSSL is shipped by default but in Windows, statically linking TLS library is preferred just like cURL and Chromium. Statically linking with Google's [BoringSSL](https://boringssl.googlesource.com/boringssl) generates at least 30% smaller executable than with OpenSSL.
[BoringSSL](https://boringssl.googlesource.com/boringssl) is an OpenSSL derivative and is mostly source-compatible, for the subset of OpenSSL retained. [SSL_in_before](https://www.openssl.org/docs/manmaster/man3/SSL_in_before.html) API used in memcached is not available in [BoringSSL](https://boringssl.googlesource.com/boringssl) but this can be disabled. memcached is using this API for rejecting peer renegotiations which is already the default in [BoringSSL](https://boringssl.googlesource.com/boringssl).

[boringssl_support_test.log](https://github.com/memcached/memcached/files/4465054/boringssl_support_test.log) contains **_make_**, **_make test_**, & **_SSL_TEST=1 make test_** logs.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -701,7 +701,9 @@ conn *conn_new(const int sfd, enum conn_states init_state,
         c->sendmsg = ssl_sendmsg;
         c->write = ssl_write;
         c->ssl_enabled = true;
+#ifndef OPENSSL_IS_BORINGSSL
         SSL_set_info_callback(c->ssl, ssl_callback);
+#endif /* #ifndef OPENSSL_IS_BORINGSSL */
     } else
 #else
     // This must be NULL if TLS is not enabled.
--- a/tls.c
+++ b/tls.c
@@ -181,6 +181,10 @@ int ssl_init(void) {
     return 0;
 }
 
+/* BoringSSL rejects peer renegotiations by default.
+ * See https://boringssl.googlesource.com/boringssl/+/HEAD/PORTING.md
+ */
+#ifndef OPENSSL_IS_BORINGSSL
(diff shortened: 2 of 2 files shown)


jefyt:
Arm64 build error check seems to be CI issue? Log is just empty/null.

dormando:
hmm... debating.

reasoning: until I get add more buildbots, each type of dep will need to be explicitly tested. So now if we claim to support boringssl we'll have to actively build against it since there are minor incompatibilities.

Been having a lot of trouble lately with broken platforms. Either A) We can sit on this until I can set a builder up or B) just say we don't explicitly support it for now. The static binaries with openssl aren't _that_ large in the grand scheme of things.

dormando:
I also re-kicked the arm build for good measure. One reason why I still want to expand my buildbots (which I have but they aren't publicly exposed anymore) is that the travis builds take approximately 13 years to run and really suck to troubleshoot.

jefyt:
@dormando  Thanks for the review!

> B) just say we don't explicitly support it for now. The static binaries with openssl aren't _that_ large in the grand scheme of things.

This is okay. Supporting this will just add load to builder. You can also just cancel this PR.

dormando:
ok, thanks for taking a look.

PR #653, opened 2020-04-18
Use GitHub CI for x86_64 build

jefyt:
GitHub now offers free CI service via Actions. It's faster and has better integration with GitHub. As of this time, ubuntu-latest (18.04 LTS) and macos-latest (Catalina 10.15) are both supported. See https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners for details. Free users can run up to 20 concurrent jobs for Ubuntu and 5 for Mac. This is already more than enough. See https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits for details.

Below are the jobs for this **ci.yml**:
1. **ubuntu-build-test**:
   OS: **_ubuntu-latest_** (latest LTS)
   Steps:
    **_GCC_** build
    **_make test_**
    **_make test_tls_**
2. **macos-build-test**
   OS: _**macos-latest**_
   Steps:
    **_Clang_** build
    **_make test || true_** (intentional || true. Currently test is failing on Mac)
    **_make test_tls || true_** (intentional || true. Currently test is failing on Mac)
3. **debian-build** (Build check against latest GCC and deps)
   OS: **_debian:testing_** container
   Steps:
    **_GCC_** build
    **No Tests. Build Only.**

Advantages:
1. Faster and better integration with GitHub.
2. Devs can immediately check for build errors on a clean CI environment after push without waiting for PR. Results can be checked on their GitHub Actions page or the commit status.

Disdvantages (as of this time):
1. No support for arm64 that's why Travis is retained.
2. Cannot manually start build without action (e.g. push, PR).

See https://github.com/jefyt/memcached-windows/actions/runs/81668653 for the runner result of this PR's push. Please take note that test (e.g. **t/expirations.t**) is failing on Mac and OK status is intentionally set. This needs separate defect investigation and not CI issue itself.


dormando:
Thanks! Having mac is neat.. but I really don't want _three_ builder systems :/ Most of my bugs have been an issue of the builders not being wide enough.

jefyt:
I can close this PR or continue with just the macOS build. What do you think?

dormando:
why no tests for the debian-testing build?

I'm getting pretty far along with my buildbot redo (https://build.memcached.org/) - it's structured to run a "fast" test (~5m at most) on a host (ubuntu 20.04) with compile/test parallelism, and if that passes it runs a full gamut of platforms... which I will have to build out over time. This will include an ARM, but I don't have an OS X one unless someone sets one up or donates.

The problem is I can't securely (not yet, anyway) throw tests at random PR's immediately. I'll build out a tool for me to label/mark/something on PR's that'll run the buildbots after a quick code review and then report back via comment/status, but it won't be immediate feedback.

So, thinking of trying this out; disabling travis and using just the latest + OS X builds from this one. Travis is currently 30+ minute delay and the ARM platform is broken on and off. If this is faster people can still get quickish feedback.

I strongly dislike the idea that people don't run tests before pushing code (and therefore full tests after a quick review would be fine), but what can I do :P

jefyt:
> why no tests for the debian-testing build?

I just thought you prefer only Ubuntu build-test based on the current Travis setup. Adding tests for **debian-testing** is not a problem although based on #659, the TLS tests will surely fail.

> I strongly dislike the idea that people don't run tests before pushing code (and therefore full tests after a quick review would be fine), but what can I do :P

If I understood your comments correctly, you prefer this:
- Completely disable Travis both for ARM and x86_64
- Enable Ubuntu build+tests
- Enable Debian build+tests (currently failing for TLS)
- Enable macOS build+tests (currently failing)

Please confirm. Thanks!

dormando:
- Completely disable Travis both for ARM and x86_64
yes
- Enable Ubuntu build+tests
yes
- Enable Debian build+tests (currently failing for TLS)
no
- Enable macOS build+tests (currently failing)
yes

bonus points for fixing the mac issue :)

jefyt:
Please confirm again the latest branch. Travis CI is now removed. See https://github.com/jefyt/memcached-windows/actions/runs/92925715 for the result.

Please take note of the following test failures on macOS:

1. testapp non-TLS: test stopped after **ok 54 - binary_illegal**, most likely caused by **test_binary_pipeline_hickup**?
2. testapp TLS: **OK**
3. **t/idle-timeout.t** and **t/ssl_cert_refresh.t**: **NG**

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        export CC=clang
+        export ac_cv_libssl_dir=/usr/local/opt/openssl@1.1
+        ./configure --enable-tls --enable-extstore
+        make -j
+    - name: Test
+      run: make test || true
+    - name: Test TLS
+      run: make test_tls || true
+
+  debian-build:
```
dormando, line 49: you can drop this one, at least for now.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        export LD_LIBRARY_PATH="${OPENSSL_PREFIX}/lib"
+        cpan Net::SSLeay
+        cpan IO::Socket::SSL
+    - name: Build
+      run: |
+        clang --version
+        ./autogen.sh
+        export CC=clang
+        export ac_cv_libssl_dir=/usr/local/opt/openssl@1.1
+        ./configure --enable-tls --enable-extstore
```
dormando, line 42: --enable-extstore is the default.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        sudo apt-get install -y libevent-dev libseccomp-dev git libsasl2-dev
+    - name: Build
+      run: |
+        gcc --version
+        ./autogen.sh
+        ./configure --enable-seccomp --enable-tls --enable-extstore --enable-sasl --enable-sasl-pwdb
+        make -j
+    - name: Test
+      run: make test
+    - name: Test TLS
```
dormando, line 22: lets skip an explicit "make tls" since it will take a really long time to do both test suites. with TLS compiled "make test" will run some basic TLS tests. need to address another bug since they won't cause a failure though.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  ubuntu-build-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install deps
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y libevent-dev libseccomp-dev git libsasl2-dev
```
dormando, line 13: doesn't need libio-socket-ssl-perl or libssl-dev?

also ubuntu-latest is always the LTS?

dormando:
OS X is always clang? There're a few people who've been reporting test issues on OS X (@devnexen ?) but I haven't heard anything about those tests failing. Makes me a bit concerned about the platform.

devnexen:
You can usee gcc from homebrew too. Finally I still get the issue I mentioned the other day but it is not consistent seemingly, I also get abort raised but not the same test as this travis build.
`f 5
frame #5: 0x0000000100001a9e testapp`cache_redzone_test at testapp.c:228:5
   225 	    char *p = cache_alloc(cache);
   226 	    char old = *(p - 1);
   227 	    *(p - 1) = 0;
-> 228 	    cache_free(cache, p);
   229 	    assert(cache_error == -1);
   230 	    *(p - 1) = old;
   231 	
`

`ok 25 - binary_set
Failed to read: Connection reset by peer
Process 26772 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
...
f 3
frame #3: 0x0000000100006cc9 testapp`safe_recv(buf=0x00007ffeefbfecb8, len=24) at testapp.c:955:17
   952 	        if (nr == -1) {
   953 	            if (errno != EINTR) {
   954 	                fprintf(stderr, "Failed to read: %s\n", strerror(errno));
-> 955 	                abort();
   956 	            }
   957 	        } else {
   958 	            if (nr == 0 && allow_closed_read) {
(lldb) f 4
frame #4: 0x0000000100005d6b testapp`safe_recv_packet(buf=0x00007ffeefbfecb8, size=1024) at testapp.c:972:10
   969 	static bool safe_recv_packet(void *buf, size_t size) {
   970 	    protocol_binary_response_no_extras *response = buf;
   971 	    assert(size > sizeof(*response));
-> 972 	    if (!safe_recv(response, sizeof(*response))) {
   973 	        return false;
   974 	    }
   975 	    response->message.header.response.keylen = ntohs(response->message.header.response.keylen);
`

dormando:
@devnexen do you get any of these other failures jefyt mentions though?

devnexen:
nope I do not get these.

.github/workflows/ci.yml
jefyt: > doesn't need libio-socket-ssl-perl or libssl-dev?

GItHub environent already installed this by default.

> also ubuntu-latest is always the LTS?

Supposedly yes. As of this time, it's still 18.04 even though 20.04 already released. Don't know when GitHub will use the 20.04 version.

.github/workflows/ci.yml
jefyt: > OS X is always clang?

I intentionally use clang since gcc is already tested on Ubuntu so it's like testing both on macOS and against clang compiler. What do you think? Do we need to change to gcc?

dormando: I'd rather keep the delta to a minimum, but if that's the default compiler (I can't remember anymore) that'll have to be I guess? I'll need to get another buildbot with clang up since I don't want this thing I don't have access to be "the bug is either clang or OS X"

.github/workflows/ci.yml
dormando: Usually the minimum time is after the 20.04.1 version releases a month after the main release.

dormando:
not sure I'm thinking straight but maybe kill the OS X bits (comment them out?). Seems like it has a delta with real users and I want this github bit to be an immediate sanity check for pull requests more than anything else right now.

On that note, you might try setting PARALLEL=5 or so for the make test suite. If there isn't much RAM available 2-4 might work, as that cuts the test suite time down nearly linearly.

.github/workflows/ci.yml
jefyt: I'll just remove explicitly setting clang. I think default is gcc.

jefyt:
> On that note, you might try setting PARALLEL=5 or so for the make test suite. If there isn't much RAM available 2-4 might work, as that cuts the test suite time down nearly linearly.

Okay, I'll try PARALLEL=5 first then will just lower if needed.

jefyt:
https://github.com/jefyt/memcached-windows/actions/runs/93396403 contains latest build.

- _**gcc --version**_ outputs **_Apple clang version 11.0.3 (clang-1103.0.32.59)_**.
- macOS test is inconsistent compared with the last run, now it stopped after **_ok 42 - binary_incr_**

Please confirm again the modifications.

dormando:
did you try OS X without TLS at all? on new enough systems the TLS bits are causing it to hang forever in testapp sometimes.

just one last sanity check for that... but I think we should drop OS X and just have the one quick test.

jefyt:
> did you try OS X without TLS at all? on new enough systems the TLS bits are causing it to hang forever in testapp sometimes.

1. memcached no TLS support: **NG**. 
2. memcached with TLS support: **NG**. It test stops before executing with TLS.


> just one last sanity check for that... but I think we should drop OS X and just have the one quick test.

Should we drop OS X build and/or test check?

dormando:
Yeah drop the whole OS X tester. I don't want to debug that as it doesn't seem to match what users have. Can look at adding it back, perhaps if we get a user who's invested enough in ensuring it works :) I don't have the spare time.

jefyt:
Please confirm again the modifications. See https://github.com/jefyt/memcached-windows/actions/runs/94352395 for the latest log. It took just **< 4 mins** to complete the job.

martin-g:
With `PARALLEL=5` this test failed here on my ARM64 VM:

```
===(   81975;131  18088/?  24/?   28/130  4/?  1/? )====================
#   Failed test 'client cert is verified'
#   at t/ssl_verify_modes.t line 20.
#          got: undef
#     expected: '1'
t/ssl_verify_modes.t ........ Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/1 subtests 
```

martin-g:
Same error with `PARALLEL=4`. 
It works fine with `PARALLEL=3`!

jefyt:
> Same error with `PARALLEL=4`.
> It works fine with `PARALLEL=3`!

@martin-g I also confirmed this error on GitHub CI environment. See https://github.com/jefyt/memcached-windows/runs/643275527?check_suite_focus=true. **_PARALLEL=5 make test_tls_** failed. **t/ssl_session_resumption.t (Wstat: 256 Tests: 8 Failed: 1)**.

As requested, this PR's latest branch removed **_make test_tls_** that's why no error.

dormando:
Excellent, thanks! Sorry I basically cut this down to nothing. Getting a result in 4 minutes instead of 30+ will be a big help.

dormando:
Out in 1.6.6

PR #655, opened 2020-04-19
crc32c.c: don't attempt to enable hardware crc32 on ia64

trofi:
Itanium has no hardware crc32 implementation and build fails as:

```
ia64-unknown-linux-gnu-gcc -DHAVE_CONFIG_H -I.  -DNDEBUG   -g -O2 -pthread \
    ... -c -o memcached-crc32c.o `test -f 'crc32c.c' || echo './'`crc32c.c
crc32c.c: In function 'crc32c_init':
crc32c.c:385:9: error: unknown register name '%edx' in 'asm'
  385 |         __asm__("cpuid" \
      |         ^~~~~~~
```

The change removes ia64 from crc32 paths.

Bug: https://bugs.gentoo.org/718136

### DIFF
--- a/crc32c.c
+++ b/crc32c.c
@@ -167,7 +167,7 @@ static uint32_t crc32c_hw_aarch64(uint32_t crc, const void* buf, size_t len)
 #endif
 
 /* Apply if the platform is intel */
-#if defined(__X86_64__)||defined(__x86_64__)||defined(__ia64__)
+#if defined(__X86_64__)||defined(__x86_64__)
 
 /* Multiply a matrix times a vector over the Galois field of two elements,
    GF(2).  Each element is a bit in an unsigned integer.  mat must have at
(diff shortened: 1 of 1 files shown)


dormando:
Thanks!

Is this always true for ia64? Think this code came from somewhere else.

trofi:
If `__ia64__` is Itanium (true for `gcc`) then yes, it never worked: ia64's register names are very different and ISA is usually very RISC-style (no CRC32 primitives).

Maybe some compiler defines `__ia64__` to mean `__x86_64__`? I hope not :)

dormando:
I found how that got added the other day. It's getting fixed along with a big-endian runtime issue separately. I'll poke here when it's released.

dormando:
https://github.com/memcached/memcached/issues/660 notes/code here, fwiw.

dormando:
Out in 1.6.6

PR #656, opened 2020-04-19
Changed the parts using strtol to use safe_strtol

minkikim89:


dormando:
Did you check for test coverage? Anything that could trigger a proper error?

dormando:
(also, thanks! sorry it took so long to check)

minkikim89:
It can be a problem when using the following command.
image

dormando:
If there's a break/fix in the code, do the tests cover them? can you update them if not?

minkikim89:
I added some test codes. However, I haven't found a proper place to add other tests.

dormando:
What were you having trouble finding a place for?

minkikim89:
I already updated some test codes for "flush_all", "slabs_reassign".
But I don't know where I have to add test codes for "verbosity", "slabs_automove".

dormando:
You can just create a new test file for those two right now. "strtol-testing.t". Looks like automove never got its own tests and verbosity isn't really tested anywhere.

minkikim89:
I added tests. Check it please.

dormando:
Looks good. I'll have to check it more carefully when I get back to it. Thanks!

memcached.c
```diff
@@ -5793,11 +5800,11 @@ static void process_flush_all_command(conn *c, token_t *tokens, const size_t nto
...
 
     if (ntokens != (c->noreply ? 3 : 2)) {
-        exptime = strtol(tokens[1].value, NULL, 10);
-        if(errno == ERANGE) {
-            out_string(c, "CLIENT_ERROR bad command line format");
+        if (!safe_strtol(tokens[1].value, &exptime_int)) {
+            out_string(c, "CLIENT_ERROR invalid exptime argument");
             return;
         }
+        exptime = realtime(EXPTIME_TO_POSITIVE_TIME(exptime_int));
```
dormando, line 5807: Note the comment below: realtime() is run after exptime is checked for a 0 value.

minkikim89: OK. I fixed it.

dormando:
Released in 1.6.7!

Issue #654, opened 2020-04-19
compile error occurs


minkikim89:
OS: CentOS Linux release 7.6.1810 (Core)
Compiler: gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)

Please check the error below
```
memcached.c:106:1: error: unknown type name 'evutil_socket_t'
 static void event_handler(const evutil_socket_t fd, const short which, void *arg);
 ^
memcached.c:186:1: error: unknown type name 'evutil_socket_t'
 static void maxconns_handler(const evutil_socket_t fd, const short which, void *arg) {
 ^
memcached.c:7499:1: error: unknown type name 'evutil_socket_t'
 void event_handler(const evutil_socket_t fd, const short which, void *arg) {
 ^
memcached.c:7931:1: error: unknown type name 'evutil_socket_t'
 static void clock_handler(const evutil_socket_t fd, const short which, void *arg) {
 ^
```

dormando:
libevent version?

minkikim89:
libevent-1.4.13-stable

jefyt:
@dormando Haven't considered older version of libevent when I introduced the **_evutil_socket_t_** change for Windows. I can handle fixing this issue. Fix is minimal only and can be applied to **memcached.h** file since both **thread.c** and **memcached.c** include this header file:
<pre>
#ifndef evutil_socket_t
#define evutil_socket_t int
#endif
</pre>

Just let me know.

jefyt:
I've manually confirmed 555ed28 against [libevent-1.4.13-stable](https://github.com/libevent/libevent/releases/tag/release-1.4.13-stable). Build logs:

[issue_654_nofix.log](https://github.com/memcached/memcached/files/4502970/issue_654_nofix.log)
[issue_654_withfix.log](https://github.com/memcached/memcached/files/4502971/issue_654_withfix.log)

Ubuntu, Debian, and macOS (clang) Build Logs: https://github.com/jefyt/memcached-windows/actions/runs/82759751

dormando:
centos is the wrench that keeps on breaking your shins...

Thanks and sorry; I'm pausing further development until my buildbots are redone for real this time.

dormando:
centos 7 seems to come with libevent 2.0.21.

are you running a ~7.6 version (1810?) without ever upgrading the packages?

I can't quickly tell if the base centos version would flip to libevent 2.0 after a yum upgrade or not. I just installed a builder from the latest C7.

[...]

PR #657, opened 2020-04-20
Fix error: unknown type name 'evutil_socket_t'

jefyt:
memcached.c:106:1: error: unknown type name '**evutil_socket_t**'
 static void event_handler(const evutil_socket_t fd, const short which, void *arg);
 ^

#647 (3d0697e) introduced this build error.

libevent [1120f04f3eaafe98259ef286eecb648337b2214f](https://github.com/libevent/libevent/commit/1120f04f3eaafe98259ef286eecb648337b2214f) (**2007-11-25 16:52:53 -0500**) submit introduced **evutil_socket_t** change. Unfortunately, this is incompatible with older versions of libevent that memcached still supports (e.g. [1.4.13-stable](https://github.com/libevent/libevent/releases/tag/release-1.4.13-stable)).

Fix is just to define if not defined. No need to check for libevent version. **evutil_socket_t** is a definition in libevent.

See #654 for the issue details.

### DIFF
--- a/memcached.h
+++ b/memcached.h
@@ -52,6 +52,11 @@
 #include <openssl/ssl.h>
 #endif
 
+/** libevent < release-2.0.3-alpha does not define/support evutil_socket_t */
+#ifndef evutil_socket_t
+#define evutil_socket_t int
+#endif /* #ifndef evutil_socket_t */
+
 /** Maximum length of a key. */
 #define KEY_MAX_LENGTH 250
 


dormando:
thanks!

dormando:
going to close this in favor of bumping the minimum libevent version to 2.0. I don't want to be on the hook for bugs in 1.4.x :(

PR #658, opened 2020-04-20
Disable bad memory test on Darwin.

devnexen:
Might be kernel issue but the signal isunfortunately not discarded.

### DIFF
--- a/testapp.c
+++ b/testapp.c
@@ -211,7 +211,7 @@ static enum test_return test_issue_161(void)
 
 static enum test_return cache_redzone_test(void)
 {
-#ifndef HAVE_UMEM_H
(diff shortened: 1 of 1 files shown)


dormando:
think I pulled out the UMEM compat from cache.c? so can probably drop that entirely.

devnexen:
even better :-)

dormando:
Ah I did must've parsed this backwards. I removed the UMEM stuff, which means that test is relevant.

Can you fix the test properly? Or at least write up what's going wrong in more detail, what version of OS X, etc. Doesn't seem like this is failing for other OS X builds.

devnexen:
Well ... I can t reproduce anymore. The redzone test which raises an abort does but the the custom signal handler was not triggered. seems with last mac update it s gone (do not know if this is really related tough).

PR #665, opened 2020-05-04
ECONNRESET fix proposal for Darwin.

devnexen:
forc+exec is somewhat a bit buggy under Darwin, prefering posix_spawn
here instead.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -7592,6 +7592,7 @@ static int server_socket(const char *interface,
     struct addrinfo *ai;
     struct addrinfo *next;
     struct addrinfo hints = { .ai_flags = AI_PASSIVE,
+                              .ai_protocol = 0,
                               .ai_family = AF_UNSPEC };
     char port_buf[NI_MAXSERV];
     int error;
--- a/testapp.c
+++ b/testapp.c
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <unistd.h>
 #include <netinet/in.h>
+#include <spawn.h>
 #include <fcntl.h>
 
 #include "config.h"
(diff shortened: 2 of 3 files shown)


dormando:
Can you expand on what's happening and why this change fixes it? Makes me suspicious to see that fork/exec doesn't work.

devnexen:
I noted I got ECONNRESET errors with connections for read operations at least leading to abort. fork under Mac is known to be not fully working as expected/being buggy unless putting quite some workaround while posix_spawn arrange the situation in this area.

dormando:
Almost all of the testapp.c tests are redundant, I'm pretty sure: https://github.com/memcached/memcached/issues/674 - might want to do this in favor of further hacking it up. Would be nice if testapp pivoted to actual unit tests and we kept the integration/end to end tests all in perl.

dormando:
I was fiddling with a Darwin machine over the weekend and didn't get this to happen; is there anything I need to do to reproduce?

devnexen:
Do you get at least the abort cases ?

dormando:
paste an example? What're your `./configure` arguments? Test suite seems to pass without any errors for me.

dormando:
This is on catalina and I haven't tried all of the configure args yet.

devnexen:
ah right because you re testing next branch maybe ? indeed it seems works better.

dormando:
Ah yeah; I'm on next. so I guess that other fix fixes... this one too? Or I can't repro anymore.

devnexen:
likely yes :)

dormando:
Closing this out. We can re-open and re-try if we have a .. re-production again :)

Thanks for the submission, as always!

Issue #668, opened 2020-05-05
run test suite (or part of it) under Valgrind memcheck


dormando:
I want to add a builder under the new buildbut setup (#664) which generates a Valgrind report, ideally failing if it detects something evil. It's fine if it just fails now; the failures will be exposed via buildbot and we can chip at it. There were more than one real bug recently which valgrind would've caught handily.

Spent some time trying to just run the test suite under valgrind but it explodes in glorious ways, not even reliably finishing 00-startup.t. I also haven't given any thought on how to collect the logs other than printing them and hoping buildbot gets it right.

Throwing up this issue so A) I don't forget, and B) anyone who comes along is welcome to figure this out and put up a PR or comment. Would be a big help :)

At very least, we need to force O0 or O1 instead of the default O2 as my bizarre coding style seems to trip false alarms.

jefyt:
> At very least, we need to force compile with O0 or O1 instead of the default O2 as my bizarre coding style seems to trip false alarms.

Based on my checking **_memcached-debug_** (default of test suite) and the optimized **_memcached_ (O2)** have same valgrind test results. But, the **_prove_** test results differ. **_memcached_ (O2)** has failing test cases not in **_memcached-debug_** (**t/idle-timeout.t**). Did not check if consistent.

> B) anyone who comes along is welcome to figure this out and put up a PR or comment. Would be a big help :)

Here's my suggestion (with test results):

1. Run all test cases under **_valgrind_** (no _**--track-origins**_ first for faster execution) and save all valgrind logs to a directory (current dir by default).
   **Log Files:**
- notls_TESTFILE.PID.log (e.g. **_notls_binary-extstore.t.12060.log_**).
- tls_TESTFILE.PID.log (e.g. **_tls_binary-extstore.t.12061.log_**).
2. If any of the test cases has **_valgrind error_** (**_not prove_** test errors), test again under **_valgrind_** but with _**--track-origins**_ option to track the origin of error.

Example result **without** _**--track-origins**_:
<pre>
Thread 3:
Conditional jump or move depends on uninitialised value(s)
   at 0x149B9D: extstore_io_thread (extstore.c:789)
   by 0x57836DA: start_thread (pthread_create.c:463)
   by 0x5ABC88E: clone (clone.S:95)
</pre>

Example result **with** _**--track-origins**_:
<pre>
Thread 3:
Conditional jump or move depends on uninitialised value(s)
   at 0x149B9D: extstore_io_thread (extstore.c:789)
[...]

Issue #666, opened 2020-05-04
t/ssl_verify_modes.t fails when tests are run in parallel 


martin-g:
As suggested at https://github.com/memcached/memcached/pull/653#issuecomment-622555648 I've tried `PARALLEL=...` on the VMs I use for testing Memcached.

```
===(   81975;131  18088/?  24/?   28/130  4/?  1/? )====================
#   Failed test 'client cert is verified'
#   at t/ssl_verify_modes.t line 20.
#          got: undef
#     expected: '1'
# Looks like you failed 1 test of 1.
t/ssl_verify_modes.t ........ Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/1 subtests 
```

On the ARM4 VM it fails for parallelism>4.
On the x86_64 VM it fails for parallelism>3

dormando:
fwiw I only use parallel for non-tls... which should be obvious because it doesn't work otherwise :) if you set parallel for normal tests, it runs a few basic TLS tests sequentially then does the rest in parallel.

martin-g:
OK. I see!
Closing ...

jefyt:
#670 PR fixed this issue.

martin-g:
Thanks for letting me know!
I just confirmed that all tests pass with PARALLEL=5 now!
Awesome!

dormando:
I thought the parallel breakage was at least partly to do with `t/ssl_cert_refresh.t`, which actually overwrites/shuffles the test certs around. So if another test runs at the same time it might fail to validate.

That test should probably be fixed to copy into temp files and then swap those around first.

PR #670, opened 2020-05-06
Improve sig_handler function

### LINKED ISSUE #669: Race condition in ssl_read function on signal handling
**Describe the bug**
Hi,
there is a race condition in ssl_read function from tls.c.
Unfortunately when tls support has been introduced, new race condition on signal handling has been introduced with it.
I found this after i managed to set up your testsuite on RHEL, because there is a system called ABRT which tracks crashes.
In your testsuite there are tests which check signal handling. We noticed that sometimes abrt catches sigsegv crash of memcached-debug executable.
I've investigated this and found out that openssl library registers atexit function which destroys rand_meth_lock before exit function is called. atexit functions are called before threads are killed. This causes that when signal handler calls exit and thread just called SSL_read from tls.c:35, rand_meth_lock is destroyed and after that memcacheds thread will sometimes manage to lock it before it is killed which obviously results in segfault.

The offending part of code is in openssl crypto/rand/rand_lib.c in RAND_get_rand_method function (CRYPTO_THREAD_write_lock) call.

[backtrace.txt](https://github.com/memcached/memcached/files/4579313/backtrace.txt)

**To Reproduce**
Run memcacheds signal handling tests with enabled tls support on machine which has cpu with more than 4 threads (the more the better). The problem does not occur every time so you will have to maybe run it multiple times. If you have enabled core dumping then you should get some core dumps (on redhat systems coredumpctl can verify).

**System Information**
 - OS/Distro: RHEL
 - Version of OS/distro: 8
 - Version of memcached: 1.5.16
 - Version of openssl: 1.1.1c
 - Hardware detail: i used CPU with 16 cores 32 threads

TomasKorbar:
Wait until all currently running callbacks are finished and then
break the base loop.
Do not exit from sig_handler.
This approach has the advantage that we should be out of all
critical sections and memcached elegantly exits with return
from main.
Another advantage is that it does not matter which thread will
be assigned with executing sig handler because exit will be
always done from the root thread and thus we can avoid further
desynchronizations.

Resolves #669

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -181,7 +181,7 @@ static enum transmit_result transmit(conn *c);
  * can block the listener via a condition.
  */
 static volatile bool allow_new_conns = true;
-static bool stop_main_loop = false;
+static int stop_main_loop = NOT_STOP;
 static struct event maxconnsevent;
 static void maxconns_handler(const evutil_socket_t fd, const short which, void *arg) {
     struct timeval t = {.tv_sec = 0, .tv_usec = 10000};
@@ -8279,8 +8279,8 @@ static void remove_pidfile(const char *pid_file) {
 }
 
 static void sig_handler(const int sig) {
+    stop_main_loop = EXIT_NORMALLY;
     printf("Signal handled: %s.\n", strsignal(sig));
-    exit(EXIT_SUCCESS);
 }
 
 static void sighup_handler(const int sig) {
@@ -8289,7 +8289,7 @@ static void sighup_handler(const int sig) {
 
 static void sig_usrhandler(const int sig) {
     printf("Graceful shutdown signal handled: %s.\n", strsignal(sig));
-    stop_main_loop = true;
+    stop_main_loop = GRACE_STOP;
 }
 
 #ifndef HAVE_SIGIGNORE
@@ -10306,9 +10306,20 @@ int main (int argc, char **argv) {
         }
     }
 
-    fprintf(stderr, "Gracefully stopping\n");
+    switch (stop_main_loop) {
+        case GRACE_STOP:
+            fprintf(stderr, "Gracefully stopping\n");
+        break;
+        case EXIT_NORMALLY:
+            fprintf(stderr, "Exiting normally\n");
+        break;
+        default:
+            fprintf(stderr, "Exiting on error\n");
+        break;
+    }
+
     stop_threads();
-    if (memory_file != NULL) {
+    if (memory_file != NULL && stop_main_loop == GRACE_STOP) {
         restart_mmap_close();
     }
 
--- a/memcached.h
+++ b/memcached.h
@@ -233,6 +233,12 @@ enum pause_thread_types {
     RESUME_WORKER_THREADS
 };
 
+enum stop_reasons {
+    NOT_STOP,
+    GRACE_STOP,
+    EXIT_NORMALLY
+};
+
 #define IS_TCP(x) (x == tcp_transport)
 #define IS_UDP(x) (x == udp_transport)
 


dormando:
thanks!

What exit code does that end up using? normally it's EXIT_SUCCESS.

So there's the stop_main_loop flag that the graceful system uses. If the event_base_loop() for main_base exits normally it'll try to do a graceful shutdown, which should only be done deliberately.

Maybe stop_main_loop needs to be an enum, with 0 being NOT_STOP, 1 being GRACE_STOP, 2 being EXIT_NORMALLY? Then we can gracefully stop all the threads but not issue the restart_mmap_close() unless the correct signal was handled.

unless I'm misunderstanding how the base loop will exit here.

TomasKorbar:
> thanks!
> 
> What exit code does that end up using? previously it was EXIT_SUCCESS.
> 
In 1.5.16 it was still EXIT_SUCCESS
> So there's the stop_main_loop flag that the graceful system uses. If the event_base_loop() for main_base exits normally it'll try to do a graceful shutdown, which should only be done deliberately.
> 
Ohh. I was doing this change originaly for 1.5.16 and did not notice that there is the new stop_main_loop flag. Although that seems to me like active waiting :/ . Maybe the approach with 
event_base_loopexit is better because there is no polling.
I'll explain how the event_base_loopexit works.
When you call event_base_loopexit with NULL as the time argument than the loop will stop processing new events and let all currently running callbacks finish (which prevents segfaults in ssl library). After that event_base_loop which has been called with 0 as flags returns 0, as loop exited without any error. That allowed memcached 1.5.16 to exit as nothing special happened and return from main EXIT_SUCCESS.
> Maybe stop_main_loop needs to be an enum, with 0 being NOT_STOP, 1 being GRACE_STOP, 2 being EXIT_NORMALLY? Then we can gracefully stop all the threads but not issue the restart_mmap_close() unless the correct signal was handled.
> 
> unless I'm misunderstanding how the base loop will exit here.
If i'm right than if you would choose this way than stop_main_loop would lose its purpose. I can edit this pullrequest accordingly if you think too that this is the better solution:)
Appologies for my overlook.

TomasKorbar:
PR updated.

dormando:
Not quite there: I did that flag on purpose. Memcached doesn't use libevent's internal locks, instead event bases are pinned to particular threads. Calling event_base_loopexit() from the wrong thread can be fatal as it adds an event into the list chain to fire later and exit the loop. Even if you're in the same thread I'm fairly sure it's not safe to call that from a signal handler. It might be more safe than what you had before but still not safe.

The main_base is only used for new connections, not any worker threads. Using EVLOOP_ONCE ensures that all pending events/callbacks/etc are handled before it bounces out of the loop. The simple check in my outer loop is nearly identical to the `while (!done) {` inside event_base_loop. main_base also handles the clock timer so it's guaranteed to run every second. It's not really polling anything.

The "safe" equivalent without using EVLOOP_ONCE would to run a recurring timer and call the loopexit() from within there after checking a variable, which now looks like my flag check but with a lot more code :)

dormando:
Your change also still doesn't deal with the original issue; SIGUSR1 is used to tell memcached to gracefully shutdown for restart mode. With this you'd end up in a graceful stop under the wrong signal.

Sorry it's pretty complicated :) Make sure you develop against the `next` branch, since I can't exactly go back and patch 1.5.16.

TomasKorbar:
Hi @dormando ,
Thanks for your patience with this:)
Now when you have explained the subject more, i see that my solution was wrong. I updated PR and changed base to the next branch.

dormando:
Looks better, thanks!

dormando:
Out in 1.6.6

jefyt:
Confirmed that this PR also fixed the parallel TLS test issue (#666) and valgrind TLS test issue (https://github.com/memcached/memcached/issues/668#issuecomment-625413582). Test results:

1. [Parallel TLS Test](https://github.com/jefyt/memcached-windows/runs/670217949?check_suite_focus=true)
2. [Valgrind TLS Test](https://github.com/jefyt/memcached-windows/runs/670251117?check_suite_focus=true)

PR #671, opened 2020-05-06
Update testing certificates

### LINKED ISSUE #659: Make ssl cerificate keys longer
**Describe the bug**
Hi, i'm the maintainer of memcached package on fedora and RHEL. I would like to use your test suite with enabled ssl testing. That is unfortunately impossible at this time when default system wide crypto policy has seclevel of CipherString 2 and thus rsa keys shorter than 2048 bits are not allowed. This keeps memcacheds testing instance from starting. I want to test memcached in environment which has hardened crypto policy so i would like to not change it because of the test.

**To Reproduce**
1. Set your seclevel of cipher string to 2 in your openssl configuration. On debian in
`/etc/ssl/openssl.cnf`, in Fedora `/etc/crypto-policies/back-ends/opensslcnf.config`
2. try to run the test suite with enabled ssl testing.
(in memcached source directory)
$./configure --enable-tls
$make
$export SSL_TEST=1
$make test

**Expected result**
Test suite successfully finishes.
**Actual result**
Test suite hangs and must be interupted with sigint.

**System Information**
 - OS/Distro: any distro with hardened crypto policy but for example Fedora
 - Version of OS/distro: Rawhide
 - Version of memcached: 1.6.3

**Solution**
Please regenerate the ssl certificates used for testing and use rsa keys at least 2048 bites long.
I would send pull request but i was not able to generate the exact same subject for these certificates.

TomasKorbar:
Keys are now 2048 bits long so systems with hardened crypto policies
will not have problem with them.
Resolves #659


tharanga:
Thanks @TomasKorbar. You may include commands (and the .cnf if used) that you used to create certs in a readme, so anyone can generate them as needed in the future. 
PS : I initially thought of scripting this so tests generate certificates on the fly and we don't have to worry about expiring them, but ran into build issues.

dormando:
thanks! Agreed; including the generation steps would be a nice bonus since we missed that the first time around :)

TomasKorbar:
Ok no problem. I also changed the base of this pr to the next branch as dormando told me that is where development happens. @dormando should i create new readme for these certificates or do you want me to put the notes into README.md in root folder?

dormando:
thanks! uhhhh lets start a README.md (or README-TLS.md) in the t/ directory and put them in there.

TomasKorbar:
PR updated. I verified that certificates created by the process described in readme work in crypto hardened environment.

dormando:
Out in 1.6.6

PR #675, opened 2020-05-13
adding openssl errors to SSL certificate loading error messages

### LINKED ISSUE #586: improve test certs for TLS tests
See: https://github.com/docker-library/memcached/pull/54 - looks like newer OpenSSL's don't like the test certs.

I'm sure it's fixable in some way. @tharanga - want to pick this up?

related things to fix:
- tls.c: the stderr fprintf's before `exit(EX_USAGE)` calls shouldn't be gated on verbosity.
- in load_server_certificate, any way to add the SSL_error string into error_msg ?

tharanga:
resolves #586

### DIFF
--- a/tls.c
+++ b/tls.c
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <sysexits.h>
 #include <sys/param.h>
+#include <openssl/err.h>
 
 #ifndef MAXPATHLEN
 #define MAXPATHLEN 4096
@@ -13,7 +14,8 @@
 
 static pthread_mutex_t ssl_ctx_lock = PTHREAD_MUTEX_INITIALIZER;
 
-const unsigned MAX_ERROR_MSG_SIZE = 128;
+const unsigned ERROR_MSG_SIZE = 64;
+const size_t SSL_ERROR_MSG_SIZE = 256;
 
 void SSL_LOCK() {
     pthread_mutex_lock(&(ssl_ctx_lock));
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! I saw this after I'd kicked off the 1.6.6 release :)

tls.c
```diff
@@ -81,54 +83,85 @@ ssize_t ssl_write(conn *c, void *buf, size_t count) {
...
+
     if (settings.ssl_ctx == NULL) {
         snprintf(error_msg, errmax, "Error TLS not enabled\r\n");
         *errmsg = error_msg;
         return false;
     }
+
+    char *ssl_err_msg = malloc(SSL_ERROR_MSG_SIZE);
+    if (ssl_err_msg == NULL) {
+        *errmsg = NULL;
```
dormando, line 121: leaks *error_msg

tls.c
```diff
@@ -81,54 +83,85 @@ ssize_t ssl_write(conn *c, void *buf, size_t count) {
...
-        snprintf(error_msg, errmax, "Error loading the key: %s\r\n", settings.ssl_key);
-        success = false;
+        print_ssl_error(ssl_err_msg, SSL_ERROR_MSG_SIZE);
+        snprintf(error_msg, errmax, "Error loading the key: %s : %s\r\n",
+            settings.ssl_key, ssl_err_msg);
     } else if (!SSL_CTX_check_private_key(settings.ssl_ctx)) {
-        snprintf(error_msg, errmax, "Error validating the certificate\r\n");
-        success = false;
+        print_ssl_error(ssl_err_msg, SSL_ERROR_MSG_SIZE);
+        snprintf(error_msg, errmax, "Error validating the certificate: %s\r\n",
```
dormando, line 139: Here and above/below, should check the result of snprintf() to see if the call was truncated; since that will snip the `\r\n` and cause a protocol hang if we somehow get a huge error from OpenSSL.

tls.c
```diff
@@ -160,6 +164,8 @@ static bool load_server_certificates(char **errmsg) {
         free(error_msg);
     } else {
         *errmsg = error_msg;
+        error_msg += (err_msg_size >= errmax ? errmax - 1: err_msg_size);
+        snprintf(error_msg, CRLF_NULLCHAR_LEN, "\r\n");
```
dormando, line 168: shouldn't this be errmax - CLRF_NULLCHAR_LEN instead of - 1?

tharanga: sprintf only appends '\0', there's no CRLF.

dormando: unless I'm missreading, you're offsetting the error_msg variable to errmax - 1, but then snprintf'ing 2 chars + null byte? so that'd be two bytes past the allocation if errmax were hit?

tharanga: errmax is CRLF_NULLCHAR_LEN bytes shorter than the size of error_msg.

dormando: After squinting again, I see I misread a - for a +. Thanks!

dormando:
Released in 1.6.7!

PR #680, opened 2020-05-13
fixing the basic tls test so it exits correctly when fails

### LINKED ISSUE #663: basic TLS tests don't exit properly on failure
If you run the basic TLS tests (make test without SSL_TEST=1) and the basic tests fail, the final error code is from the rest of the test-suite running without TLS.

Meaning it breaks under CI/buildbots :)

@tharanga if interested. making sure I note this here at least.

tharanga:
resolves #663

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -114,8 +114,8 @@ test_tls:
 test_basic_tls:
 	@if test $(SSL_TEST)1 != 1; then \
 	  echo "Running basic tests with TLS"; \
-	  $(builddir)/testapp \
-	  prove $(srcdir)/t/binary.t $(srcdir)/t/getset.t $(srcdir)/t/ssl*; \
+	  $(builddir)/testapp && \
+	  prove $(srcdir)/t/binary.t $(srcdir)/t/getset.t $(srcdir)/t/ssl* && \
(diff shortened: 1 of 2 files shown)


testapp.c
```diff
@@ -574,8 +573,16 @@ static pid_t start_server(in_port_t *port_out, bool daemon, int timeout) {
     }
 
     /* Yeah just let us "busy-wait" for the file to be created ;-) */
-    while (access(filename, F_OK) == -1) {
-        usleep(10);
+    useconds_t wait_timeout = 1000000;
```
dormando, line 576: max wait time of 1s, right? I've already seen under valgrind that the unix socket can take a while. Lets just ham it to 10 seconds? 5 seconds?

Makefile.am
```diff
@@ -114,7 +114,7 @@ test_tls:
 test_basic_tls:
 	@if test $(SSL_TEST)1 != 1; then \
 	  echo "Running basic tests with TLS"; \
-	  $(builddir)/testapp; \
+	  $(builddir)/testapp \
```
dormando, line 117: This causes the prove perl tests to not actually run :(

dormando:
Released in 1.6.7!

PR #682, opened 2020-05-14
Restore SAN entries in testing TLS certificates

LINKIWI:
In the new hardened certs introduced by #671, it seems that the SAN names were removed from the certificates. This change regenerates the certs to include client and server SANs to be consistent with those present in 1.6.5 and earlier:

https://github.com/memcached/memcached/blob/1.6.5/t/client_crt.pem#L31

https://github.com/memcached/memcached/blob/1.6.5/t/server_crt.pem#L31

Currently, memcached does not actually depend on SANs for any functionality. This PR is a patch we use internally to fix some (additional out-of-tree) regressed unit tests, and I figure it would be harmless to propose it upstream as well.

This change also fixes some minor typos in the documented openssl commands used to generate the certs.

cc @TomasKorbar


dormando:
awesome, thanks! Sorry I didn't look at the README too closely.

Issue #681, opened 2020-05-14
When noreply is used server closes unexpectedly the connection


pfreixes:
**Describe the bug**
When Im trying to send a set command with the `noreply` flag the server closes the connection unexpectedly.

The command sent has the following data `set key 0 0 5 noreply\r\nvalue\r\n`

The Memcached server started with three levels of verbosity says:
```
<19 new auto-negotiating client connection
19: going from conn_new_cmd to conn_waiting
19: going from conn_waiting to conn_read
19: going from conn_read to conn_parse_cmd
19: Client using the ascii protocol
<19 set key 0 0 5 noreply
19: going from conn_parse_cmd to conn_nread
> NOT FOUND key
>19 NOREPLY STORED
19: going from conn_nread to conn_new_cmd
19: going from conn_new_cmd to conn_mwrite
Failed to write, and not due to blocking: Message too long
19: going from conn_mwrite to conn_closing
<19 connection closed.
19: going from conn_closing to conn_closed
```

While the tcmpdump between the client and the server is the following one

```
22:30:43.969767 IP6 localhost.54547 > localhost.11211: Flags [S], seq 2204362658, win 65535, options [mss 16324,nop,wscale 6,nop,nop,TS val 1262036203 ecr 0,sackOK,eol], length 0
22:30:43.969862 IP6 localhost.11211 > localhost.54547: Flags [S.], seq 3730121260, ack 2204362659, win 65535, options [mss 16324,nop,wscale 6,nop,nop,TS val 1262036203 ecr 1262036203,sackOK,eol], length 0
22:30:43.969875 IP6 localhost.54547 > localhost.11211: Flags [.], ack 1, win 6371, options [nop,nop,TS val 1262036203 ecr 1262036203], length 0
22:30:43.969888 IP6 localhost.11211 > localhost.54547: Flags [.], ack 1, win 6371, options [nop,nop,TS val 1262036203 ecr 1262036203], length 0
22:30:43.970634 IP6 localhost.54547 > localhost.11211: Flags [P.], seq 1:31, ack 1, win 6371, options [nop,nop,TS val 1262036203 ecr 1262036203], length 30
22:30:43.970652 IP6 localhost.11211 > localhost.54547: Flags [.], ack 31, win 6371, options [nop,nop,TS val 1262036203 ecr 1262036203], length 0
[...]

PR #683, opened 2020-05-17
Fix TCP failure under OS X.

dormando:
sendmsg() on darwin doesn't like zero sized messages, apparently. Bails
out with "message too long". Don't think I intended to run the syscall
when there's nothing actually to send, anyway.

Should fix https://github.com/memcached/memcached/issues/681

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -6817,6 +6817,12 @@ static enum transmit_result transmit(conn *c) {
     msg.msg_iov = iovs;
 
     iovused = _transmit_pre(c, iovs, iovused, TRANSMIT_ALL_RESP);
+    if (iovused == 0) {
+        // Avoid the syscall if we're only handling a noreply.
+        // Return the response object.
+        _transmit_post(c, 0);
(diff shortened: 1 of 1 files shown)


jefyt:
GitHub CI macOS test results:

1. No Fix: https://github.com/jefyt/memcached-windows/runs/682422242?check_suite_focus=true
   Result: TCP NG, TLS OK
2. With Fix (1st Run): https://github.com/jefyt/memcached-windows/runs/682491265?check_suite_focus=true
   Result: TCP t/expirations.t NG, TLS OK
3. With Fix (2nd Run): https://github.com/jefyt/memcached-windows/runs/682503471?check_suite_focus=true
   Result: TCP OK, TLS OK

With this fix, it seems macOS tests (TCP and TLS) are now OK.

dormando:
https://github.com/memcached/memcached/issues/684 is probably a better permanent fix for a lot of the remaining test flakiness like with expirations.t.

PR #686, opened 2020-05-20
Do not join lru and slab maintainer threads if they do not exist

### LINKED ISSUE #685: Memcached attempts to join lru_maintainer thread even if it does not exist
**Describe the bug**
Hi,
When we fixed the race condition in #669 , it seems that it created a new issue which showed itself in testsuite. Once again i used ABRT to check for crashes and i found several segfaults. Since memcached does not call exit() on sigint/sigterm anymore, it always calls stop_threads. stop_threads function always calls stop_lru_maintainer_thread which attempts to join lru maintainer thread. The problem is that this thread does not always exist. In several tests is used option `--no-modern` which causes that lru_maintainer thread is not started, ergo `pthread_join(lru_maintainer_tid, NULL)` in items.c stop_lru_maintainer function gets called with 0 as an argument and segfaults.

**To Reproduce**
Run testsuite and check for crashes (coredumpctl for example). In my testing environment i found 8 segfaults.
**Steps to reproduce**
(in memcached source folder)
1. $ ./configure --enable-sasl --enable-tls
2. $ export SSL_TEST=1
3. $ make
4. $ make test
5. $ coredumpctl

**Expected results**
No segfaults.

**System Information**
 - OS/Distro: Fedora
 - Version of OS/distro: Rawhide
 - Version of memcached: 1.6.6
 - Hardware detail: not relevant (reproduced even on single core system)

**Detail**
Including full backtrace of all threads generated by gdb from my debugging session
[backtrace.txt](https://github.com/memcached/memcached/files/4656742/backtrace.txt)

TomasKorbar:
If no_modern option was supplied then these threads did not run
but memcached still attempted to join them, which resulted in a
segfault.

resolve #685

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -226,12 +226,16 @@ void stop_threads(void) {
     stop_item_crawler_thread(CRAWLER_WAIT);
     if (settings.verbose > 0)
         fprintf(stderr, "stopped lru crawler\n");
-    stop_lru_maintainer_thread();
-    if (settings.verbose > 0)
-        fprintf(stderr, "stopped maintainer\n");
-    stop_slab_maintenance_thread();
-    if (settings.verbose > 0)
-        fprintf(stderr, "stopped slab mover\n");
+    if (settings.lru_maintainer_thread) {
+        stop_lru_maintainer_thread();
+        if (settings.verbose > 0)
+            fprintf(stderr, "stopped maintainer\n");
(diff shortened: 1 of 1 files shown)


dormando:
thanks!

dormando:
Released in 1.6.7!

PR #687, opened 2020-05-23
net: carve response buffers from read buffers

dormando:
balancing start arguments for memory limits on read vs response buffers
is hard. with this change there's a custom sub-allocator that cuts the
small response objects out of larger read objects.

it is somewhat complicated and has a few loops. We may be able to
optimize out some of the loops, but it also may not matter.

needs more testing and fleshing out + adjusting start arguments. tests
pass locally.

TODO:
- [x] fiddle with offsets/sizes to ensure few bytes are wasted out of the bundle. I think 14 objects per bundle is reasonable.. with fully optimizing the wbuf maximum it could be 16 though.
- [x] ensure access alignment. test on BE/strict system + valgrind.
- [x] Add some inline justification comments.
- [x] remove debug prints.
- [x] remove/stub out the response obj cache detail/limits?

why is this kind of complicated:

It could be very simple if we simply allocated front to back and then freed response bundles when refcount hit 0... but if random connections are holding onto response objects for a long time, we will end up with holes. If one out of every 14 response objects gets "hung", you will end up with an entire wasted 16KB per hung client.

so instead there is some heavily short-circuited logic around having a linked list of bundles. If there are no free objects, the bundle is unlinked entirely. If there are any free objects, it is initially linked to the head. If there are further free objects added, it is not reshuffled. The head object should always have at least one free item so we avoid reshuffling when "any" object gets freed.

@tharanga


dormando:
good news: does appear to work under my jank big endian test host. seems like everything important ends up aligned, but it's probably fragile.

dormando:
Originally thought about using a bit vector but... updating the proper bit vector position on free required adding the mask to the response object which would be larger than the byte wasted for a free marker.

Instead I've adjusted it (not pushed yet) to have an iterative "next_check" pointer to avoid looping further into the structure on every allocation. Further, thinking of making it hold 1 blob object per worker to avoid extra function calls + the initialization loop for the common case (which will be < 14 response objects in flight at once).

Amazingly, aside from the stats buffer there's almost nothing that allocates at runtime anymore. This addition feels a bit overcomplex/gross but it's a big usability bump so I'm happy with it.

dormando:
Think this is complete now. Should run it under some mctester workloads/prod testing before merging.

dormando:
note to self: since the first thing we _always_ do after resp_allocate is that memset, I can pull the memset in and write the bundle ptr on alloc and make the line more readable.

dormando:
another note to self: try to bring back response_bj_total (response_obj_used?) counter so users can track bottlenecks more clearly.

dormando:
I've "merged" this into `next` to get it out of the way of other work. if I need to do a release and don't have confidence in it, I'll comment out the two hook lines to disable it.

dormando:
closing this, further work in other issues/PR's.

Issue #420, opened 2018-08-20
setrlimit(RLIMIT_NOFILE fails with EPERM when run under valgrind


thatsafunnyname:
Hello and thanks for memcached,

I was running memcached as part of build system tests, the build system tests run under valgrind with --trace-children=yes, memcached is run from within these test processes, valgrind fails with EPERM when setrlimit(RLIMIT_NOFILE is called when the max/hard limit is changed:

With these ulimits for nofile:

    > ulimit -S -n ; ulimit -H -n
    1024
    4096

Without valgrind we set maxconns to 256:

    > /usr/bin/memcached -p 54321 -c 256
    ^CSignal handled: Interrupt.

With valgrind this fails because valgrind returns EPERM for setrlimit with a change to the hard/max nofile limit:

    > valgrind /usr/bin/memcached -p 54321 -c 256
    failed to set rlimit for open files. Try starting as root or requesting smaller maxconns value.

When maxconns is not passed, if the soft/current ulimit is not 1024 setrlimit fails:

    > ulimit -H -n ;ulimit -S -n
    2048
    2048

    > valgrind /usr/bin/memcached -p 54321
    failed to set rlimit for open files. Try starting as root or requesting smaller maxconns value.

This is with valgrind-3.13.0 on RHEL7 hosts and Valgrind-3.11.0 on Ubuntu 16.

I could not come up with a PR that I thought would be OK.  Always calling setrlimit for the soft/current limit, and only calling setrlimit for the hard/max limit when needed was the approach I thought would work best, but I didn't think the complexity was worth it for the benefit?

I plan to use the valgrind options ```--trace-children-skip*=patt1,patt2,..."```, for the tests that are testing client side libraries.

http://valgrind.org/docs/manual/dist.news.old.html
  "The setrlimit system call now simply updates the emulated limits as best
  as possible - the hard limit is not allowed to move at all and just
  returns EPERM if you try and change it."

dormando:
That's frustrating. memcached has operated this way for 15 years. :/

I'll have to come back to this, unless someone else has a great idea. thanks for bringing it up with plenty of detail!

PR #688, opened 2020-06-16
Setrlimit fix

pkarumanchi9:
fixes https://github.com/memcached/memcached/issues/420

In debug builds, if the soft and hard limits are already greater than the enforced limit, skip calling setrlimit. Calling setrlimit is not allowed when calling under valgrind.

We can always set the the hard and soft limits outside the memcached process and not let memcached take this code path for debug builds.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -10052,7 +10052,7 @@ int main (int argc, char **argv) {
         exit(EX_OSERR);
     } else {
 #ifdef MEMCACHED_DEBUG
-        if (rlim.rlim_cur < settings.maxconns && rlim.rlim_max < settings.maxconns) {
+        if (rlim.rlim_cur < settings.maxconns || rlim.rlim_max < settings.maxconns) {
 #endif
         rlim.rlim_cur = settings.maxconns;
         rlim.rlim_max = settings.maxconns;
--- a/memcached.c
+++ b/memcached.c
@@ -10051,12 +10051,18 @@ int main (int argc, char **argv) {
         fprintf(stderr, "failed to getrlimit number of files\n");
         exit(EX_OSERR);
     } else {
+#ifdef MEMCACHED_DEBUG
+        if (rlim.rlim_cur < settings.maxconns && rlim.rlim_max < settings.maxconns) {
+#endif
         rlim.rlim_cur = settings.maxconns;
         rlim.rlim_max = settings.maxconns;
         if (setrlimit(RLIMIT_NOFILE, &rlim) != 0) {
             fprintf(stderr, "failed to set rlimit for open files. Try starting as root or requesting smaller maxconns value.\n");
             exit(EX_OSERR);
         }
+#ifdef MEMCACHED_DEBUG
+        }
+#endif
     }
 
     /* lose root privileges if we have them */


dormando:
thanks!

Does valgrind set some kind of environment variable/hint to the program that it's being val-grind-ded?

pkarumanchi9:
Yes, there is way to detect it by running a macro "RUNNING_ON_VALGRIND", but this would require copying the valgrind/*.h files into memcached repo.
https://www.valgrind.org/docs/manual/manual-core-adv.html#manual-core-adv.clientreq

I think including it would be bit too much. Would like to hear your thoughts on it.

dormando:
No I wouldn't advocate that. This might be fine for now; I'll check when I get a chance.

Fwiw you wouldn't copy in the header files. You'd detect them in autoconf and conditionally include them if the system has them (perhaps if a specific --configure variable were enabled too). We probably should offer this but not sure what of all the features it exposes are useful.

pkarumanchi9:
Ah ok, that makes sense

dormando:
so, I kind of can't remember anymore since it was 9 years ago, but apparently t/maxconns.t "never worked", per the comment:
```
# NOTE: This test never worked. Memcached would ignore maxconns requests lower
# than the current ulimit. Test needs to be updated.
```
.. but I don't see how that was in the code. Maybe it just didn't do anything, heh. That probably means this approach is fine.

dormando:
If you want some bonus points, can a test for maxconns_fast (fixing t/maxconns.t) actually work with this change?

pkarumanchi9:
While trying to add a testcase for max conns, I found the following -
the reserved_fds that we calculate is 5 per thread. And reserved_fds comes to 4(default num of threads)*5 = 20. But I noticed that under the reserved_fd's, we haven't accounted for  (0,1,2) for stdin, stdout, stderr.  I noticed that there are 4 additional fd's used by memcached which are not accounted in reserved_fd's. More details below

Brought up memcached using domain socket (the way most of tests invoke memcached)
> $ ./memcached-debug -c 30 -o relaxed_privileges -s /tmp/memcached.111.0

verified from the gdb that these fd's are indeed used by memcached. 
> $ (gdb) info threads
  Id   Target Id         Frame
* 1    Thread 0x7f05b45b8740 (LWP 28984) "memcached-debug" 0x00007f05b3c5ebb7 in epoll_wait (epfd=3,
    events=0x55ab6e3a45c0, maxevents=32, timeout=1000) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
  2    Thread 0x7f05b3b3c700 (LWP 28985) "memcached-debug" 0x00007f05b3c219d0 in __GI___nanosleep (
    requested_time=requested_time@entry=0x7f05b3b3be70, remaining=remaining@entry=0x0)
    at ../sysdeps/unix/sysv/linux/nanosleep.c:28
  3    Thread 0x7f05b333b700 (LWP 28986) "memcached-debug" 0x00007f05b3c5ebb7 in epoll_wait (epfd=8,
    events=0x55ab6e3ab420, maxevents=32, timeout=-1) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
  4    Thread 0x7f05b2b3a700 (LWP 28987) "memcached-debug" 0x00007f05b3c5ebb7 in epoll_wait (epfd=13,
    events=0x55ab6e3abf20, maxevents=32, timeout=-1) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
  5    Thread 0x7f05aa339700 (LWP 28988) "memcached-debug" 0x00007f05b3c5ebb7 in epoll_wait (epfd=18,
    events=0x55ab6e3aca40, maxevents=32, timeout=-1) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
  6    Thread 0x7f05b2339700 (LWP 28989) "memcached-debug" 0x00007f05b3c5ebb7 in epoll_wait (epfd=23,
    events=0x55ab6e3ad560, maxevents=32, timeout=-1) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
  7    Thread 0x7f05b1b38700 (LWP 28990) "memcached-debug" 0x00007f05b3f3b9f3 in futex_wait_cancelable (
    private=<optimized out>, expected=0, futex_word=0x55ab6dfafdc8 <maintenance_cond+40>)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:88
  8    Thread 0x7f05b1337700 (LWP 28991) "memcached-debug" 0x00007f05b3f3b9f3 in futex_wait_cancelable (
    private=<optimized out>, expected=0, futex_word=0x55ab6dfb798c <lru_crawler_cond+44>)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:88
  9    Thread 0x7f05b0b36700 (LWP 28992) "memcached-debug" 0x00007f05b3c219d0 in __GI___nanosleep (
    requested_time=requested_time@entry=0x7f05b0b35c40, remaining=remaining@entry=0x0)
    at ../sysdeps/unix/sysv/linux/nanosleep.c:28
  10   Thread 0x7f05abfff700 (LWP 28993) "memcached-debug" 0x00007f05b3f3b9f3 in futex_wait_cancelable (
    private=<optimized out>, expected=0, futex_word=0x55ab6dfa0e08 <slab_rebalance_cond+40>)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:88

count of fd's used by memcached - 27.  I've annotated the epoll fd's used below by the corresponding threads for book keeping

> $ ls -l /proc/28984/fd
total 0
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 0 -> /dev/pts/0
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 1 -> /dev/pts/0
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 10 -> 'pipe:[1574168]'
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 11 -> 'pipe:[1574169]'
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 12 -> 'pipe:[1574169]'
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 13 -> 'anon_inode:[eventpoll]'.   --> thread 4
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 14 -> 'pipe:[1574170]'
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 15 -> 'pipe:[1574170]'
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 16 -> 'pipe:[1574171]'
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 17 -> 'pipe:[1574171]'
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 18 -> 'anon_inode:[eventpoll]'.    --> thread 5 
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 19 -> 'pipe:[1574172]'
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 2 -> /dev/pts/0
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 20 -> 'pipe:[1574172]'
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 21 -> 'pipe:[1574173]'
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 22 -> 'pipe:[1574173]'
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 23 -> 'anon_inode:[eventpoll]'     --> thread 6
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 24 -> 'pipe:[1574174]'
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 25 -> 'pipe:[1574174]'
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 26 -> 'socket:[1574175]'
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 3 -> 'anon_inode:[eventpoll]'    --> thread 1
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 4 -> 'pipe:[1574166]'
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 5 -> 'pipe:[1574166]'
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 6 -> 'pipe:[1574167]'
l-wx------ 1 vagrant vagrant 64 Jul  7 21:41 7 -> 'pipe:[1574167]'
lrwx------ 1 vagrant vagrant 64 Jul  7 21:41 8 -> 'anon_inode:[eventpoll]'.    --> thread 3
lr-x------ 1 vagrant vagrant 64 Jul  7 21:41 9 -> 'pipe:[1574168]'


I started writing the tests to account for this. But hit another roadblock here.  Basically the connections are rejected before we increment the counter "rejected_conns" here - 
accept4 syscall errors out with "accept4(): Too many open files" and errors out.  errno we see is EMFILE., essentially the process hitting the open files limit.

https://github.com/memcached/memcached/blob/master/memcached.c#L7120-L7124
https://github.com/memcached/memcached/blob/master/memcached.c#L7142-L7145

Can we decouple the test case in a different PR as we might need to make changes to memcached.c for accounting rejected_conns. Another suggestion is - if we can add the reserved_fds to max_conns instead of subtracting it from max_conns to arrive at the total number of open files allowed for that process.

I've updated the test to keep track of the curr_conns limit instead of rejected_conns due to above mentioned issue. Once memcached.c is updated with the fix to track the rejected_conns, I will add another test and cut a different PR for it.

dormando:
thanks!

This is why writing tests is good :) I figured there were a few more bugs lurking after the last one.

Please open another issue and/or PR to track this. I've already merged the code in this PR.

dormando:
also, fwiw (sorry just re-reading your comment now), the main point of the `maxconns_fast` mode is to be able to open the connection and print an error to the user. So if accept() bails out on too many connections the feature isn't working right, so rejected_conns not incrementing would be part of the bug, not something to move around.

pkarumanchi9:
I see. With the current code structure, the purpose of maxconns_fast is negated as accept()/accept4() fails with  errno - "EMFILE The per-process descriptor table is full."

One suggestion is to increment the rejected_conns if accept* syscall fails with EMFILE and get rid of maxconns_fast mode.

Took a stab at this in the following pull request - https://github.com/memcached/memcached/pull/700. We can move the conversation there.

PR #692, opened 2020-06-30
net: buffer unity followup

pkarumanchi9:
Testing in progress. Addressed the assertions that I ran into using the debug build.


pkarumanchi9:
Allocation path:
        1. If the current thread's bundle is full
                a. Go to the previous bundle in the list
        2. If buffer allocation is not successful in one attempt, allocate a new bundle and point the thread to the new bundle. 

Free path:
        1. If refcount drops to zero
                a. And If the current thread is pointing to it.
                        i. Case A: If this the only one in the list, leave it in place.
                        ii. Case B: If it is part of the list, remove from the list and free up the memory.
                b. If this is a dangling pointer(not pointed by the current thread), free up the memory
        2. Else
                a. If its already in the list and pointed by the current thread, leave it in place
                b. Else add the thread bundle to the head of the list and point the thread to the new head.

dormando:
Looks like you caught the missing free = false. I lost that in a refactor.

I'm starting another change in #693 - just writing C based tests, though unsure how to upstream them. Still looking through so this might not be the only bug, but freeing interleaved objects didn't work since the list was being re-chained backwards.

So: allocate a bunch of objects; enough to have 3 rbuf's, free every other object. then re-allocate half of the objects. it'll leak the chained objects. They may later attempt to re-chain themselves, which would hit the assert you saw.

Give this one a test in your mctester setup?

dormando:
I also do want to add the resp object counter back. You missed the thread locks so I'll get that too.

pkarumanchi9:
> Looks like you caught the missing free = false. I lost that in a refactor.
> 
> I'm starting another change in #693 - just writing C based tests, though unsure how to upstream them. Still looking through so this might not be the only bug, but freeing interleaved objects didn't work since the list was being re-chained backwards.
> 
> So: allocate a bunch of objects; enough to have 3 rbuf's, free every other object. then re-allocate half of the objects. it'll leak the chained objects. They may later attempt to re-chain themselves, which would hit the assert you saw.
> 
> Give this one a test in your mctester setup?

Sure! Let me run this branch with mctester.

pkarumanchi9:
Ran mctester on the debug build and hit the following assertion. This is the case I was referring to [here](https://github.com/memcached/memcached/pull/692#issuecomment-652105432). When resp_free is called on a bundle which is not linked and the thread managing the connection also doesn't point to it.

> (gdb) f 4
#4  0x0000555555563e08 in resp_free (c=0x7fffe8221320, resp=0x7fffe8cdc000) at memcached.c:1110
1110	            assert((b->next || b->prev) || b == th->open_bundle);

dormando:
Ok thanks. You're sure that was on my branch? That assert definitely shouldn't fire. Any bundle which has a refcount below max _must_ be linked.

What's your mctester incantation? did you modify it at all?

pkarumanchi9:
Ran mctester on the debug build and hit another assertion. 

> (gdb) f 4
#4  0x0000555555563b5b in resp_allocate (c=0x7fffe0180e80) at memcached.c:1063
1063	                assert(b->prev == NULL);
(gdb) l
1058
1059	        if (resp != NULL) {
1060	            b->refcount++;
1061	            resp->free = false;
1062	            if (b->refcount == MAX_RESP_PER_BUNDLE) {
1063	                assert(b->prev == NULL);
1064	                // We only allocate off the head. Assign new head.
1065	                th->open_bundle = b->next;
1066	                // Remove ourselves from the list.
1067	                if (b->next) {
(gdb) p b
$1 = (mc_resp_bundle *) 0x7fffe066d768
(gdb) p *b
$2 = {refcount = 13 '\r', next_check = 1 '\001', next = 0x7fffe066d768, prev = 0x7fffe066d768,
  r = 0x7fffe066d780}

Verified that I am on the correct branch.
> $ git show
commit 97a0539a50055153285b3cde8c575b8ff340165b (HEAD -> fix_rbuf_test, origin/fix_rbuf_test)

I've verified with mctester on _`buffer-unity-followup`_ branch and it seems to hold up.

And using the mctester as-is
> git show
 commit 709de4c2767be35fdb431bbb305d5b0c2b6ce633 (HEAD -> master, origin/master, origin/HEAD)

dormando:
Thanks, what are your start arguments though? :)

dormando:
Ok, got my local tests to fail reliably. I'll give it another round later tonight.

pkarumanchi9:
For memcached 
> -c 50000 -m 512 nobody -p 11211 -l 127.0.0.1

For mctester
> ./basic -conncount 20000 -valuesize 10000

The changes in this branch are also running on a prod cluster. It will take couple of days to hit the mem spike issue we noticed before. I will keep you posted on that. Will take a break and check in later.

dormando:
I got it to fail because refcount is maxed and it is linked, and when refcount is 0 and isn't linked :) Both should be invalid conditions, so I've probably got one more link thing backwards. Bit groggy today, will try again later as I said. Would be nice if we (eventually?) get a queue.h-alike so I stop making these mistakes.

dormando:
@pkarumanchi9 okay, pushed another one-liner to the repo. This now passes all my unit tests.

I'll probably end up holding onto my test branch until some further refactoring lets me do proper C tests for these kinds of functions.

pkarumanchi9:
> I got it to fail because refcount is maxed and it is linked, and when refcount is 0 and isn't linked :) Both should be invalid conditions, so I've probably got one more link thing backwards. Bit groggy today, will try again later as I said. Would be nice if we (eventually?) get a queue.h-alike so I stop making these mistakes.

Totally agree with the queue.h like functionality.

dormando:
been wanting to do that since forever but it's a lot of work, hard to verify, and nobody wants to pay for it :)

pkarumanchi9:
> @pkarumanchi9 okay, pushed another one-liner to the repo. This now passes all my unit tests.
> 
> I'll probably end up holding onto my test branch until some further refactoring lets me do proper C tests for these kinds of functions.

That is the smoking gun. I verified on my end as well. I will go to bed peacefully :)

dormando:
Excellent! thanks for checking. I'll add the missing stats and merge it up as soon as I can.

dormando:
@pkarumanchi9 alright I've added the counter. I'm going to prepare for a release (reviewing PR's/merging into `next`), but I'll hold until you can confirm that you're not seeing a leak anymore. That okay?

pkarumanchi9:
Sounds good. I will deploy and let it run for few days to confirm.

One suggestion on the counters -
Since the size of response_obj is different from read_buf, when we look at the stats, it may not give the complete picture of usage of resp bundle vs resp obj vs read buf.

How about the following - 
- Add a stat for number of resp bundles in use since the amount of memory(16kb) managed by the bundle is same as read buf.

response_obj_count        --> This keeps track of just resp objects not giving the full picture. 
read_buf_count                --> shows total read buffers in use (including response bundle).
response_bundle_count  --> This will give the number of 16k buffers used just for response bundle.

This additional stat would have helped in identifying the leak when we see a discrepancy in response_obj_count and response_bundle_count.  If response_bundle_count is going up by a lot., it would have given a clue to us. 

Thoughts?

dormando:
I was thinking about that. On the fence between bytes vs bundle_count, since bundle_count won't have much meaning to an end user unfamiliar with the internals.

On the other hand, with the new stat response_obj_count is _only_ for in-flight requests, vs before it also counted the held-around-forever bytes. So if response_obj_count is slowly increasing you're guaranteed a problem. An idle system should always report 1 for response_obj_count (meaning the obj used to send the stats response:)

dormando:
obj_count alone doesn't tell you the percentage of rbuf's bytes used though; so finishing my thought from the last comment, I would use the byte amount vs the word bundle_count.

pkarumanchi9:
reporting byte amount is much better., that will take the guessing (and additional math) out of the equation for end users.

dormando:
You should be fine to test as-is though? obj_count should follow the traffic patterns and not increase forever.

pkarumanchi9:
I agree, I can go ahead with my testing as-is without the additional stat. I will share the results by early next week.

dormando:
Going to close this. lets continue any discussion in #693

PR #693, opened 2020-07-01
fix leak in merged resp/read buffers

dormando:
the list grows toward next, not prev.

also didn't unmark free for first resp object. (thanks Prudhviraj!)

- [x] add response_obj_count counter back. I removed a little too much.
- [x] add response_obj_bytes counter. which should be size of bundles * rbuf_size.

I think this is right:
- bundles are only dangling if they're at max refcount.
- drop below max refcount and you get linked.
- if refcount hits 0, you must be linked or head.

paranoid there's a case where I'm not clearing prev/next so I'll check one more time.
edit: there wassss. wasn't clearing next.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -1204,6 +1204,8 @@ integers separated by a colon (treat this as a floating point number).
 |                       |         | by the server                             |
 | response_obj_oom      | 64u     | Connections closed by lack of memory      |
 | response_obj_count    | 64u     | Total response objects in use             |
+| response_obj_bytes    | 64u     | Total bytes used for resp. objects. is a  |
+|                       |         | subset of bytes from read_buf_bytes.      |
 | read_buf_count        | 64u     | Total read/resp buffers allocated         |
 | read_buf_bytes        | 64u     | Total read/resp buffer bytes allocated    |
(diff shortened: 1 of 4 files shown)


dormando:
@pkarumanchi9 - think I'm done with this now. might squash and pull into next shortly and do another PR if we have any further fixes.

pkarumanchi9:
Changes look good. 
I've deployed this change in prod and will share the results early next week.

dormando:
@pkarumanchi9 how'd it look?

pkarumanchi9:
Tests are still in-progress. I will need around 2 more days for the workload to run before sharing the results.

dormando:
Oh, huh. I thought it showed up after one peak cycle last time. I will hold the release a few more days I guess

theonajim:
A couple of questions: What are the conditions for triggering the original issue that this PR addresses ? And which versions of memcached have the issue?

dormando:
@theonajim this is fixing something that wasn't in a release yet.

pkarumanchi9:
Test results look good with this change-set. I wanted to run it bit longer to confirm. Buffer usage has grown up gradually and has plateaued. This is good to be released.

dormando:
thanks! think we're just about good to stamp a release, then.

dormando:
Released in 1.6.7!

PR #694, opened 2020-07-01
Split text/binary protocol handlers off of memcached.c

dormando:
First pass at removing mass-code from the memcached.c main file.

I've had this branch for a while and am basing other work on top of it. The text and binary protocol handlers mostly-cleanly pull into their own files. The rest will be more work.

This at least makes it easier to move around and slightly easier on my brain when working on the top level abstractions, so I want to upstream this early to reduce conflicts.


dormando:
urf. that was quite the merge conflict.

dormando:
Released in 1.6.7!

PR #695, opened 2020-07-02
Enhance slab reallocation for burst of evictions

mjhags:
Extension to the existing slab reallocation model to better cope with eviction patterns that appear in short timeframe bursts.

This means that if more than 25% of the total evictions in the window appear in this slab it becomes a candidate for
needing more pages.


dormando:
thanks!

dormando:
Released in 1.6.7!

PR #698, opened 2020-07-04
illumos build fix proposal.

devnexen:
the socket id wrapper not defined under illumos version.
explicit cast to avoid possible pedantic build flags complains (ie can be the old iovec interface)


dormando:
thanks!

why isn't it defined? isn't that tied to the libevent version?

dormando:
I think in sanitycheck() it completely fails to start with a 1.x libevent. Should make this a compile time check.

configure.ac
```diff
@@ -393,6 +394,22 @@ if test $ac_cv_libevent_dir != "(system)"; then
   fi
 fi
 
+AC_RUN_IFELSE(
+   [AC_LANG_PROGRAM([
+#include <event.h>
+   ], [dnl
+const char *ver = event_get_version();
+return (ver != NULL && *ver == '2') ? 0 : 1;
```
dormando, line 402: what if libevent ever releases version 3? :)

configure.ac
```diff
@@ -377,6 +377,7 @@ AC_CACHE_CHECK([for libevent directory], ac_cv_libevent_dir, [
 ])
   fi
 ])
+
```
dormando, line 380: spurious whitespace

configure.ac
dormando: Also, I meant to say that I don't think the daemon even starts with libevent1. There's a runtime check just in case you build against one version but start on another (which is why sanitycheck got added in the first place, I think)

devnexen: yes in fact it s just to make the whole thing compilable :-). if there was no runtime check a compile error pragma would fit better I believe.

dormando:
Hey @devnexen 

Lets wrap this up. :) Any reason why this isn't just an #ifndef -> #define for the missing type, rather than the whole version check?

devnexen:
> Hey @devnexen
> 
> Lets wrap this up. :) Any reason why this isn't just an #ifndef -> #define for the missing type, rather than the whole version check?

I remember it was not that simple having tried this approach still got build issue.

dormando:
any idea what the error was? was it defined in older versions but wrong or something?

devnexen:
It s basically there was no way of getting the version by a constant but programmatically.

dormando:
I mean checking for `evutil_socket_t` being undef and defining it instead of trying to find out if it should be there based on the libevent version?

dormando:
also how old is the libevent on illumos? I remember vaguely checking how long ago evutil_socket_t got added and it was a long time ago.

devnexen:
> I mean checking for `evutil_socket_t` being undef and defining it instead of trying to find out if it should be there based on the libevent version?

Ah right I ve tried this approach first but did not work out fully at the time, indeed it s just a constant originally but I had a build case when the build failed.

dormando:
What was the error/failure? The only way I can think of it failing is if the define existed but was wrong somehow.

devnexen:
I got `evutil_socket_t` undefined type when the old libevent was installed and picked up over the libevent2

dormando:
hmmmmmmmmmmm... suspicious. requires a libevent more than 10 years old to be missing. reading back I think I failed to communicate this earlier: instead of defining it, if the libevent is older than 2.x we shouldn't even compile. There're other features and bugs and things and > 10 years old library isn't reasonable to support :)

It currently fails to start if you don't have at least libevent 2.x, but if you're failing to compile because of something missing that can be made more explicit too.

dormando:
looks good, thanks!

I moved at least one of the lines for the void change so I might apply this on top of that to lessen the merge conflict.

do you actually have an illumos system for testing this or was it a passive curiosity?

devnexen:
Yes I do :)

dormando:
Excellent! So I've merged the other thing into next. if you want to point this branch at next and rebase it, it should be good to go.

dormando:
I did the PR base change for you, and also squashed + edited the commit message and pushed into next. Thanks!

dormando:
released in 1.6.9!

PR #700, opened 2020-07-09
Limit max connections 

pkarumanchi9:
- adding tests for max conns
- removing maxconns_fast related code.


pkarumanchi9:
If you are ok with this, I will remove the code with all maxconn_fast references.

dormando:
Ah, no, please don't remove the maxconns_fast code! the whole point of this exercise is to fix it. The code is supposed to balance the FD's such that when the final FD is used, maxconns_fast triggers and informs the client it's closing the connection because of the connection limit.

dormando:
This did at some point work fine, and I probably broke it somewhere, or it's only broken when you set the maxconns limit really low; I haven't actually looked to confirm myself.

pkarumanchi9:
Ah I see your point. I wanted to initiate the discussion with this PR to get more details around maxconns_fast. 

I think if we account the reserved fd’s correctly, the existing code might just do it. I’ll revert the change and check if we can accurately account for reserved fd’s. Thing out loud here - One idea is to rely on getrlimit to tell us the curr value of open fd’s and compare it with maxconns as opposed to doing the current way of accounting - “curr_conn” + “reserved_fds” <= maxconns.

dormando:
Give it a shot, see how it goes :) Keep in mind things like extstore

pkarumanchi9:
Fixed and verified it on my setup.

Basically the fd can never go beyond the ulimit's that are set for the process. So when a new connection (rather the last fd) is used for establishing the connection, we will be returning "ERROR Too many open connections" to the clients as expected. 

If any of the clients close the connections etc, then for the the subsequent connections, a lower fd will be assigned and we will be ok. This fix should be agnostic of whether we use with extstore, domain socket based solution etc. Also this fix doesn't need to book keep reservedfds any more. I was initially trying to take that route of accounting, but found this approach simple and solving the purpose.  Comments, please?

memcached.c
```diff
@@ -3221,7 +3221,7 @@ static void drive_machine(conn *c) {
             bool reject;
             if (settings.maxconns_fast) {
                 STATS_LOCK();
-                reject = stats_state.curr_conns + stats_state.reserved_fds >= settings.maxconns - 1;
+                reject = sfd >= settings.maxconns - 1;
```
dormando, line 3224: neat! good point. I'll have to give this some deeper thought/testing myself before we merge but on a first pass it seems solid.

since maxconns can't be changed at runtime, we can probably move the STATS_LOCK(); back around just the rejected_conns++ (inside the if). Then there's no lock on connect again.

t/maxconns.t
```diff
@@ -11,22 +11,25 @@ use FindBin qw($Bin);
...
-
-ok(defined($sock), 'Connection 0');
-push (@sockets, $sock);
-
-
-foreach my $conn (1..10) {
-  $sock = $server->new_sock;
-  ok(defined($sock), "Made connection $conn");
-  push(@sockets, $sock);
+my $server = new_memcached('-c 30','','true');
```
dormando, line 14: random "true" arguments are poor usability. just as a nitpick:
new_memcached_nodebug('-c 30'); - which would be a sub that just calls `new_memcached()` with the extra flag. something like that makes it more obvious to the test writer.

t/maxconns.t
```diff
@@ -11,22 +11,25 @@ use FindBin qw($Bin);
...
+my $stat_sock = $server->sock;
+my @sockets = ();
+my $num_sockets;
+my $rejected_conns = 0;
+for (1 .. 5) {
+  my $sock = $server->new_sock;
+  if (defined($sock)) {
+    push (@sockets, $sock);
+  } else {
+    $rejected_conns = $rejected_conns + 1;
```
dormando, line 25: hmm is this actually right? You should be able to read the error message from the socket and validate you're getting it.

t/lib/MemcachedTest.pm
```diff
@@ -255,8 +262,9 @@ sub is_running {
 }
 
 sub new_memcached {
-    my ($args, $passed_port) = @_;
+    my ($args, $passed_port, $passed_non_debug_build_flag) = @_;
     my $port = $passed_port;
+    my $non_debug_build_flag = $passed_non_debug_build_flag;
```
dormando, line 267: don't need this line; just call the above variable non_debug_build_flag. I think $port is being re-assigned because of some old code (it gets re-written sometimes)

t/maxconns.t
pkarumanchi9: good idea. fixed

memcached.c
pkarumanchi9: moved the lock inside the if block.

t/maxconns.t
pkarumanchi9: fixed the test now

t/lib/MemcachedTest.pm
pkarumanchi9: fixed it.

pkarumanchi9:
Couldn't figure out why CI tests are failing. Tests seem to pass on my local setup. I ran multiple times to be sure. And an unrelated test is also failing on CI test `t/lru-maintainer.t`.

dormando:
hm "Seccomp policy failure. Caught syscall 041. This is either an exploit attempt, or your system is not supported yet."

you building with seccomp? I hate that thing :/ have to figure out if change was in CI or mc

dormando:
Took some clicking but looks like the `next` branch passed CI when I last pushed it a week ago.

pkarumanchi9:
I built it with and without seccomp. Ran them locally. The tests passed on both variations.

dormando:
Sorry it took me so long to check: just realized I could re-run tests directly from the actions tab.

The next branch is fine. Your middle commit is fine. The "fix maxconns code path" and later seem to trip the seccomp thing.

Feel up to tracking it down?

t/maxconns.t
```diff
@@ -1,32 +1,42 @@
...
-
-# start up a server with 10 maximum connections
-my $server = new_memcached('-c 100');
-my $sock = $server->sock;
-my @sockets;
-
-ok(defined($sock), 'Connection 0');
-push (@sockets, $sock);
-
+#my $server = new_memcached_nodebug('-c 30');
```
dormando, line 11: can drop the nodebug changes since you're not using it anymore, yes?

dormando:
just the cleanup comment since it looks like you're not using that code anymore. a squash would be good too but I can do that for you if you'd like.

should be ready to go in then, thanks! sorry for the long delay.

t/maxconns.t
pkarumanchi9: yes.

dormando:
looks good! want me to squash it for you?

pkarumanchi9:
yes please!

dormando:
Alright! I rebased, squashed, edited the commit message a bit and pushed to `next`. It'll bake there until the next release.

dormando:
released in 1.6.9!

Issue #684, opened 2020-05-18
time travel debug commands


dormando:
In the ill-fated original 1.5.x branch, one of the things Dustin added was time traveling to avoid sleeps/races for the test suite.

The command would be used to pause the timer event, move `current_time` forward or backward by N seconds, and resume time. This could just compiled on debug mode only. Note that it won't help for cases where we wait for threads to do things, but it'll fix issues like `expirations.t` failing on some platforms, and reduce the amount of time the test suite takes by cutting probably 90% of the sleeps.

This isn't too hard to do but it's not my top priority right now, so putting this up as a note to self or open to someone else if they want to try.

Off the top of my head I'd suggest:
```
debugtime p # pauses clock
debugtime r # resumes clock
debugtime -5 # move 5 seconds into the past (could underflow! that's fine.)
debugtime 30 # move 30 seconds into the future
```

a decision to be made is if debugtime is implemented as a static delta to apply to the internal timer before. IE: if you pause, add 30s, then resume; without holding the delta somewhere time will snap back to 30 seconds ago.

dormando:
Out in 1.6.10

PR #706, opened 2020-08-10
Add time travel debug command

minkikim89:
resolves https://github.com/memcached/memcached/issues/684

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3705,6 +3705,10 @@ static int server_socket_unix(const char *path, int access_mask) {
  */
 volatile rel_time_t current_time;
 static struct event clockevent;
+#ifdef MEMCACHED_DEBUG
+volatile bool is_paused;
+volatile int64_t delta;
+#endif
 #if defined(HAVE_CLOCK_GETTIME) && defined(CLOCK_MONOTONIC)
 static bool monotonic = false;
 static int64_t monotonic_start;
@@ -3740,19 +3744,31 @@ static void clock_handler(const evutil_socket_t fd, const short which, void *arg
     event_base_set(main_base, &clockevent);
     evtimer_add(&clockevent, &t);
 
+#ifdef MEMCACHED_DEBUG
+    if (is_paused) return;
+#endif
+
 #if defined(HAVE_CLOCK_GETTIME) && defined(CLOCK_MONOTONIC)
     if (monotonic) {
         struct timespec ts;
         if (clock_gettime(CLOCK_MONOTONIC, &ts) == -1)
             return;
+#ifdef MEMCACHED_DEBUG
+        current_time = (rel_time_t) (ts.tv_sec - monotonic_start + delta);
+#else
         current_time = (rel_time_t) (ts.tv_sec - monotonic_start);
+#endif
         return;
     }
 #endif
     {
         struct timeval tv;
         gettimeofday(&tv, NULL);
+#ifdef MEMCACHED_DEBUG
+        current_time = (rel_time_t) (tv.tv_sec - process_started + delta);
+#else
         current_time = (rel_time_t) (tv.tv_sec - process_started);
+#endif
     }
 }
 
--- a/proto_text.c
+++ b/proto_text.c
@@ -2067,6 +2067,31 @@ static void process_misbehave_command(conn *c) {
         out_string(c, "OK");
     }
 }
+
+static void process_debugtime_command(conn *c, token_t *tokens, const size_t ntokens) {
+    struct timeval tv;
+
+    if (strcmp(tokens[1].value, "p") == 0) {
+        gettimeofday(&tv, NULL);
+        if (!is_paused) {
+            is_paused = true;
+        }
+    } else if (strcmp(tokens[1].value, "r") == 0) {
+        gettimeofday(&tv, NULL);
+        if (is_paused) {
+            is_paused = false;
+        }
+    } else {
+        int64_t time_delta = 0;
+        if (!safe_strtoll(tokens[1].value, &time_delta)) {
+            out_string(c, "ERROR");
+            return;
+        }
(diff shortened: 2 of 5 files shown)


dormando:
Thanks! I'm running pretty behind right now so it could take a few days to a week to start reviewing this. Many apologies.

minkikim89:
@dormando 
Probably busy, but could you check this?

memcached.c
```diff
@@ -231,6 +231,76 @@ static rel_time_t realtime(const time_t exptime) {
     }
 }
 
+/*
```
dormando, line 234: this was just moved to make the clock bits available to the protocol?

for future reference it's a bit harder to review when chunks of code are moved and changed at the same time.

memcached.c
```diff
@@ -231,6 +231,76 @@ static rel_time_t realtime(const time_t exptime) {
...
+
+        authfile_load(settings.auth_file);
+    }
+
+    evtimer_set(&clockevent, clock_handler, 0);
+    event_base_set(main_base, &clockevent);
+    evtimer_add(&clockevent, &t);
+
+#ifdef MEMCACHED_DEBUG
+    if (is_paused) return;
```
dormando, line 285: this is modified from under a lock, but read without the lock here. also no volatile pin. probably worth just throwing the lock on? else someone will trundle by with their static analyzer and claim a bug on it :/

t/lib/MemcachedTest.pm
```diff
@@ -72,6 +73,16 @@ sub mem_stats {
     return $stats;
 }
 
+sub mem_move_time {
```
dormando, line 76: think I see how the wording on my issue may have complicated things slighty. I want to be able to pause and resume time, /as well as/ move it. So the act of moving time here should simply b `debugtime $mover\r\n`, there shouldn't need to be pause/resume.

I'll add more notes in the top level review bit.

memcached.c
```diff
@@ -5552,6 +5622,41 @@ static void process_misbehave_command(conn *c) {
...
 }
+
+static void process_debugtime_command(conn *c, token_t *tokens, const size_t ntokens) {
+    struct timeval tv;
+
+    if (strcmp(tokens[1].value, "p") == 0) {
+        gettimeofday(&tv, NULL);
+        pthread_mutex_lock(&debugtime_lock);
+        if (!is_paused) {
+            paused_time = (rel_time_t)tv.tv_sec;
```
dormando, line 5633: tracking time spent paused is clever but I think unnecessary. or at least, I can't think of a use case for it and it adds a bit to this code here.

memcached.c
```diff
@@ -231,6 +231,76 @@ static rel_time_t realtime(const time_t exptime) {
...
+#ifdef MEMCACHED_DEBUG
+    if (is_paused) return;
+#endif
+
+#if defined(HAVE_CLOCK_GETTIME) && defined(CLOCK_MONOTONIC)
+    if (monotonic) {
+        struct timespec ts;
+        if (clock_gettime(CLOCK_MONOTONIC, &ts) == -1)
+            return;
+        current_time = (rel_time_t) (ts.tv_sec - monotonic_start);
```
dormando, line 293: So I was thinking the debugtime would add/remove against a `volatile int64_t delta` variable, which would get added here if `MEMCACHED_DEBUG` - then you're not replicating the time code below or changing process_started, which could lead to bugs in test code.

memcached.c
```diff
@@ -5552,6 +5622,41 @@ static void process_misbehave_command(conn *c) {
...
+
+    if (strcmp(tokens[1].value, "p") == 0) {
+        gettimeofday(&tv, NULL);
+        pthread_mutex_lock(&debugtime_lock);
+        if (!is_paused) {
+            paused_time = (rel_time_t)tv.tv_sec;
+            is_paused = true;
+        }
+        pthread_mutex_unlock(&debugtime_lock);
+    } else if (strcmp(tokens[1].value, "r") == 0) {
```
dormando, line 5637: see my other comments about fixing this: here you only apply the change on resume, but the workflow should be:

```
pause clock
run some tests
push time forward 3s
run more tests
push time forward another 2s
run even more tests
resume clock
```
which I don't believe works here since you only update the clock on resume, and I want it to stay paused while the tests go. It's good if the clock can be adjusted without having to pause first too, for hand debugging or other things.

I don't actually have a perfect idea offhand for how to do this cleanly. Can think about it more if you're stuck and I'm more caught up. probably just more logic to be done if `is_paused` is seen in the clock event.

dormando (changes requested):
looks like this got written before my protocol refactoring :( You'll have to update where the code goes at least.

dormando:
really sorry about the long delay. 2020 sucks

minkikim89:
@dormando 
I fixed it. It took a long time sorry.

proto_text.c
```diff
@@ -2067,6 +2067,31 @@ static void process_misbehave_command(conn *c) {
         out_string(c, "OK");
     }
 }
+
+static void process_debugtime_command(conn *c, token_t *tokens, const size_t ntokens) {
+    struct timeval tv;
+
+    if (strcmp(tokens[1].value, "p") == 0) {
+        gettimeofday(&tv, NULL);
```
dormando, line 2075: what's this line doing? it's checkin the clock but nothing uses tv here.

dormando:
Thanks! Much simpler.

Just the one line at a glance, but I need to take a closer look and try to use it.

proto_text.c
minkikim89: sorry. it was just a mistake. i removed it.

minkikim89:
@dormando 
Please check this PR

dormando:
Great, thanks! Sorry for the long wait here.

dormando:
Out in 1.6.10

PR #710, opened 2020-08-19
Check if memory is allocated

klebertarcisio:
Hi everyone,

I think we need to check if memory is really allocated before use it.

Regards

### DIFF
--- a/testapp.c
+++ b/testapp.c
@@ -165,6 +165,9 @@ static enum test_return cache_reuse_test(void)
     int ii;
     cache_t *cache = cache_create("test", sizeof(uint32_t), sizeof(char*),
                                   NULL, NULL);
+    if(cache == NULL){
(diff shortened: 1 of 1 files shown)


dormando:
Hey,

Thanks! I have a couple issues here:

1) Since this is just the testapp I don't think this is strictly necessary. It won't be invoked while under high memory pressure :)
2) The code style seems a bit off.
3) Doesn't actually pass the test suite: whitespace.t is failing.

So I'll probably pass on this, but for future reference please run/check the test output on your own first :)

klebertarcisio:
Hi @dormando, 
Thanks for the feedback.
I removed some whitespaces.
Regards

dormando:
squashed and merged for next, thanks!

dormando:
released in 1.6.9!

PR #711, opened 2020-08-19
arm64: Re-add arm crc32c hw acceleration

AGSaidi:
Use the .arch_extension directive so that a config options and
special cflags aren't required. Add a few tests for both the
software and hardware implementations

tested with both clang and llvm on Ubuntu

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ noinst_PROGRAMS = memcached-debug sizes testapp timedrun
 
 BUILT_SOURCES=
 
-testapp_SOURCES = testapp.c util.c util.h stats_prefix.c stats_prefix.h jenkins_hash.c murmur3_hash.c hash.h cache.c
+testapp_SOURCES = testapp.c util.c util.h stats_prefix.c stats_prefix.h jenkins_hash.c murmur3_hash.c hash.h cache.c crc32c.c
 
 timedrun_SOURCES = timedrun.c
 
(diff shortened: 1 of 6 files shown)


dormando:
Thanks!

Apologies for the delay. On a skim this looks good. I should be catching up with reviews this week. The previous patch for the hw instruction was a lot more invasive within the crc32c file, but this looks more manageable.

martin-g:
I've ran the tests on my aarch64 VM and it successfully uses the hardware CRC32 impl! :+1:

AGSaidi:
Great! It’s also possible to setup native CI using Travis for Arm so we could do that too if you’d find Travis acceptable @dormando.

martin-g:
The repo used TravisCI at some point for testing ARM64 but it was unreliable because of the reduced capacity of the arm containers in Travis: https://github.com/memcached/memcached/commits/master/.travis.yml
Now BuildBot being is used as a CI with dedicated ARM64 node: https://build.memcached.org/#/

martin-g:
Hm, I've just checked BuildBot and none of the builders has built anything since 2 months. @dormando you just released 1.6.7 with many merges from `next` to `master`. Maybe something has stuck ?!

dormando:
The buildbot workers have been off since a power outage here. I had to get a new UPS for it; but mainly there weren't any commits in next since two months ago. :) I added a tiny change to a test and that's it, was just really slow getting the release out while focusing on other things.

The master is still up so the VM you gave me should still be running though.

dormando:
Lovely and clean change! Merged into the next branch. Got my buildbots back in order so hopefully it'll pass the full set fine.

dormando:
released in 1.6.9!

PR #712, opened 2020-08-19
Introduce NAPI ID based worker thread selection

ssamudrala:
By default memcached assigns connections to worker threads in
a round-robin manner. This patch introduces an option to select
a worker thread based on the incoming connection's NAPI ID if
SO_INCOMING_NAPI_ID socket option is supported by the OS.

This allows a memcached worker thread to be associated with a
NIC HW receive queue and service all the connection requests
received on a specific RX queue. This mapping between a memcached
thread and a HW NIC queue streamlines the flow of data from the
NIC to the application. In addition, an optimal path with reduced
context switches is possible, if epoll based busy polling
(sysctl -w net.core.busy_poll = <non-zero value>) is also enabled.

This feature is enabled via a new command line parameter -N <num>
or "--napi_ids=<num>", where <num> is the number of available/assigned
NIC hardware RX queues through which the connections can be received.
The number of napi_ids specified cannot be less than the number
of worker threads specified using -t/--threads option.
If the option is not specified, or the conditions not met, the code
defaults to round robin thread selection.

Signed-off-by: Kiran Patil <kiran.patil@intel.com>
Signed-off-by: Sridhar Samudrala <sridhar.samudrala@intel.com>

### DIFF
--- a/doc/napi_ids.txt
+++ b/doc/napi_ids.txt
@@ -1,0 +1,41 @@
+NAPI ID based worker thread selection
+  -N <num_napi_ids>  | --napi_ids=<num_napi_ids>
+
+By default memcached assigns connections to worker threads in a round-robin
+manner. NAPI ID based worker thread selection enables worker threads to be
+selected based on the NIC HW RX queue on which the incoming request is
+received.
+
+This is enabled using SO_INCOMING_NAPI_ID socket option that is supported
+in linux kernels 4.12 or higher. This socket option returns a system level
+unique ID called NAPI ID that is associated with a RX queue on which the
+last packet associated with that socket is received.
+
+This allows memcached to split the incoming flows among threads based on
+the RX queue on which they are received. Each worker thread is associated
+with a NIC HW receive queue and services all the connection requests
+received on a specific RX queue. This mapping between a memcached thread
+and a HW NIC queue streamlines the flow of data from the NIC to the
+application. In addition, an optimal path with reduced context switches is
+possible, if epoll based busy polling
+(sysctl -w net.core.busy_poll = <non-zero value>) is also enabled.
+
+This feature is enabled via a new command line parameter -N <num> or
+"--napi_ids=<num>", where <num> is the number of available/assigned NIC
+hardware RX queues through which requests associated with a connection are
+received. The number of napi_ids specified cannot be greater than the number
+of worker threads specified using -t/--threads option. If the option is
+not specified, or the conditions not met, the code defaults to round robin
+thread selection.
+
+During a normal run, each worker thread gets associated with a napi_id and
+this will establish a 1:1 mapping between the thread and queues. If a new
+napi_id is received after each thread is associated with its own napi_id
+(this can happen if num_napi_ids argument doesn't exactly match with the
+number of RX queues OR if the NIC driver goes through a reload), a stats
+error counter called 'unexpected_napi_ids' is incremented and all the
+napi_id's associated with the threads are reset.
+
+If -N option is used, but the connection requests are received from a
+virtual interface like loopback, napi_id returned can be 0. This condition
+is tracked via a stats counter called 'round_robin_fallback'.
--- a/memcached.c
+++ b/memcached.c
@@ -63,6 +63,10 @@
 #include <sys/sysctl.h>
 #endif
 
+#ifndef SO_INCOMING_NAPI_ID
+#define SO_INCOMING_NAPI_ID 56
+#endif
+
 /*
  * forward declarations
  */
@@ -293,6 +297,7 @@ static void settings_init(void) {
 #ifdef MEMCACHED_DEBUG
     settings.relaxed_privileges = false;
 #endif
+    settings.num_napi_ids = 0;
 }
 
 extern pthread_mutex_t conn_lock;
@@ -1837,6 +1842,8 @@ void server_stats(ADD_STAT add_stats, conn *c) {
         APPEND_STAT("time_since_server_cert_refresh", "%u", now - settings.ssl_last_cert_refresh_time);
     }
 #endif
+    APPEND_STAT("unexpected_napi_ids", "%llu", (unsigned long long)stats.unexpected_napi_ids);
+    APPEND_STAT("round_robin_fallback", "%llu", (unsigned long long)stats.round_robin_fallback);
 }
 
 void process_stat_settings(ADD_STAT add_stats, void *c) {
@@ -1921,6 +1928,7 @@ void process_stat_settings(ADD_STAT add_stats, void *c) {
     APPEND_STAT("ssl_wbuf_size", "%u", settings.ssl_wbuf_size);
     APPEND_STAT("ssl_session_cache", "%s", settings.ssl_session_cache ? "yes" : "no");
 #endif
+    APPEND_STAT("num_napi_ids", "%s", settings.num_napi_ids);
 }
 
 static int nz_strcmp(int nzlength, const char *nz, const char *z) {
@@ -3708,6 +3716,16 @@ static int server_socket(const char *interface,
             continue;
         }
 
+        if (settings.num_napi_ids) {
+            socklen_t len = sizeof(socklen_t);
+            int napi_id;
+            error = getsockopt(sfd, SOL_SOCKET, SO_INCOMING_NAPI_ID, &napi_id, &len);
+            if (error != 0) {
+                fprintf(stderr, "-N <num_napi_ids> option not supported\n");
+                exit(EXIT_FAILURE);
+            }
+        }
+
 #ifdef IPV6_V6ONLY
         if (next->ai_family == AF_INET6) {
             error = setsockopt(sfd, IPPROTO_IPV6, IPV6_V6ONLY, (char *) &flags, sizeof(flags));
@@ -4231,6 +4249,7 @@ static void usage(void) {
     verify_default("ssl_keyformat", settings.ssl_keyformat == SSL_FILETYPE_PEM);
     verify_default("ssl_verify_mode", settings.ssl_verify_mode == SSL_VERIFY_NONE);
 #endif
+    printf("-N, --napi_ids            number of napi ids. see doc/napi_ids.txt for more details\n");
     return;
 }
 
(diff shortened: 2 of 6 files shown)


ssamudrala:
This is an update of the same feature submitted earlier by Kiran addressing the feedback.  Please review.

dormando:
Thanks! Sorry I haven't been able to review it yet. I should be getting to it within the next week.

ssamudrala:
Rebased the patch to the latest master. Please review.

doc/napi_ids.txt
```diff
@@ -0,0 +1,37 @@
+NAPI ID based worker thread selection
+  -N <num_napi_ids>  | --napi_ids=<num_napi_ids>
+
+By default memcached assigns connections to worker threads in a round-robin
+manner. NAPI ID based worker thread selection enables worker threads to be
+selected based on the NIC HW RX queue on which the incoming request is
+received.
+
+This is enabled using SO_INCOMING_NAPI_ID socket option that is supported
```
thinkingfish, line 9: This socket option doesn't seem to be part of the socket man page. Any plan on filling in the gap there?

ssamudrala: Yes. looks like we missed submitting this update to the socket man page Will do soon.

doc/napi_ids.txt
```diff
@@ -0,0 +1,37 @@
...
+received on a specific RX queue. This mapping between a memcached thread
+and a HW NIC queue streamlines the flow of data from the NIC to the
+application. In addition, an optimal path with reduced context switches is
+possible, if epoll based busy polling
+(sysctl -w net.core.busy_poll = <non-zero value>) is also enabled.
+
+This feature is enabled via a new command line parameter -N <num> or
+"--napi_ids=<num>", where <num> is the number of available/assigned NIC
+hardware RX queues through which requests associated with a connection are
+received. The number of napi_ids specified cannot be less than the number
```
dormando, line 26: this says the number of napi ids cannot be less than worker threads, but the code says that it can't be greater than. which is true?

memcached.c
```diff
@@ -3679,6 +3685,8 @@ static int server_socket(const char *interface,
     int error;
     int success = 0;
     int flags =1;
+    socklen_t len;
+    int napi_id;
```
dormando, line 3689: doesn't seem to be used anywhere?

I've been moving some of these temp variables to be defined closer to where they're used. ie: in this case both of these are only used inside the `if (settings.num_napi_ids)` block so might as well stuff them in there.

memcached.c
```diff
@@ -4232,6 +4249,7 @@ static void usage(void) {
     verify_default("ssl_keyformat", settings.ssl_keyformat == SSL_FILETYPE_PEM);
     verify_default("ssl_verify_mode", settings.ssl_verify_mode == SSL_VERIFY_NONE);
 #endif
+    printf("-N, --napi_ids            max number of napi ids for NAPI ID based thread selection\n");
```
dormando, line 4252: add a "see doc/napi_ids.txt for detail" or similar? Also saying NAPI twice is kinda silly. "max number of ids for NAPI ID based thread" :)

thread.c
```diff
@@ -597,11 +665,12 @@ void dispatch_conn_new(int sfd, enum conn_states init_state, int event_flags,
         return;
     }
 
-    int tid = (last_thread + 1) % settings.num_threads;
-
-    LIBEVENT_THREAD *thread = threads + tid;
+    LIBEVENT_THREAD *thread;
 
-    last_thread = tid;
+    if (settings.num_napi_ids)
```
dormando, line 670: I'd probably flip this so the default branch is the common case (round robin) but I'm not going to be strict about it here.

thread.c
```diff
@@ -581,6 +581,74 @@ static void thread_libevent_process(evutil_socket_t fd, short which, void *arg)
...
+{
+    LIBEVENT_THREAD *thread;
+    int napi_id, err, i;
+    socklen_t len;
+    int tid = -1;
+
+    len = sizeof(socklen_t);
+    err = getsockopt(sfd, SOL_SOCKET, SO_INCOMING_NAPI_ID, &napi_id, &len);
+    if ((err == -1) || (napi_id == 0))
+        return select_thread_round_robin();
```
dormando, line 623: so, can you be stuck permanently in this case? I feel like I had a complaint on the original thread where you can enter a failure mode like this but there's no way for the user to know the feature isn't working.

thread.c
```diff
@@ -581,6 +581,74 @@ static void thread_libevent_process(evutil_socket_t fd, short which, void *arg)
...
+             break;
+         }
+         if (thread->napi_id == napi_id) {
+             tid = i;
+             break;
+         }
+    }
+
+    if (tid == -1) {
+        fprintf(stderr, "Connection received with unexpected NAPI ID. Resetting all napi ids associated with worker threads. You may need to restart the application if you are noticing any performance drops.\n");
```
dormando, line 641: been trying to not add more stderr stuff. there's the LOGGER_LOG bits (which are super complex, sorry...) and stats counters which people do tend to monitor. ie; record the number of times it gets reset.

what situation would cause performance drops?

dormando: (er, record the number of times it gets reset and document what that means in doc/protocol.txt)

dormando (changes requested):
mostly nitpicks and small things. Sorry for disappearing again. 2020 etc.

doc/napi_ids.txt
ssamudrala: It can't be greater than. will fix

memcached.c
ssamudrala: ok.

memcached.c
ssamudrala: ok

thread.c
ssamudrala: ok.

thread.c
ssamudrala: This can happen only if the user specifies -N option and the requests are coming from a virtual interface like loopback.

thread.c
ssamudrala: This condition is documented towards the end of napi_ids.txt
    During a normal run, each worker thread gets associated with a napi_id and
    this will establish a 1:1 mapping between the thread and queues. If a new
    napi_id is received after each thread is associated with its own napi_id
    (this can happen if num_napi_ids argument doesn't exactly match with the
    number of RX queues OR if the NIC driver goes through a reload), an error
    message is logged and all the napi_id's associated with the threads are
    reset.

Are you suggesting removing the fprintf() and add a counter to record the number of times reset is called? Can you give an example of such counter that is documented in doc/protocol.txt?

doc/napi_ids.txt
ssamudrala: Done. Submitted the update and you can see the commit here.
  socket.7: Document SO_INCOMING_NAPI_ID
  https://github.com/mkerrisk/man-pages/commit/e8500ecc786f5ee5854d4aa810dcf5124e8ed23f

thread.c
dormando: listen_disabled_num (poorly named, but is what it is..), oom counters, etc. there're a few of them.

I read the doc about things being reset, but "being reset" doesn't describe the impact to the running service? halve performance, errors, etc?

thread.c
dormando: so how does the user know what to do and how to fix it?

thread.c
ssamudrala: The performance impact would be dependent on the real number of HW queues and actual number of napi_ids, worker threads specified. I don't think i can quantify the impact. Basically this is a misconfiguration and incorrect usage of -N option.

dormando: Ok, well just to repeat my _intention_ here: managing expectations for end users. Whatever you can think of to help with that is great.

thread.c
ssamudrala: It is a misconfiguration. The user should not use -N option if he is using virtual interfaces.  -N should be used only if the user is aware and can control the exact number of NIC receive queues through which the connections are received.
Should i add a counter to track this condition too?

dormando: A counter or status code or something? I think I had this struggle with the original PR too: I think you folks are too close to the work. Imagine being some poor sod who downloads this thing and tries to make it work, except they don't work for intel and have never heard of or seen this before. Worse yet, memcached came from a package install so they'll have to be told to go out of their way to look up what a counter means or what an option means. Memcached has _millions_ of installs.

If a user starts a piece of software and it goes into a failure case like this, you get two outcomes:

1) User gets the same, or worse, performance as before. Gets mad at someone (me, you, etc) saying that the feature doesn't work.
2) User deploys without validating before/after performance. Doesn't have any idea if it's working or not, but it starts so they assume it does. Doesn't get mad at anyone but this is super depressing to imagine.

Or even worse still:
3) An admin deploys some change to the production fleet, which causes a driver reload/change/something and suddenly cluster performance tanks and their service/company goes down completely. There are no counters to compare and sparse documentation about what may have suddenly stopped working or why. This extends the amount of time it takes to get the website back online.

Which is a misery I've enjoyed at too many jobs. :)

ssamudrala: OK. I will add a counter and also add some documentation to the napi_ids.txt file to explain when we can hit this condition.

ssamudrala:
> mostly nitpicks and small things. Sorry for disappearing again. 2020 etc.

Submitted another revision based on the feedback. Please review.

dormando (approved):
No more complaints, yay!

dormando:
Alright this is now merged into the next branch. I typically keep PR's open until the next version ships. The next release may be a few weeks off since we just cut 1.6.8.

dormando:
build failure on old/other platforms:

```thread.c:627:    err = getsockopt(sfd, SOL_SOCKET, SO_INCOMING_NAPI_ID, &napi_id, &len);```
since the ifndef/def is handled in memcached.c only:
```
memcached.c:64:#ifndef SO_INCOMING_NAPI_ID
memcached.c:65:#define SO_INCOMING_NAPI_ID 56
```

Need to finish fixing buildbot try to catch these before pushing to next... but at least that's why they automatically run on every push to next :)

going to just move the ifndef/def to memcached.h in a fix patch.

dormando:
though I also wonder now if the feature should specifically linux-gate. figured it would gracefully fail on other platforms but now I'm not as sure.

ssamudrala:
> though I also wonder now if the feature should specifically linux-gate. figured it would gracefully fail on other platforms but now I'm not as sure.

Sure. This should be linux only feature as the socket option and napi_id is a linux networking concept.

dormando:
released in 1.6.9!

will do the linux gating if we get any complaints.

PR #714, opened 2020-08-30
generalized deferred IO queue

dormando:
This turns extstore's deferred IO queue into a more general subsystem, allowing other features to hook into it simultaneously. Pulling this work out of another feature I've been working on. 

This finishes abstracting extstore via the storage.c interfacing, which is now the only consumer of extstore.h. Moving storage.c to be a wrapper itself should allow more experimentation or dynamic backends. Finally, this removes the compile gating around the IO queue code.

Together this removes quite a lot of ifdef's, fixes layer violations, and fixes some naming to hopefully make the code more clear.

One more change added: io_pending_t now reserves space at the end. users cast it into a more specific structure, removing the extra allocations and storage-specific fields from io_pending_t.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -517,6 +517,14 @@ void conn_close_idle(conn *c) {
 
 /* bring conn back from a sidethread. could have had its event base moved. */
 void conn_worker_readd(conn *c) {
+    if (c->state == conn_io_queue) {
+        c->io_queues_submitted--;
+        // If we're still waiting for other queues to return, don't re-add the
+        // connection yet.
+        if (c->io_queues_submitted != 0) {
+            return;
+        }
+    }
     c->ev_flags = EV_READ | EV_PERSIST;
     event_set(&c->event, c->sfd, c->ev_flags, event_handler, (void *)c);
     event_base_set(c->thread->base, &c->event);
@@ -530,13 +538,8 @@ void conn_worker_readd(conn *c) {
     if (c->state == conn_closing) {
         drive_machine(c);
         return;
-    }
-
-    // If we had IO objects, process
-    if (c->io_queued) {
-        c->io_queued = false;
-        // state should be conn_io_queue already, so it will know how to
-        // dequeue and finalize the async work.
+    } else if (c->state == conn_io_queue) {
+        // machine will know how to return based on secondary state.
         drive_machine(c);
     } else {
         conn_set_state(c, conn_new_cmd);
@@ -692,7 +695,7 @@ conn *conn_new(const int sfd, enum conn_states init_state,
     c->last_cmd_time = current_time; /* initialize for idle kicker */
     // wipe all queues.
     memset(c->io_queues, 0, sizeof(c->io_queues));
-    c->io_pending = 0;
+    c->io_queues_submitted = 0;
 
     c->item = 0;
 
@@ -3171,26 +3174,25 @@ static void drive_machine(conn *c) {
              * remove the connection from the worker thread and dispatch the
              * IO queue
              */
-            if (c->io_pending) {
-                assert(c->io_queued == false);
-                conn_set_state(c, conn_io_queue);
-                event_del(&c->event);
-                c->io_queued = true;
-                // TODO: write as for loop?
-                /*io_queue_t *q = c->io_queues;
-                while (q->type != IO_QUEUE_NONE) {
-                    if (q->stack_ctx != NULL) {
-                        q->submit_cb(q->ctx, q->stack_ctx);
-                    }
-                    q++;
-                }*/
+            if (c->io_queues[0].type != IO_QUEUE_NONE) {
+                assert(c->io_queues_submitted == 0);
+                bool hit = false;
+
                 for (io_queue_t *q = c->io_queues; q->type != IO_QUEUE_NONE; q++) {
-                    if (q->stack_ctx != NULL) {
+                    if (q->count != 0) {
+                        assert(q->stack_ctx != NULL);
+                        hit = true;
                         q->submit_cb(q->ctx, q->stack_ctx);
+                        c->io_queues_submitted++;
                     }
                 }
-                stop = true;
-                break;
+                if (hit) {
+                    conn_set_state(c, conn_io_queue);
+                    event_del(&c->event);
+
+                    stop = true;
+                    break;
+                }
             }
 
             switch (!IS_UDP(c->transport) ? transmit(c) : transmit_udp(c)) {
--- a/memcached.c
+++ b/memcached.c
@@ -533,16 +533,14 @@ void conn_worker_readd(conn *c) {
     }
     c->state = conn_new_cmd;
 
-#ifdef EXTSTORE
     // If we had IO objects, process
     if (c->io_queued) {
         c->io_queued = false;
         conn_set_state(c, conn_mwrite);
         drive_machine(c);
     }
-#endif
 }
-#ifdef EXTSTORE
+
 void conn_io_queue_add(conn *c, int type, void *ctx, io_queue_add_cb cb, io_queue_free_cb free_cb) {
     io_queue_t *q = c->io_queues;
     while (q->type != IO_QUEUE_NONE) {
@@ -565,7 +563,7 @@ io_queue_t *conn_io_queue_get(conn *c, int type) {
     }
     return NULL;
 }
-#endif
+
 conn *conn_new(const int sfd, enum conn_states init_state,
                 const int event_flags,
                 const int read_buffer_size, enum network_transport transport,
@@ -671,11 +669,9 @@ conn *conn_new(const int sfd, enum conn_states init_state,
     c->mset_res = false;
     c->close_after_write = false;
     c->last_cmd_time = current_time; /* initialize for idle kicker */
-#ifdef EXTSTORE
     // wipe all queues.
     memset(c->io_queues, 0, sizeof(c->io_queues));
     c->io_pending = 0;
-#endif
 
     c->item = 0;
 
@@ -751,7 +747,6 @@ void conn_release_items(conn *c) {
         c->item = 0;
     }
 
-#ifdef EXTSTORE
     io_queue_t *q = c->io_queues;
     while (q->type != IO_QUEUE_NONE) {
         if (q->head_pending) {
@@ -766,7 +761,6 @@ void conn_release_items(conn *c) {
         }
         q++;
     }
-#endif
 
     // Cull any unsent responses.
     if (c->resp_head) {
@@ -3159,7 +3153,6 @@ static void drive_machine(conn *c) {
 
         case conn_write:
         case conn_mwrite:
-#ifdef EXTSTORE
             /* have side IO's that must process before transmit() can run.
              * remove the connection from the worker thread and dispatch the
              * IO queue
@@ -3180,7 +3173,7 @@ static void drive_machine(conn *c) {
                 stop = true;
                 break;
             }
-#endif
+
             switch (!IS_UDP(c->transport) ? transmit(c) : transmit_udp(c)) {
             case TRANSMIT_COMPLETE:
                 if (c->state == conn_mwrite) {
--- a/memcached.c
+++ b/memcached.c
@@ -14,9 +14,7 @@
  *      Brad Fitzpatrick <brad@danga.com>
  */
 #include "memcached.h"
-#ifdef EXTSTORE
 #include "storage.h"
-#endif
 #include "authfile.h"
 #include "restart.h"
 #include <sys/stat.h>
@@ -102,9 +100,6 @@ static void conn_init(void);
 static bool update_event(conn *c, const int new_flags);
 static void complete_nread(conn *c);
 
-#ifdef EXTSTORE
-static void _get_extstore_cb(void *e, obj_io *io, int ret);
-#endif
 static void conn_free(conn *c);
 
 /** exported globals **/
@@ -1649,9 +1644,6 @@ void server_stats(ADD_STAT add_stats, conn *c) {
     threadlocal_stats_aggregate(&thread_stats);
     struct slab_stats slab_stats;
     slab_stats_aggregate(&thread_stats, &slab_stats);
-#ifdef EXTSTORE
-    struct extstore_stats st;
-#endif
 #ifndef WIN32
     struct rusage usage;
     getrusage(RUSAGE_SELF, &usage);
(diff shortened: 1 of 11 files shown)


dormando:
bug I spotted after reviewing my notes: since the queues can be handled by different threads c->io_pending can't be decremented from side threads (not without a lock, anyway). I left a note for this but forgot :(

have a couple options for fixing. need to think it through.

dormando:
looks like I should also add a proper state machine state for running finalizers inline, to allow callbacks to run on the main thread. need this for the main project + also allows extstore to get the callback processing off of the IO threads.

dormando:
this is slightly uglier than I would've liked but passes tests. will try to use it in my feature and bake this for a while before merging into next.

dormando:
how does this pass tests... :( recache_or_free() is getting called before sending the items back now, which will usually free them. I guess that will work but ends up being a race. Thought I had better assert's over this.

dormando:
Now that this is all written out it's clear I have an ownership issue:
 - mc_resp's are the top level object. but they do not contain io_pending's. io_pending's contain mc_resp's.
 - hard to avoid accidentally freeing mc_resp before the io_pending is handled.
 - mc_resp's are freed on connection close and inline with transmit(), post-send() call.

This was easy to avoid with just extstore as a user, since the pending IO objects were simply released during conn_clean_items(), which runs outside, after, the transmit() call or during connection close.

I want re-queueable IO's. This also means the linked list chain of pending IO's is problematic, since only some IO's may requeue. This may work better if I expand the write_and_free system for mc_resp. Re-use the write_and_free ptr as a context (io_pending) and add a function ptr for how to deal with it.

- io_pending's are tagged onto mc_resp's (thus a resp can't exist in multiple pending IO's)
- io_pending's hold a reference to their queue type object
- io_pending's get a state machine: [unsent, sent, finalized]
- c->io_pending++ for each one added.
- on conn_mwrite: loop through all resp's. for each with a pending IO in unsent state, stack IO and mark sent.
- afterward call the queue's submit function.
then we bail the loop and wait.
when the connection is readded, it gets state (conn_io_queued)
- loop all resp's again. for each sent pending IO, if a post-callback exists; call it. this may change the IO's state back from sent to unsent. or sent to finalized. if no cb, mark as finalized.
- change state to conn_mwrite
- if more IO's are pending, loop, else:
- transmit()
- in transmit_post(), resp_finish()'s are called for each completed mc_resp. call finalizers for the IO's and free the main objects. frees the resp's.
- conn_release_items also calls resp_finish(), which should finalize and release the IO objects just the same, if the connection is closed. finalizers can use the io_pending's state to know if it was unsent/sent/finalized and do appropriate cleanup.

This fixes:
- ownership is now mc_resp -> IO in all cases. mc_resp cannot be freed before the IO object. any calls to free mc_resp also frees the IO.
- since the linked list is set up just prior to submission we don't have to worry about unlinking/relinking IO objects later on a looped retry.

This sets us up for (in the future):
- being able to transmit() mc_resp's that are completed, while waiting for deferred IO's as necessary. Due to strict response ordering this will transmit up to the first deferred object.
- allowing sorting of the resp list. prior to transmit, if the connection is in a clean state (ie; all responses are re-orderable), the list can be sorted. Either just push deferred IO to the back and submit what is ready, sort by response size, or so on.

Still some complications... for extstore the io_pending's themselves don't need to be stacked at all (and only were so they could be freed later), but the sub-IO objects for the underlying extstore_submit() call itself must be stacked. This is currently done as objects are queued but could be done late instead via callbacks and contexts.

Not clear yet if this solves my other use case well enough. It's also a much bigger change than I was hoping for, so I'll see if I can punt with something simpler for now :/

dormando:
Maybe a halfway point between what I have and what the above garbage is:

- Inline chain IO objects as-is. They go into q->head_pending.
- new: reference io_pending into mc_resp.
- conn_mwrite: if (io_pending) { for each queue: q->submit_callback(stack) }
- move q->head_pending to q->head_sent
- set state to conn_io_queue
- /note/ we _never_ return to the worker thread until all IO's are processed. Else this doesn't work.
- return to worker
- walk over head_sent, call q->complete_cb(q, io_pending);
- if it wants to re-queue, inline stack into q->head_pending. c->io_pending++
- set state conn_mwrite
- **at this point**: q->head_sent is garbage. the items within should be considered (and may need to be) unchained.
- if c->io_pending, loop again. when q->head_sent gets overwritten, this is fine.
- transmit()
- resp_finish() -> if io_pending object, call its handler's finalize callback.

We no longer need to track the chain of IO objects in order to free them, they free with the resp's (and any time a resp would free or reset), and IO's can requeue since once they've been processed the order doesn't matter.

I dislike some jank here; feels like two separate systems. If I can come up with a better API for late-stacking and sticking to just walking resp's that might be better.

dormando:
trying again:

- io_queue_t gets an additional `void *stack_ctx`. remove head_pending/head_sent. remove io_pending->next
- mc_resp gets io_pending ptr (+ io_queue ptr?)
- when objects add to a queue they use stack_ctx for any batching/ordering. this can be a struct (init'ed with the queue) or just a pointer. extstore just needs a pointer.
- conn_mwrite: if (c->io_pending) { submit the stack_ctx ptr }
- state -> conn_io_queue. wait.
- conn_io_queue: tmp = q->stack_ctx; q->stack_ctx = NULL; q->complete_cb(c, q, tmp)
- IO's can stack into q->stack_ctx again.
- state -> conn_mwrite
- if io_pending -> GOTO etc.
- resp_finish(): if (resp->io_pending) { r->io->q->finalize_cb(io); free(r->io); }

no excess resp looping. adds a branch in resp_finish(). removes double linked listing.

_should_ allow for mc_resp reordering. could add a state/flag into mc_resp to know if it's "ready", which transmit_pre() can use to filter with if reordering is allowed.

dormando:
This latest attempt feels a lot better. Going back to the main project to see if it works this time...

dormando:
I was able to successfully make use of this! Will give it a further audit and then stage into next after I clear the release queue.

if anyone else wants to eyeball it that's always great too :)

dormando:
Just a note: still incomplete until a lock/atomic/restructuring around `c->io_pending` is done. I'm punting it in case of being struck by inspiration. The only function the counter serves is as a note for when to ship the connection back to the main worker thread. It's also used to check for pending IO's before shipping to side threads, but that can be replaced with something that doesn't need locking around to to avoid locks in the worker thread.

dormando:
Maybe: `final` flag in io_queue_t (though it'll cause some byte waste?). on submission, mark the last item as final. during callback handling, can check for the final flag and redispatch. No counters/atomics to check before resubmission.

dormando:
Okay. Found a new method: passes tests, doesn't use a mutex, appears to be equivalent.

dormando:
Gave it another read... aside from style/cleanliness fixups I don't see anything problematic.

Think I'll chuck this into `next` and let it bake there.

memcached.h
```diff
@@ -725,11 +751,10 @@ struct conn {
     /* data for the swallow state */
     int    sbytes;    /* how many bytes to swallow */
 
+    int io_queues_submitted; /* see notes on io_queue_t */
+    io_queue_t io_queues[3]; /* set of deferred IO queues. */
```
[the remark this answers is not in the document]
dormando: short answer: because the possible queues are extstore, proxy, and the final must be NULL.

longer answer: eventually this could be used by a DSO system hook, so at that point it would change to either a realloc as hooks get added or a linked list. Hardcoding was simplest for now.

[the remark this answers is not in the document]
pkarumanchi9: Hi @dormando  - Few questions not specfic to this change-set, but came across while reviewing -

How are we protected by parallel extstore threads writing to the same pages? 
I see the following mutex initialized, but not used anywhere in the code. 
pthread_mutex_init(&e->pages[i].mutex, NULL);

typedef struct _store_page {
    pthread_mutex_t mutex; /* Need to be held for most operations */          <<<<<<
...
} store_page;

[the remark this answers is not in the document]
dormando: that mutex is hit in a bunch of places in extstore.c, usually called like `p->mutex`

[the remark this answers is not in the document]
pkarumanchi9: Oh yes, that makes sense. Thank you.

thread.c
```diff
@@ -538,6 +536,14 @@ static void thread_libevent_process(evutil_socket_t fd, short which, void *arg)
                     }
                 } else {
                     c->thread = me;
+#ifdef EXTSTORE
+                    if (c->thread->storage) {
+                        conn_io_queue_add(c, IO_QUEUE_EXTSTORE, c->thread->storage,
+                            storage_submit_cb, storage_complete_cb, storage_finalize_cb);
+                    }
+#endif
+                    conn_io_queue_add(c, IO_QUEUE_NONE, NULL, NULL, NULL, NULL);
```
pkarumanchi9, line 545: 1. what will happen if this line is not present? Does it have any impact on non-extstore code path?  

2. When conn_new_cmd state is seen and if extstore is enabled, we take the io_queue for IO_QUEUE_EXTSTORE. And the io_queue entry will be populated by "storage_get_item()".  storage_submit_cb gets called which will assign the request to one of the extstore IO threads by calling "

From the code, I didn't notice any precedence of processing READ's over WRITE's. What If the extstore io thread's queues have lot of write requests queued up before the READ request?

memcached.c
```diff
@@ -535,17 +538,54 @@ void conn_worker_readd(conn *c) {
...
+    while (q->type != IO_QUEUE_NONE) {
+        // Reuse the same submit stack. We zero it out first so callbacks can
+        // queue new IO's if necessary.
+        if (q->stack_ctx) {
+            void *tmp = q->stack_ctx;
+            q->stack_ctx = NULL;
+            q->complete_cb(q->ctx, tmp);
+        }
+        q++;
     }
```
pkarumanchi9, line 588: should we call memset on the queue we processed after calling the complete_cb.

pkarumanchi9: I think I see it now..
since the same queue structure is used for all extstore specific code for the life of the connection, this doesn't need to be zeroed out and it will be reused. That makes sense. That's why stack_ctx is set to null to make it usable for the next IO to be submitted.

thread.c
dormando: 1. IO_QUEUE_NONE? I think it's just a formality here; the last one should always be the NULL object so other code that looks at queues can be safely generalized rather than having to also look at a global queue count somewhere.
2. WRITE's within extstore are batched, and not actually submitted at all through the deferred IO queue process. That's all in storage.c's background thread.

Extstore throttles actual IO write calls such that only one buffer can be pending to disk at a time. (per bucket type, I think?). So the writeback thread bounces and retries if it tries to flush objects while a write is pending. Normally writes just write to the memory buffer so they don't interact with the IO system at all.

memcached.c
dormando: Yes the queues are "global" (per worker thread) structures.

thread.c
dormando: might be handy to keep this open as a reference: https://github.com/memcached/memcached/wiki/Extstore#technical-breakdown

pkarumanchi9: Thanks for the pointers. That's very helpful.

2. WRITE's within extstore are batched, and not actually submitted at all through the deferred IO queue process. That's all in storage.c's background thread.

I see that the deferred queue is only used by the READ code. My question is around processing of the read/write requests(OBJ_IO_READ/OBJ_IO_WRITE) by the extstore io thread itself -  "extstore_io_thread". The queue I was referring to is store_io_thread's obj_io queue and not the deferred queue. 

My original question was around processing of the obj_io queue. Shouldn't we be giving READ's higher preference than the WRITE's? Or Is this not a concern at all?

memcached.c
pkarumanchi9: Thank you. Did you mean to say per worker thread or per connection structures?

thread.c
dormando: No it hasn't been a problem anywhere I've seen. The entire design constricts the number of OBJ_IO_WRITE's that can be in flight at a time to 1-2, and self-throttles so they can't queue on top of each other. So delaying writes doesn't do much besides increase the risk of running out of RAM since it would slow down writes. Between each OBJ_IO_WRITE all OBJ_IO_READ's will complete, and due to the # of IO threads other reads will still process regardless. I was going to dedicate write threads initially but it didn't seem to affect the benchmarks.

The big trick with the way extstore does writes is that typical NAND drives have a large amount of writeback cache on the device itself (and the filesystem cache; the OS will buffer writes). Since writes are constrained so specifically (depth of 1, batched, linear space, etc) the the writes queue through almost instantly to the device and it doesn't end up interfering with reads at all. If you split up the writes more the device gets busier and things go more sideways.

The bigger problem is pressure on the buffered IO subsystem, which is caused more by the reads than writes (ie; a read can block on a write to free memory). the O_DIRECT changes would help there.

dormando: To be completely specific; if you drill down really hard you will see the impact of a write, but that's not going to go away if you change the flush model. Extstore also tries to increase the amount of _time_ between writes; when a buffer is being written out, it does not buffer more writes. So once a write is complete it needs to do work for some time before submitting the next one. This gives the OS/device some time to clear out their writeback buffers.

memcached.c
dormando: I guess the callbacks are global but each connection gets a queue. All the funcs are labelled `conn_io_etc` where appropriate. ie; independent connections will all have the same queue types with the same callback funcs, but individual stack pointers for each connection.

thread.c
pkarumanchi9: Thanks for the explanation. This addresses the questions I had.

memcached.c
pkarumanchi9: Thanks, that makes sense.

storage.c
```diff
@@ -15,7 +16,453 @@
...
+    }
+
+    // We can't bail out anymore, so mc_resp owns the IO from here.
+    resp->io_pending = (io_pending_t *)p;
+
+    eio->buf = (void *)new_it;
+    p->c = c;
+
+    // We need to stack the sub-struct IO's together for submission.
+    eio->next = q->stack_ctx;
```
pkarumanchi9, line 333: can you help me understand these two lines 333-334?
Can we ever have q->stack_ctx having a value other than NULL?

pkarumanchi9: Here is my understanding -
for a given connection, if multiple read requests arrive at the same time before we could complete the request, we keep on adding the previous q->stack_ctx to eio->next and then update the q->stack_ctx to eio.

The extstore subsystem will handle all the IO's pointed to by _obj_io and the stack_ctx always points to the head of this list. Is this correct? 

How are we protected in the event that this _obj_io is being handled by the extstore subsystem and _obj_io is not meddled with?

dormando: not sure I understand the question but I'll detail the ownership process:
- a network read fills an input buffer
- commands are read out of the input buffer. requests are stacked into `q->stack_ctx`
- if end of buffer or max reqs per yied hit, switch to send result data
- if queue has stack, submit stack to extstore
- client connection processing suspends
- extstore works through all submitted obj_io's from that connection. Once the queue hits 0, the connection is signaled to resume
- worker thread resumes client connection, processes the result callbacks for each obj_io that was submitted.
- etc

So the ownership of the _whole_ queue passes between worker thread and extstore IO threads. There shouldn't be any time when both are able to access them.

pkarumanchi9: I think I found the answer to my question - client requests will not be accepted on that connection until we call redispatch_conn. So we are guaranteed not to have _obj_io getting modified while being processed by the extstore_io_thread.

dormando: Yes. That's what I said in my detail here too. This happens either during an ascii multiget (get key key key), or any time a client submits multiple requests at once (can be unrelated gets/metagets). Rest of the time it doesn't/can't batch.

There's a complication that I added to make proxy work, in that code running on a deferred queue can add new deferred IO's to its stack. IE: in proxy's case a request needs to be able to queue up a retry while it's technically running in a callback already. That's why there are two levels of callbacks (pre and post cleanup). Might be a bit hard to follow.

pkarumanchi9: Yes, the details you posted above make it much clearer to follow. (I guess you beat me to it in responding sooner :) ). I've a much better understanding of the deferred IO queue subsystem and the extstore code paths now.  Thank you!

I will have start reviewing the proxy PR now  - https://github.com/memcached/memcached/pull/716. 
I am sure I will understand the complication you mentioned above once I start reading the code.

dormando: ok, cool! Keep in mind proxy is under a lot of work at the moment :) I also frequently rebase/squash/etc the branch.

PR #719, opened 2020-09-14
wrong condition check when return item to cache

bitground:
When returning item to the cache and the items allocated exceeded cache->limit, we should free the item and not put it to the cache.

The current code will free any item that tried returned to the cache when cache->total < cache->limit which is true in most cases.

### DIFF
--- a/cache.c
+++ b/cache.c
@@ -142,7 +142,7 @@ void do_cache_free(cache_t *cache, void *ptr) {
     }
     ptr = pre;
 #endif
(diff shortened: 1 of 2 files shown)


dormando:
Good catch, thanks! I'll validate this soon.

Looks like this passed my own tests originally because I would fill the cache and then adjust it, so it would usually not free objects until hitting the cache limit once.

dormando:
out in 1.6.8

PR #720, opened 2020-09-17
Improve opening of authentication file

TomasKorbar:
Hi,
Static code analysers have been complaining about this for a
while, so i decided to improve this code a little. I realize that
there is a little danger in this particular usecase but i think
that the way i updated the code is just safer.

```
memcached-1.6.7/authfile.c:36: fs_check_call: Calling function "stat" to perform check on "file".
memcached-1.6.7/authfile.c:46: toctou: Calling function "fopen" that uses "file" after a check function. This can cause a time-of-check, time-of-use race condition.
#   34|       auth_t auth_entries[MAX_ENTRIES];
#   35|   
#   36|->     if (stat(file, &sb) == -1) {
#   37|           return AUTHFILE_MISSING;
#   38|       }
```

### DIFF
--- a/authfile.c
+++ b/authfile.c
@@ -33,22 +33,15 @@ enum authfile_ret authfile_load(const char *file) {
     char *auth_data = NULL;
     auth_t auth_entries[MAX_ENTRIES];
 
-    if (stat(file, &sb) == -1) {
-        return AUTHFILE_MISSING;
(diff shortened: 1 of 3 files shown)


dormando:
Thanks!

dormando:
out in 1.6.8

PR #721, opened 2020-09-25
Fix leaking file descriptors

TomasKorbar:
Hi @dormando,
This is another change fixing complains from static code analyser.
This change guarantees that file descriptor fd is always closed
before return from daemonize function thus, the handle never
leaks.
I'm not sure about the meaning of condition on line 81 `if (fd > STDERR_FILENO)`.
I removed it but please if it is there for some reason then correct me.

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -67,23 +67,25 @@ int daemonize(int nochdir, int noclose)
     if (noclose == 0 && (fd = open("/dev/null", O_RDWR, 0)) != -1) {
         if(dup2(fd, STDIN_FILENO) < 0) {
             perror("dup2 stdin");
-            return (-1);
+            goto err_cleanup;
         }
         if(dup2(fd, STDOUT_FILENO) < 0) {
             perror("dup2 stdout");
-            return (-1);
+            goto err_cleanup;
         }
         if(dup2(fd, STDERR_FILENO) < 0) {
             perror("dup2 stderr");
-            return (-1);
+            goto err_cleanup;
         }
 
-        if (fd > STDERR_FILENO) {
-            if(close(fd) < 0) {
-                perror("close");
-                return (-1);
-            }
+        if(close(fd) < 0) {
+            perror("close");
+            return (-1);
         }
     }
     return (0);
+
+    err_cleanup:
+        close(fd);
+        return (-1);
 }


dormando:
thanks!

I wish static analyzers found real bugs sometimes :'(

dormando:
Looks like the `STDERR_FILENO` check was added here: 2c196ed6cacfd3eea845a3b9a5758a698096d97b - not sure what it's doing offhand. I'll come back to this or you can dig more into this commit and see if you can figure it out :)

TomasKorbar:
The only thing that comes to my mind about this is that on some weird OS stderr is not open by default and that means that
new fd will have value 2 (the same as macro STDERR_FILENO). Because of that, if you close fd you would also close stderr and you would be facing crash later if you attempted to use it.
Fortunately i do not know any OS which would practice such behaviour :smile:

dormando:
Apparently I pegged the wrong commit: 98ca309daff66756330ad307a500a83d20f50751

... from a bugfix to "support nexenta", via a daemon.c file that looks to have been imported directly from NetBSD.

@TomasKorbar does NetBSD still have this file? :P nexenta would be illumos now but that would be odd.

sergey-dryabzhinsky:
NetBSD still has that weird code. Look here: https://github.com/NetBSD/src/blob/trunk/lib/libc/gen/daemon.c#L71
And no comments on it.

dormando:
I'm honestly tempted to track this down just for fun but am really limited on time at the moment.

daemon.c
```diff
@@ -67,23 +67,25 @@ int daemonize(int nochdir, int noclose)
...
-            return (-1);
+            goto err_cleanup;
         }
 
-        if (fd > STDERR_FILENO) {
-            if(close(fd) < 0) {
-                perror("close");
-                return (-1);
-            }
+        if(close(fd) < 0) {
```
sergey-dryabzhinsky, line 81: If `fd == -1` in outer block then we never reach this line.

daemon.c
```diff
@@ -67,23 +67,25 @@ int daemonize(int nochdir, int noclose)
...
+            perror("close");
+            return (-1);
         }
     }
     return (0);
+
+    err_cleanup:
+        close(fd);
```
sergey-dryabzhinsky, line 89: If `fd == -1` in previour block then we got error here.

sergey-dryabzhinsky:
What can I say? Don't trust static code analyzers too much.

What harm can be done by original code? Let's think...

1. If `/dev/null` suddenly disappear (broken lxc/udev/fs/etc...) - then no std-streams will be closed. So that?
It clearly not memcached problem.
No leaks on fd I think, cause kernel must close them on process safe exit.

2. If on some broken (non-posix) system `stderr` suddenly disapper, weirdly not available for process on startup - then opened file descriptor will not be closed and will act like `stderr` (not closed). So that?
It again not memcached problem.
No leaks on fd probably, kernel must close them on process safe exit.
It may be problem only if memcached constantly restarting at hell speed.

My thoughts - just add comment line with `@TODO: inspect probability of STDERR disappearance at process startup`.
No more weird code need to be added.

sergey-dryabzhinsky:
@dormando That code is back from 1993. Probably no one even left alive to clarify it.

Most recent daemons code just don't contain lines with `fd > STDERR_FILENO` check.

dormando:
Out in 1.6.10

PR #722, opened 2020-10-03
configure.ac: use pkg-config to retrieve openssl

ffontaine:
Use pkg-config to retrieve openssl dependencies such as -latomic or -lz

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ list to ask questions, github issues aren't seen by everyone!
 * libseccomp, (optional, experimental, linux) - enables process restrictions for
   better security. Tested only on x86-64 architectures.
 * openssl, (optional) - enables TLS support. need relatively up to date
-  version.
+  version. pkg-config is needed to find openssl dependencies (such as -lz).
 
 ## Environment
 
--- a/configure.ac
+++ b/configure.ac
@@ -406,80 +406,83 @@ AC_ARG_WITH(libssl,
 dnl ----------------------------------------------------------------------------
 dnl libssl detection.  swiped from libevent.  modified for openssl detection.
 
+PKG_PROG_PKG_CONFIG
 OPENSSL_URL=https://www.openssl.org/
 if test "x$enable_tls" = "xyes"; then
-  AC_CACHE_CHECK([for libssl directory], ac_cv_libssl_dir, [
-    saved_LIBS="$LIBS"
-    saved_LDFLAGS="$LDFLAGS"
-    saved_CPPFLAGS="$CPPFLAGS"
-    le_found=no
-    for ledir in $trylibssldir "" $prefix /usr/local ; do
+  PKG_CHECK_MODULES(OPENSSL, openssl, [LIBS="$LIBS $OPENSSL_LIBS" CFLAGS="$CFLAGS $OPENSSL_CFLAGS"], [
+    AC_CACHE_CHECK([for libssl directory], ac_cv_libssl_dir, [
+      saved_LIBS="$LIBS"
+      saved_LDFLAGS="$LDFLAGS"
+      saved_CPPFLAGS="$CPPFLAGS"
+      le_found=no
+      for ledir in $trylibssldir "" $prefix /usr/local ; do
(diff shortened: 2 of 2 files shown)


dormando:
Thanks!

@tharanga - any interest in validating?

ffontaine:
Do you have comments on this PR?

dormando:
Hey, sorry. Not yet. It probably won't make it into this next release since I let too large of a backlog in. We'll be reviewing it for the next one (in non-pandemic years, releases are typically monthly)

dormando:
@LINKIWI - Since you've been doing things with the TLS code, feel like reviewing/testing? :)

LINKIWI:
Haven't looked super closely at this yet but looks like I'm hitting an error on a slightly older (Ubuntu 18.04) distro:

```
./configure: line 5912: PKG_PROG_PKG_CONFIG: command not found
./configure: line 5915: syntax error near unexpected token `OPENSSL,'
./configure: line 5915: `  PKG_CHECK_MODULES(OPENSSL, openssl, LIBS="$LIBS $OPENSSL_LIBS" CFLAGS="$CFLAGS $OPENSSL_CFLAGS",'
```

ffontaine:
What is the result of `dpkg -l pkg-config` on your system?

LINKIWI:
Ah nevermind, I had `pkg-config` available but I tried against a clean working directory and it works as expected. Tried this against a few different environments and it seems to work pretty well.

Perhaps documentation should be updated to include the requirement for `apt-get install pkg-config` or similar for building with TLS?

ffontaine:
Sure, I can update README.md.

dormando:
looks like this got forgotten again: @LINKIWI was there anything else pending?

LINKIWI:
The change looks good to me.

ffontaine:
Do you plan on merging this PR?

ffontaine:
This PR is opened since nearly two years :(

dormando:
To be fair, it's been a crap two years :(

It's tough to keep up with backlog, but I haven't forgotten it. My buildbots all broke so I wasn't able to include this in the last round.

dormando:
rebased and merged this... hopefully it doesn't break anything :)

Thanks and sorry for ... sigh.

dormando:
Well, this does break most of my builders.. which are fairly old but still supported.

```
./configure: line 5930: PKG_PROG_PKG_CONFIG: command not found
./configure: line 5933: syntax error near unexpected token `OPENSSL,'
./configure: line 5933: `  PKG_CHECK_MODULES(OPENSSL, openssl, LIBS="$LIBS $OPENSSL_LIBS" CFLAGS="$CFLAGS $OPENSSL_CFLAGS",'
```

seems to bust even when it's not trying to configure tls.

I'll fiddle with the builders a bit and see if this is worth it or if I should revert..

dormando:
(to be clear; the problem is most of them don't seem to have pkg-config installed by default)

dormando:
Think I'm solving this on the buildbots by moving the autogen step to the VM host instead of the VM machines... think I'm okay with this since I want to test memcached as a tarball more than a development target.

With the buildbot VM's I already abstracted away a lot of the dependencies by having the host server run git/etc, then it rsyncs the files into the VM via ssh then runs configure/make/test to minimize what needs to be installed on the VM's and avoid having to install buildbot workers within the VM's...

I said VM a lot. it looks fine so far.

dormando:
okay, getting:

`configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."`

from most of my older builders. I'll look into this later unless someone else does.. I've already merged the change into next so either have to figure this out or back it out.

dormando:
Okay I fixed the workaround harder (rsync set to resolve symlinks instead of copy them, duh) and after fixing something unrelated the buildbots are all green for the first time today.

Change seems to be holding. Sorry for the two year wait...

ffontaine:
Thanks for merging this, then I'm closing this PR.

PR #725, opened 2020-10-10
Extend test CA validity to 2049

bmwiedemann:
Extend test CA validity to 2049
so that tests do not fail after 2025

Background:
As part of my work on  [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I check that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our [memcached package build tests](https://github.com/bmwiedemann/openSUSE/blob/master/packages/m/memcached/memcached.spec#L123).

Tests failed thus:
```
 #   Failed test 'client cert is verified'
 #   at ./t/ssl_verify_modes.t line 20.
 #          got: undef
 #     expected: '1'
 ./t/ssl_verify_modes.t ........
```

Below is a diff of `certtool -i`

```diff
 Root Certificate Authority
        Validity:
-               Not Before: Thu May 14 22:11:49 UTC 2020
-               Not After: Tue May 13 22:11:49 UTC 2025
+               Not Before: Sat Oct 10 17:49:30 UTC 2020
+               Not After: Fri Dec 31 17:49:32 UTC 2049
```

This PR was done while working on reproducible builds for openSUSE.


dormando:
thanks! interesting change.

can you also update t/README-TLS.md - which includes instructions on re-generating the files? A note for why they're used for so long would be best too

bmwiedemann:
I updated the README-TLS and used the commands to create certs with 500y validity. My tests showed that every test now passes in 2035 (both x86_64 and i586).

dormando:
No complaints! merged this up for the next release. Thanks for taking the time.

dormando:
released in 1.6.9!

PR #731, opened 2020-11-04
Fix long sleep in idle connections checkup thread

sergey-dryabzhinsky:
If `idle_timeout` option is setup to high value, like `610s`, then service can't be gracefuly stopped in short time.
Most init-systems will kill it with <s>fire</s>`SIGKILL` after short wait-time.

As long as `idle_timeout` is set in seconds, we should wait a little and check if thread break flag is set.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -361,9 +361,10 @@ static void *conn_timeout_thread(void *arg) {
             fprintf(stderr,
                     "idle timeout thread finished pass, sleeping for %ds\n",
                     sleep_time);
-        /* Sleep time may be too long, so split it by 1sec to exit from service normaly */
+        /* Sleep time may be too long, so split it by 0.1 sec to check exit flag and shutdown service gracefuly */
+        sleep_time *= 10;
         while (sleep_time) {
-            usleep((useconds_t) 1000000);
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! Think this was one of the TODO's from back when I shipped the restat feature, sorry. There're some similar problems, ie; waits for the item crawler to finish which can take a while if there are a lot of items.

I might rewire this a bit.

sergey-dryabzhinsky:
@dormando Thanks.
I was stumble upon it just searching what hung my home PC restart process.
And VM/containers at work.

dormando:
Hey,

I ended up fixing this here: https://github.com/memcached/memcached/commit/2896cc6c49c2bcd064e00da8bdf0dcaf1a481e73 - instead of sleep I use a cond_timedwait to allow interrupting the sleep. This is better than really short sleep loops since the frequent wakeups are problems on some systems.

sergey-dryabzhinsky:
@dormando Thanks! That code is much clearer now.
If all works as expected I'll close PR.

sergey-dryabzhinsky:
All working fine with that commit.
Close PR.
Hope `next` will be out soon.

PR #734, opened 2020-11-07
meta: allow base64'ed binary keys with 'b' flag

dormando:
ie: ms [key] b
if 'k' flag is given and key is binary, returns as binary encoded.

key is stored in raw binary; ie takes up less storage space. Instead of storing a large UUID as a 40+ character string, you can store as a 24byte or whatever value.

TODO:
- [x] Add 'b' option to `ME` for binary keys
- [x] adding 'b' to response when 'k' is requested is the right thing to do?
- [x] re-audit codebase to ensure binary keys don't leak anywhere. This should've been done a long time ago.
- [x] check against proxy branch: the inline decoding of the key might make things more difficult there? but this only happens when the process_m stuff is actually run.
- [x] doc/protocol.txt updates
- [ ] properly calculate remaining wbuf length in META_KEY()
- [ ] handle key length check in `_meta_preparse` to account for b64 encodings of 250 raw byte keys being longer than 250 bytes.

### DIFF
--- a/proto_text.c
+++ b/proto_text.c
@@ -7,6 +7,7 @@
 #include "proto_text.h"
 #include "authfile.h"
 #include "storage.h"
+#include "base64.h"
 #ifdef TLS
 #include "tls.h"
 #endif
@@ -744,8 +745,7 @@ static void process_stat(conn *c, token_t *tokens, const size_t ntokens) {
     }
 }
 
-
-
+// TODO: if 'b' after "ME key" decode the key.
 // slow snprintf for debugging purposes.
 static void process_meta_command(conn *c, token_t *tokens, const size_t ntokens) {
     assert(c != NULL);
@@ -808,6 +808,7 @@ struct _meta_flags {
     unsigned int no_reply :1;
     unsigned int has_cas :1;
     unsigned int new_ttl :1;
+    unsigned int key_binary:1;
     char mode; // single character mode switch, common to ms/ma
     rel_time_t exptime;
     rel_time_t autoviv_exptime;
@@ -822,6 +823,7 @@ struct _meta_flags {
 static int _meta_flag_preparse(token_t *tokens, const size_t ntokens,
         struct _meta_flags *of, char **errstr) {
     unsigned int i;
+    size_t ret;
     int32_t tmp_int;
     uint8_t seen[127] = {0};
     // Start just past the key token. Look at first character of each token.
@@ -834,6 +836,19 @@ static int _meta_flag_preparse(token_t *tokens, const size_t ntokens,
         }
         seen[o] = 1;
         switch (o) {
+            // base64 decode the key in-place, as the binary should always be
+            // shorter and the conversion code buffers bytes.
+            case 'b':
+                ret = base64_decode((unsigned char *)tokens[KEY_TOKEN].value, tokens[KEY_TOKEN].length,
+                            (unsigned char *)tokens[KEY_TOKEN].value, tokens[KEY_TOKEN].length);
+                if (ret == 0) {
+                    // Failed to decode
+                    *errstr = "CLIENT_ERROR error decoding key";
+                    of->has_error = 1;
+                }
+                tokens[KEY_TOKEN].length = ret;
+                of->key_binary = 1;
+                break;
             /* Negative exptimes can underflow and end up immortal. realtime() will
                immediately expire values that are greater than REALTIME_MAXDELTA, but less
                than process_started, so lets aim for that. */
@@ -961,6 +976,20 @@ static int _meta_flag_preparse(token_t *tokens, const size_t ntokens,
     p += 2; \
 }
 
+// TODO: calc bytes remaining in buffer
+#define META_KEY(p, key, nkey, bin) { \
+    META_CHAR(p, 'k'); \
+    if (!bin) { \
+        memcpy(p, key, nkey); \
+        p += nkey; \
+    } else { \
+        p += base64_encode((unsigned char *) key, nkey, (unsigned char *)p, WRITE_BUFFER_SIZE); \
+        *p = ' '; \
+        *(p+1) = 'b'; \
+        p += 2; \
+    } \
+}
+
 static void process_mget_command(conn *c, token_t *tokens, const size_t ntokens) {
     char *key;
     size_t nkey;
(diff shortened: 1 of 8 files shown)


dormando:
thinking through if using an ITEM flag is necessary:
- it should be impossible to fetch a binary key without first specifying the `b` flag. you might be able to do it by luck so long as you never hit spaces/newlines though?
- all other cases where keys are printed are either URI encoded or similar (check cachedump?)
- so printback only needs to be adjusted if the request had `b`

probably still want to send `b` back in the response to make clients/proxies easier to code.

still need to audit the codebase for all ITEM_key() calls.

dormando:
re: proxy, I think it's going to be fine as-is for now. the proxy's hook into the parser will be higher level (ie; `process_command()`), which is before key encoding/decoding. if it wants to internally call a command on a request it'll have to do it with a copy of the request string regardless since tokenize_command() mangles it.

dormando:
ME now uses `b`. it re-encodes the key from the internal memory as an added safety check, since it was doing that before.

just need to do the `ITEM_key()` audit and docs now.

dormando:
All done. @smukil - if you wanted to take a look before I merged.

proto_text.c
```diff
@@ -961,6 +993,20 @@ static int _meta_flag_preparse(token_t *tokens, const size_t ntokens,
     p += 2; \
 }
 
+// TODO: calc bytes remaining in buffer
```
dormando, line 996: ah, should do this before merge.

dormando:
@smukil - going to hold this off for the next release (+ hopefully someone can look at it).

need to make a high level decision on if b64 encoded keys still have a 250 byte wire limit, or if it won't pose any issues to ensure the *decoded* limit is 250 bytes. IE: clients will need to ensure they check length limit before encoding, but reserve buffer space for larger, etc.

250 byte raw binary key feels like a lot though.

dormando:
Aiming to get this wrapped up and released this week.

proto_text.c
smukil: Isn't this already implicitly calculated? since 'p' itself is the 'wbuf' ? And 'wbytes' will later be adjusted based as 'p = resp->wbuf'

smukil:
> handle key length check in _meta_preparse to account for b64 encodings of 250 raw byte keys being longer than 250 bytes.
I'm assuming you mean "b64 decodings" instead of "encodings".

How should we handle this? Should we reject such keys? Or is it okay to consider them valid since the value stored is still <= 250 ?

smukil:
A high level question. Using the tesuto example:
```
my $tesuto = "44OG44K544OI";
print $sock "ms $tesuto S2 b\r\npo\r\n";
```

Reading the code, is a "get tesuto" possible? I'm assuming not, since I don't see any b64 encoding/decoding in the regular get code path.

smukil:
Another high level question, how would the `LRU metadump` dump these keys? I see that the `cachedump` code skips over them.

dormando:
> I'm assuming you mean "b64 decodings" instead of "encodings".

I meant as 'encoded values being longer than 250 bytes', so decodings would be less than 250, but I think you understood :)
 
> How should we handle this? Should we reject such keys? Or is it okay to consider them valid since the value stored is still <= 250 ?

That's the question! It's simplest to reject encoded keys that end up longer than 250 bytes but it's an artificial limit in key length.

dormando:
> A high level question. Using the tesuto example:
> 
> ```
> my $tesuto = "44OG44K544OI";
> print $sock "ms $tesuto S2 b\r\npo\r\n";
> ```
> 
> Reading the code, is a "get tesuto" possible? I'm assuming not, since I don't see any b64 encoding/decoding in the regular get code path.

No it's not possible outside of metaget. Technically UTF8 keys always worked, unless the encodings ended up with restricted characters in them :) In any case it's useful here; it's impossible to use metaset to drop in a binary key and then accidentally fetch it using a system that can't display it properly.

> 
> If this is the case, we'd need to clearly call it out in the docs that using "ms" with the 'b' flag means it is only retrievable through "mget".

Will add more emphasis to this, yes. This is a meta only feature; the classical ascii commands aren't extensible in this way. I'd have to add a whole new set of commands to support it, which is why I went with meta with positional flags.

dormando:
> Another high level question, how would the `LRU metadump` dump these keys? I see that the `cachedump` code skips over them.

metadump URI encodes keys, as I wrote it to support binary keys. Everything internally had been binary clean since the binprot was added, aside from cachedump apparently.

why did I do URI encoding instead of base64 for metadump? - because typical ascii keys would stay human readable and URI encoding is a standard library in just about all modern languages. It was a very user friendly option. Should dump encoding have been a set of options? Yeah, probably. I don't have a time machine and nobody brought it up when I wrote the feature initially :)

proto_text.c
dormando: Think I meant to actually use the bytes remaining to avoid potentially overflowing.

smukil:
> > Another high level question, how would the `LRU metadump` dump these keys? I see that the `cachedump` code skips over them.
> 
> metadump URI encodes keys, as I wrote it to support binary keys. Everything internally had been binary clean since the binprot was added, aside from cachedump apparently.
> 

So just to be clear, if we have b64 encoded keys in memory, a metadump will URI encode said keys and print it out?

Also, a secondary patch-unrelated question. How do I run a single test from the test suite? Can I do a "make test <test ID>" or something?

dormando:
> So just to be clear, if we have b64 encoded keys in memory, a metadump will URI encode said keys and print it out?

```
ms 44OG44K544OI S2 b
hi
OK 
lru_crawler metadump all
key=%E3%83%86%E3%82%B9%E3%83%88 exp=-1 la=1607016606 cas=1 fetch=no cls=1 size=70
```
URI encoding means all non-URI printable characters get %FF'd.

> Also, a secondary patch-unrelated question. How do I run a single test from the test suite? Can I do a "make test " or something?

`prove t/lru-crawler.t`

dormando:
looks like I'm going with a non-modifying parser for the proxy code so I might need to re-evaluate how this was written. I want to reuse the `meta_preparse` function but can't if it's modifying the request string.

shouldn't be too bad but it does need a change.

smukil:
> looks like I'm going with a non-modifying parser for the proxy code so I might need to re-evaluate how this was written. I want to reuse the `meta_preparse` function but can't if it's modifying the request string.
> 
> shouldn't be too bad but it does need a change.

I'll wait for this change too. It looks like it may require intermediate allocations, unless I'm missing something.

smukil:
> > I'm assuming you mean "b64 decodings" instead of "encodings".
> 
> I meant as 'encoded values being longer than 250 bytes', so decodings would be less than 250, but I think you understood :)
> 
> > How should we handle this? Should we reject such keys? Or is it okay to consider them valid since the value stored is still <= 250 ?
> 
> That's the question! It's simplest to reject encoded keys that end up longer than 250 bytes but it's an artificial limit in key length.

After thinking about it, IMO rejecting such keys would be reasonable.

dormando:
Think I'm going to take this as-is after double checking the TODO items. I'll feel better about the code if the decoded key only exists in function stack memory, rather than decoding from the tokenized data; but I don't have to fix that now.

I'm trending toward having a const input string. that will require a few other changes due to the tokenizer modifying the string. When that comes around I'll move the decode stage. A few other chunks of redundant meta function header code can be abstracted at the same time.

dormando:
Out in 1.6.10

PR #735, opened 2020-11-09
item crawler hash table walk mode

dormando:
code compiles but not tested/completed yet.

specifying 'hash' instead of 'all' will make the LRU crawler iterate
over every bucket in the hash table once, instead of what's in the LRU.
This also doesn't suffer from missing items because of LRU reordering
or high lock contention.

### DIFF
--- a/assoc.c
+++ b/assoc.c
@@ -309,25 +309,28 @@ void *assoc_get_iterator(void) {
     return iter;
 }
 
-item *assoc_iterate(void *iterp) {
+bool assoc_iterate(void *iterp, item **it) {
     struct assoc_iterator *iter = (struct assoc_iterator *) iterp;
+    *it = NULL;
     // - if locked bucket and next, update next and return
     if (iter->bucket_locked) {
         if (iter->next != NULL) {
             iter->it = iter->next;
             iter->next = iter->it->h_next;
-            return iter->it;
+            *it = iter->it;
         } else {
             // unlock previous bucket, if any
             item_unlock(iter->bucket);
             // iterate the bucket post since it starts at 0.
             iter->bucket++;
             iter->bucket_locked = false;
+            *it = NULL;
         }
+        return true;
     }
 
     // - loop until we hit the end or find something.
-    while (iter->bucket != hashsize(hashpower)) {
+    if (iter->bucket != hashsize(hashpower)) {
         // - lock next bucket
         item_lock(iter->bucket);
         iter->bucket_locked = true;
@@ -336,16 +339,18 @@ item *assoc_iterate(void *iterp) {
         if (iter->it != NULL) {
             // - set it, next and return
             iter->next = iter->it->h_next;
-            return iter->it;
+            *it = iter->it;
         } else {
             // - nothing found in this bucket, try next.
             item_unlock(iter->bucket);
             iter->bucket_locked = false;
             iter->bucket++;
         }
+    } else {
+        return false;
     }
 
-    return NULL;
+    return true;
 }
 
 void assoc_iterate_final(void *iterp) {
--- a/assoc.c
+++ b/assoc.c
@@ -292,3 +292,67 @@ void stop_assoc_maintenance_thread() {
     pthread_join(maintenance_tid, NULL);
 }
 
+struct assoc_iterator {
+    unsigned int bucket;
+    bool bucket_locked;
+    item *it;
+    item *next;
+};
+
+void *assoc_get_iterator(void) {
+    struct assoc_iterator *iter = calloc(1, sizeof(struct assoc_iterator));
+    if (iter == NULL) {
+        return NULL;
+    }
+    // this will hang the caller while a hash table expansion is running.
+    mutex_lock(&maintenance_lock);
+    return iter;
+}
+
+item *assoc_iterate(void *iterp) {
+    struct assoc_iterator *iter = (struct assoc_iterator *) iterp;
+    // - if locked bucket and next, update next and return
+    if (iter->bucket_locked) {
+        if (iter->next != NULL) {
+            iter->it = iter->next;
+            iter->next = iter->it->h_next;
+            return iter->it;
+        } else {
+            // unlock previous bucket, if any
+            item_unlock(iter->bucket);
+            // iterate the bucket post since it starts at 0.
+            iter->bucket++;
+            iter->bucket_locked = false;
+        }
+    }
+
+    // - loop until we hit the end or find something.
+    while (iter->bucket != hashsize(hashpower)) {
+        // - lock next bucket
+        item_lock(iter->bucket);
+        iter->bucket_locked = true;
+        // - only check the primary hash table since expand is blocked.
+        iter->it = primary_hashtable[iter->bucket];
+        if (iter->it != NULL) {
+            // - set it, next and return
+            iter->next = iter->it->h_next;
+            return iter->it;
+        } else {
+            // - nothing found in this bucket, try next.
+            item_unlock(iter->bucket);
+            iter->bucket_locked = false;
+            iter->bucket++;
+        }
+    }
+
+    return NULL;
+}
+
+void assoc_iterate_final(void *iterp) {
+    struct assoc_iterator *iter = (struct assoc_iterator *) iterp;
+    if (iter->bucket_locked) {
+        item_unlock(iter->bucket);
+    }
+    mutex_unlock(&maintenance_lock);
+    free(iter);
+}
--- a/crawler.c
+++ b/crawler.c
@@ -349,6 +349,58 @@ static void lru_crawler_class_done(int i) {
         active_crawler_mod.mod->doneclass(&active_crawler_mod, i);
 }
 
+static void item_crawl_hash(void) {
+    // get iterator from assoc. can hang for a long time.
+    // - blocks hash expansion
+    void *iter = assoc_get_iterator();
+    int crawls_persleep = settings.crawls_persleep;
+    item *it = NULL;
+
+    // loop while iterator returns something
+    // - iterator func handles bucket-walking
+    // - iterator returns with bucket locked.
+    while ((it = assoc_iterate(iter)) != NULL) {
+        /* Get memory from bipbuf, if client has no space, flush. */
+        if (active_crawler_mod.c.c != NULL) {
+            int ret = lru_crawler_client_getbuf(&active_crawler_mod.c);
+            if (ret != 0) {
+                // fail out and finalize.
+                break;
+            }
+        } else if (active_crawler_mod.mod->needs_client) {
+            // fail out and finalize.
+            break;
+        }
+
+        // double check that the item isn't in a transitional state.
+        if (refcount_incr(it) < 2) {
+            refcount_decr(it);
+            continue;
+        }
+
+        // FIXME: missing hv and i are fine for metadump eval, but not fine
+        // for expire eval.
+        active_crawler_mod.mod->eval(&active_crawler_mod, it, 0, 0);
+
+        // - sleep bits from orig loop
+        if (crawls_persleep-- <= 0 && settings.lru_crawler_sleep) {
+            pthread_mutex_unlock(&lru_crawler_lock);
+            usleep(settings.lru_crawler_sleep);
+            pthread_mutex_lock(&lru_crawler_lock);
+            crawls_persleep = settings.crawls_persleep;
+        } else if (!settings.lru_crawler_sleep) {
+            // TODO: only cycle lock every N?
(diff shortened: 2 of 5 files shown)


smukil:
I'll try out this patch and review it by EOD.

dormando:
might want to wait a few hours :) I haven't even tried running it yet, but will do shortly.

dormando:
holy crap it seems to work :P adding some automated tests and auditing the code again.

assoc.c
```diff
@@ -292,3 +292,73 @@ void stop_assoc_maintenance_thread() {
...
+            // unlock previous bucket, if any
+            item_unlock(iter->bucket);
+            // iterate the bucket post since it starts at 0.
+            iter->bucket++;
+            iter->bucket_locked = false;
+        }
+    }
+
+    // - loop until we hit the end or find something.
+    while (1) {
```
dormando, line 331: think I'll change the loop to while (iter->bucket != hashsize(hashpower).

crawler.c
```diff
@@ -349,6 +349,59 @@ static void lru_crawler_class_done(int i) {
...
+    void *iter = assoc_get_iterator();
+    int crawls_persleep = settings.crawls_persleep;
+    item *it = NULL;
+
+    // loop while iterator returns something
+    // - let the iterator func handle sub-looping
+    // - iterator returns with bucket locked.
+    while ((it = assoc_iterate(iter)) != NULL) {
+        // - getbuf from bipbuf stuff
+        /* Get memory from bipbuf, if client has no space, flush. */
```
dormando, line 364: redundant comment

crawler.c
```diff
@@ -645,13 +710,18 @@ enum crawler_result_type lru_crawler_crawl(char *slabs, const enum crawler_run_t
...
+    bool hash_crawl = false;
 
     /* FIXME: I added this while debugging. Don't think it's needed? */
     memset(tocrawl, 0, sizeof(uint8_t) * POWER_LARGEST);
     if (strcmp(slabs, "all") == 0) {
         for (sid = 0; sid < POWER_LARGEST; sid++) {
             tocrawl[sid] = 1;
         }
+    } else if (strcmp(slabs, "hash") == 0) {
+        // TODO: Bounce if not metadump. expire eval is too closely tied to
```
dormando, line 722: need to actually do this. making expire eval actually work will require more refactoring I don't really need to do right now; hash table walking for an expiration check is pointless.

dormando:
@smukil should be ready for you now. did some minor cleanups + finished the rest of the TODO's and added a test. so far as I can tell it seems to work fine.

smukil:
I'm yet to do a full code walkthrough, but while testing it out locally, I found something interesting. I started `memcached` with this patch, filled it until it was full and then stopped all traffic to it. I then did a "lru_crawler metadump hash" multiple times and counted the number of keys returned each time.

While I expected them to be the same number of keys each time, they seemed to always return different numbers.

`

STAT curr_items 699008
STAT total_items 13152534


sailesh@sailesh-VirtualBox:~/work/memcached$ echo 'lru_crawler metadump hash' | nc localhost 11211 > all_keys

sailesh@sailesh-VirtualBox:~/work/memcached$ echo 'lru_crawler metadump hash' | nc localhost 11211 > all_keys_2

sailesh@sailesh-VirtualBox:~/work/memcached$ echo 'lru_crawler metadump hash' | nc localhost 11211 > all_keys_3

sailesh@sailesh-VirtualBox:~/work/memcached$ wc -l all_keys_3 
698975 all_keys_3

sailesh@sailesh-VirtualBox:~/work/memcached$ wc -l all_keys_2 
698980 all_keys_2

sailesh@sailesh-VirtualBox:~/work/memcached$ wc -l all_keys
698993 all_keys
`

While there are only '699008' keys, I also noticed it returning 699027 keys once.

I'm going to continue looking at it, but I just wanted to point this out quickly.

crawler.c
```diff
@@ -349,6 +349,58 @@ static void lru_crawler_class_done(int i) {
...
+            refcount_decr(it);
+            continue;
+        }
+
+        // FIXME: missing hv and i are fine for metadump eval, but not fine
+        // for expire eval.
+        active_crawler_mod.mod->eval(&active_crawler_mod, it, 0, 0);
+
+        // - sleep bits from orig loop
+        if (crawls_persleep-- <= 0 && settings.lru_crawler_sleep) {
```
smukil, line 386: Wouldn't this hold up a bucket's lock for longer than necessary?

If we need to adhere to the sleep, maybe we can do so between buckets instead of 'crawls_persleep' for the hash table walk? That would require some changes to assoc_iterate() to return identifiably between buckets.

crawler.c
```diff
@@ -624,12 +681,25 @@ int lru_crawler_start(uint8_t *ids, uint32_t remaining,
...
-    for (int sid = POWER_SMALLEST; sid < POWER_LARGEST; sid++) {
-        if (ids[sid])
-            starts += do_lru_crawler_start(sid, remaining);
+    if (ids == NULL) {
+        /* NULL ids means to walk the hash table instead. */
+        starts = 1;
+        /* FIXME: hack to signal hash mode to the crawler thread.
+         * Something more clear would be nice.
+         */
+        crawler_count = -1;
```
smukil, line 690: Would setting this to '-1' affect any pending crawl requests?

For example: Could crawler_count be something greater than 0 when we set it to -1 here? If it could, then that would affect the crawlers yet to run (i.e. the ones linked to 'heads' and 'tails' in item.c)

dormando: normal crawler shouldn't be able to start if it's already running. EXPIRE is special cased to allow restarts while it's already running. I might want to specifically test for this.

smukil: Trying to think of an example:

If one client sends "lru_crawler metadump all", and while that's just starting to be serviced (meaning crawler_count is definitely > 0);
another client could send in "lru_crawler metadump hash", which takes the 'lru_crawler_lock' when the item crawler thread is sleeping (because of 'crawls_persleep'), and set crawler_count = -1.

When the item crawler thread wakes up again, it would send the crawler_count to a negative value and all book keeping from then on is incorrect.

crawler.c
dormando: good catch! that one totally slipped. You're probably right but I'll give it a think. The iterator never returns without a lock, but I could hand it a between-unlocks callback during init without making the interface too gross.

crawler.c
dormando: need to do a wider refactor to make expire work, and to keep the interface generic for modularization; when doing that I'll definitely need a different signal that crawler_count=-1.

dormando: look at the is_running tests in lru_crawler_start()

crawler.c
```diff
@@ -604,6 +655,12 @@ int lru_crawler_start(uint8_t *ids, uint32_t remaining,
         return -1;
     }
 
+    /* hash table walk only supported with metadump for now. */
+    if (type != CRAWLER_METADUMP && ids == NULL) {
```
smukil, line 659: nit: We can catch this much earlier at lru_crawler_crawl() at L726 without needing to take the lock.

crawler.c
smukil: Ah I missed that. You're right, it only affects EXPIRE. We'd need to still fix that to make sure nothing breaks during regular LRU maintenance , right?

dormando: shouldn't have to; there's another rule ensuring only one active module at a time. if you want to start metadump it has to finish and pause expire.. there's some cruddy code in there that blocks expire from restarting if you're issuing manual commands and they're bouncing.

crawler.c
dormando: lru_crawler_crawl() is "externally" facing, so future-proofing puts this check a bit deeper, though it ends up giving a non user friendly error.

crawler.c
dormando: I'm not really thrilled with the way that ended up but I didn't want to invest a ton of time rewiring it to be a stack/queue at the time, so I ended up just putting the timed block code in to make it work.

dormando: (it also special cases so hash is metadump only, as I have a few other things to refactor before expire will work with a hash walk :/)

smukil: I think I'm missing something about expire and hash walk conflicting with each other. Am I right in understanding that a "lru_crawler metadump hash" would not break an already running AUTOEXPIRE, but stop it from running to completion, potentially leaving expired items around for longer than they need to be?

And if yes, would you be submitting further changes to fix that in this patch? Or a follow on patch?

dormando: It waits for the autoexpire to finish, preventing new sub-LRU autoexpires from firing. only one type of LRU crawler can run at a time so metadump always blocks the expire check; it's always been like that.

The internal autoexpire routine uses a histogram to ensure it's only running when it thinks it can recover a lot of data, and when it does it targets specific LRU's that it think it'll get data from, so typically it probably doesn't run much.

Don't think I'm going to be refactoring the autoexpire stuff as part of this patch series. There's no benefit to having autoexpire run on hash mode so it can be done later. I'm not really liking the code structure so having more time to think about it would help too. If you have more immediate needs I can do a different PR for it sooner than later :)

dormando: to be more specific, if you try to run metadump while expire is running, the metadump will return an error (crawler already running). internally there's a hack that says "someone is trying to run a manual command; don't start any automated internal commands for another minute), to allow you to re-try the metadump command once every N seconds and it'll eventually guarantee to work.

I'd prefer to just queue them, probably. running multiple types at once is no good for a lot of reasons.

smukil: Ah this makes it a lot clearer. I re-read the code after your explanation and it looks about right. Other than your code restructuring changes, I think this part looks fine functionally.

smukil:
I'm going to try it out an a much larger instance and see how it fares. I'll also try to restructure the assoc_iterator to return without a lock between buckets to sleep and compare results. I'll get back with our results and the urgency of this patch after that.

dormando:
I wrote a callback for iter locally this morning but it feels off. going to give a few more minutes to think about it.

think you normally run with the sleep set to 0? or you should or else metadump will take a long time :)

dormando:
okay well this doesn't feel too disgusting. it passes the test, but I think I also want to expand the test to verify all of the expected key names were actually seen, to verify it's not double returning or missing anything.

assoc.c
```diff
@@ -292,3 +292,72 @@ void stop_assoc_maintenance_thread() {
...
+            iter->bucket_locked = false;
+            *it = NULL;
+        }
+        return true;
+    }
+
+    // - loop until we hit the end or find something.
+    if (iter->bucket != hashsize(hashpower)) {
+        // - lock next bucket
+        item_lock(iter->bucket);
```
tharanga, line 335: We used to do `item_trylock` instead of `item_lock`. Did we change that to increase accuracy?

dormando: LRU crawler does `item_trylock` instead of `item_lock` because the lock order is inverted: item needs to be gotten before LRU or else it can deadlock. So if you fail to get item you can't wait for it.

Since the LRU isn't involved in a hash table walk we can do a full wait. it's very similar to doing a straight GET request.

tharanga: Cool. I believe the difference (between GET) is we keep the lock until we finish walking the linked list.

dormando: Correct. Have to hold the lock through the whole bucket, but depth will rarely be higher than 1, not more than 2-3 except for some extreme cases.

dormando: I mean, to be specific GET does hold the lock until it finds the right item, which could be at top or bottom of the bucket :)

dormando:
Any luck? I want to merge this + binkeys and cut a release this week. I still need to add the other test but that's quick.

smukil:
> Any luck? I want to merge this + binkeys and cut a release this week. I still need to add the other test but that's quick.

I'll have a final pass by EOD on both.

smukil:
I was unable to test this with a really large extstore instance due to some other infra issues on my side. But I should have it working in a day or 2 hopefully. If you're looking to get this in soon, we can merge it since I don't see any other issues with the patch from my local testing + code review.

I can submit bug fixes if I find any from my extstore testing.

smukil (approved):

dormando:
Thanks! I have some test servers I can load up and run a test on so I'll do that. It won't be the same as your production data but for mc that doesn't typically matter.

Targeting wednesday fora release; not in a rush but want to take next week off, so good to leave a few days for immediate bug reports :)

dormando:
rebased + extra test. seems fine. leaving the branch here un-squashed but going to squash and merge to next locally.

dormando:
released in 1.6.9!

PR #736, opened 2020-11-11
extstore: use fcntl locking on disk file

dormando:
if you accidentally start memcached with the same options twice,
extstore is initiated before the listener sockets and will happily
truncate its own file.

So this avoids that. Keep in mind any other process can still wipe the
file clean!

### DIFF
--- a/extstore.c
+++ b/extstore.c
@@ -253,15 +253,32 @@ void *extstore_init(struct extstore_conf_file *fh, struct extstore_conf *cf,
     e->page_size = cf->page_size;
     uint64_t temp_page_count = 0;
     for (f = fh; f != NULL; f = f->next) {
-        f->fd = open(f->file, O_RDWR | O_CREAT | O_TRUNC, 0644);
+        f->fd = open(f->file, O_RDWR | O_CREAT, 0644);
         if (f->fd < 0) {
             *res = EXTSTORE_INIT_OPEN_FAIL;
 #ifdef EXTSTORE_DEBUG
-            perror("open");
+            perror("extstore open");
 #endif
             free(e);
             return NULL;
         }
+        // use an fcntl lock to help avoid double starting.
+        struct flock lock;
+        lock.l_type = F_WRLCK;
+        lock.l_start = 0;
+        lock.l_whence = SEEK_SET;
+        lock.l_len = 0;
+        if (fcntl(f->fd, F_SETLK, &lock) < 0) {
+            *res = EXTSTORE_INIT_OPEN_FAIL;
+            free(e);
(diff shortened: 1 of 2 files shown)


extstore.c
```diff
@@ -253,15 +253,32 @@ void *extstore_init(struct extstore_conf_file *fh, struct extstore_conf *cf,
...
             free(e);
             return NULL;
         }
+        // use an fcntl lock to help avoid double starting.
+        struct flock lock;
+        lock.l_type = F_WRLCK;
+        lock.l_start = 0;
+        lock.l_whence = SEEK_SET;
+        lock.l_len = 0;
+        if (fcntl(f->fd, F_SETLK, &lock) < 0) {
```
smukil, line 271: lockf() seems to do the same thing in a more concise way, but if you're worried about portability, this would be better.

Also, reading the docs, fcntl() seems preferrable to flock() for our case here.

extstore.c
```diff
@@ -253,15 +253,32 @@ void *extstore_init(struct extstore_conf_file *fh, struct extstore_conf *cf,
...
+            *res = EXTSTORE_INIT_OPEN_FAIL;
```
smukil, line 272: nit: Not important, but maybe a "goto failure" block.

smukil (approved):
LGTM. Couple of minor comments.

pkarumanchi9 (approved):

extstore.c
dormando: portability is a concern yes. this was the oldest method I could find :)

extstore.c
dormando: that crossed my mind while doing this; but since there's another big PR open on extstore currently I want to keep the line changes to a minimum to save time on rebase.

extstore.c
```diff
@@ -253,15 +253,32 @@ void *extstore_init(struct extstore_conf_file *fh, struct extstore_conf *cf,
...
+            free(e);
```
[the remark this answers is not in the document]
dormando: looks like I dropped the close back when this became a loop. Will need to circle back after the other outstanding PR is dealt with and use an error section to properly loop close all FD's. Since it kills the application any freeing/closing is just to satisfy people's static analyzers anyway.

dormando:
none of you said "needs a test!" ;)

I'll be adding a test in a bit then sending this through next I think. just had an adventure fixing a portability issue in freebsd to get the build clean first.

dormando:
added quick test, pushed to `next` for the full buildbot treatment.

dormando:
released in 1.6.9!

PR #737, opened 2020-11-15
Expose memory_file path in stats settings

LINKIWI:
This change proposes the ability to read back the wam restart memory file path in a `stats settings` command, in similar spirit as some other options like `ssl_chain_cert`. We have found this useful when querying metrics targeting only clusters with warm restart enabled (since `stats settings` output is scraped into the metrics system).

With a memory file:

```bash
$ echo "stats settings" | nc -q0 localhost 11211 | grep memory_file
STAT memory_file /tmp/restart-test
```

Without:

```bash
$ echo "stats settings" | nc -q0 localhost 11211 | grep memory_file
STAT memory_file (null)
```


dormando:
Thanks! This was on the restart TODO list I think.

dormando:
released in 1.6.9!

PR #740, opened 2020-11-16
Graceful shutdown protocol command

### LINKED ISSUE #739: Online command for graceful shutdown
Currently, the `--enable-shutdown` flag enables the use of the `shutdown` command to raise a `SIGINT` to the process: https://github.com/memcached/memcached/blob/8335dd84ca85651115d518d7e3d963de9d82e22f/proto_text.c#L2353

I'm wondering if there is any interest in exposing another text command (maybe `shutdown graceful` or just `graceful`?) to instead trigger a graceful shutdown with `SIGUSR1`.

My use case is allowing application logic to control the memcached process lifecycle with warm restart enabled. I'm happy to submit a patch if there's interest in this functionality.

LINKIWI:
- Add a new subcommand `shutdown graceful` that raises a `SIGUSR1` for graceful shutdown.
- Add some documentation on the shutdown command in `doc/protocol.txt`
- Also add a simple unit test for shutdown behavior. I realized that there's already a test defined in testapp, I can get rid of the Perl test added here in favor of updating that one if that is preferred.

Manual test:

Shutdown enabled:

```
$ echo "shutdown" | nc -q0 localhost 11211
...
Signal handled: Interrupt.
$ echo "shutdown graceful" | nc -q0 localhost 11211
...
Graceful shutdown signal handled: User defined signal 1.
Gracefully stopping
```

```
$ echo "shutdown asdfsad" | nc -q0 localhost 11211
CLIENT_ERROR invalid shutdown mode
```

Shutdown not enabled (default):

```
$ echo "shutdown" | nc -q0 localhost 11211
ERROR: shutdown not enabled
$ echo "shutdown graceful" | nc -q0 localhost 11211
ERROR: shutdown not enabled
$ echo "shutdown foo" | nc -q0 localhost 11211
ERROR: shutdown not enabled
```

Resolves #739.

### DIFF
--- a/t/shutdown.t
+++ b/t/shutdown.t
@@ -1,0 +1,39 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use Test::More;
+use FindBin qw($Bin);
+use lib "$Bin/lib";
+use MemcachedTest;
+
+my $server;
+my $sock;
+
+# Disabled shutdown (default)
+$server = new_memcached();
+$sock = $server->sock;
+print $sock "shutdown\r\n";
+is(scalar <$sock>, "ERROR: shutdown not enabled\r\n",
+    "error when shutdown is not enabled");
+$server->DESTROY();
+
+# Normal shutdown
+$server = new_memcached("-A");
+$sock = $server->sock;
+print $sock "version\r\n";
+like(scalar <$sock>, qr/VERSION/, "server is alive");
+print $sock "shutdown\r\n";
+print $sock "version\r\n";
+is(scalar <$sock>, undef, "server has been shut down");
+
+# Graceful shutdown
+$server = new_memcached("-A");
+$sock = $server->sock;
+print $sock "version\r\n";
+like(scalar <$sock>, qr/VERSION/, "server is alive");
+print $sock "shutdown_graceful\r\n";
+print $sock "version\r\n";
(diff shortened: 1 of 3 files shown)


dormando:
Thanks! I'd like "shutdown graceful" better - doesn't add another endpoint to the top level parser and lines up more with how other commands work.

dormando:
Perfect, thanks!

t/shutdown.t
```diff
@@ -0,0 +1,46 @@
...
+my $server;
+my $sock;
+
+# Disabled shutdown (default)
+$server = new_memcached();
+$sock = $server->sock;
+print $sock "shutdown\r\n";
+is(scalar <$sock>, "ERROR: shutdown not enabled\r\n",
+    "error when shutdown is not enabled");
+$server->DESTROY();
```
dormando, line 19: I'm going to edit these out, but fwiw putting the tests in blocks and `my`ing is better. you shouldn't be calling DESTROY(); directly

dormando:
made a couple edits and pushed to `next`, thanks again!

dormando:
the buildbots mostly failed on the shutdown test. I didn't think that one through too hard during review. I've made some more edits directly upstream. It now directly tests for the pid exiting, since otherwise it's a race + fails really weirdly in perl and just exits the test.

dormando:
released in 1.6.9!

Issue #743, opened 2020-11-24
memcached 1.6.9 - HWCAP_CRC32 undefined - arm64 build with gcc-4.8


sergey-dryabzhinsky:
**Describe the bug**
memcached 1.6.9 arm64 build on Ubuntu 14.04 - produce error `HWCAP_CRC32 undefined`.

**To Reproduce**
Try to build memcached on Launchpad for Ubuntu 14.04 for arm64

**System Information**
 - OS/Distro: Ubuntu
 - Version of OS/distro: 14.04
 - Version of memcached: 1.6.9
 - Hardware detail: arm64

**Detail (please include!)**
```
gcc -std=gnu99 -DHAVE_CONFIG_H -I.  -DNDEBUG -D_FORTIFY_SOURCE=2  -g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -O2 -mcpu=cortex-a53 -mtune=cortex-a53 -pthread -pthread -Wall -Werror -pedantic -Wmissing-prototypes -Wmissing-declarations -Wredundant-decls -c -o memcached-crc32c.o `test -f 'crc32c.c' || echo './'`crc32c.c
crc32c.c: In function 'crc32c_init':
crc32c.c:336:16: error: 'HWCAP_CRC32' undeclared (first use in this function)
     if (auxv & HWCAP_CRC32)
                ^
crc32c.c:336:16: note: each undeclared identifier is reported only once for each function it appears in
make[3]: *** [memcached-crc32c.o] Error 1
```

sergey-dryabzhinsky:
Version 1.6.8 builds fine.

Seems like in old libc/gcc constant `HWCAP_CRC32` not defined.
So check it.

PR: #744

dormando:
Out in 1.6.10

PR #744, opened 2020-11-24
Issue 743 quick fix hw crc32 capability check on arm64

sergey-dryabzhinsky:
For #743 

Just check if `HWCAP_CRC32` ever defined after including `sys/auxv.h`


dormando:
Thanks! I'll see if it's worth adding that define if missing too.

sergey-dryabzhinsky:
I think it should not be custom defined since gcc-4.8 has partial arm64 support.
AFAIK hw crc32 available on ARM8.1.
So it's fine if memcached safely builds w/o hw crc32 on such old distros.

sergey-dryabzhinsky:
But maybe it should be checked by autotools if possible.

dormando:
Note to self: looks like the user's repo is gone. need to recover the change manually.

dormando:
Out in 1.6.10

PR #746, opened 2020-12-06
Check github actions weekly with dependabot

jlosito:
This allows dependabot to check github actions which this project uses and submit PRs with version bumps.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


dormando:
Thanks! I'm not really clear on what this would be updating though? I don't look at the actions much and the only one we have is that test kickoff for PRs.

jlosito:
I believe the only action which this project currently uses is the checkout action linked below. This action looks to be on the latest version, so dependabot wouldn't be changing anything at this moment. Dependabot would only be monitoring this single checkout action and submit a PR if a v3.x.x were to ever come out.

https://github.com/memcached/memcached/blob/master/.github/workflows/ci.yml#L9

This change might not be worth the hassle.

dormando:
Think we're good yeah. Sorry for the long lag here.

PR #748, opened 2020-12-08
core: support malloc'ed blobs for body read

dormando:
conn_nread state handles c->item like an item, but allow it to be a
temporary malloced blob via setting c->item_malloced = true.

to be used for buffering value reads in the proxy code.

----

just putting this up as a PR for visibility. took a while of thinking about it to simplify it to this level: it's not safe to examine c->item as an item when it's a malloced buffer, but it never will if we check `c->item_malloced` first :)

Seems a simple enough change otherwise. conn_nread has a few bail points that pass through closing, and `c->item` is handled via that path. I might go back to using the slabber for buffering SETs depending on how large they are or if they intend to be duped locally later.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -704,6 +704,7 @@ conn *conn_new(const int sfd, enum conn_states init_state,
     c->rcurr = c->rbuf;
     c->ritem = 0;
     c->rbuf_malloced = false;
+    c->item_malloced = false;
     c->sasl_started = false;
     c->set_stale = false;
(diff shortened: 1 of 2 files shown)


dormando:
missed a spot. think that's it though.

any cases which end at store_item() must be using the slabber, so no need to check there. only shared code is the nread bits.

dormando:
Out in 1.6.10

PR #751, opened 2020-12-15
Allow metadump to output keys without URI encoding them

smukil:
When "lru_crawler metadump all/<slabid>" is given the "noencode"
option, it outputs non-URL-encoded keys.

TODO: Add test

### DIFF
--- a/crawler.c
+++ b/crawler.c
@@ -40,18 +40,19 @@ typedef void (*crawler_doneclass_func)(crawler_module_t *cm, int slab_cls);
 typedef void (*crawler_finalize_func)(crawler_module_t *cm);
 
 typedef struct {
-    crawler_init_func init; /* run before crawl starts */
-    crawler_eval_func eval; /* runs on an item. */
+    crawler_init_func init;           /* run before crawl starts */
+    crawler_eval_func eval;           /* runs on an item. */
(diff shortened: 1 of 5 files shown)


smukil:
This patch just passes a boolean parameter from process_command() all the way to lru_crawler_start() to convey the "noencode" information to the crawler. An alternate method that uses a new crawler module type is written here:
https://github.com/memcached/memcached/pull/752

I'd like to get your input on which one is better. I personally prefer #752

crawler.c
```diff
@@ -237,19 +238,24 @@ static void crawler_expired_eval(crawler_module_t *cm, item *search, uint32_t hv
...
     /* Ignore expired content. */
     if ((it->exptime != 0 && it->exptime < current_time)
         || is_flushed) {
         refcount_decr(it);
         return;
     }
-    // TODO: uriencode directly into the buffer.
-    uriencode(ITEM_key(it), keybuf, it->nkey, KEY_MAX_URI_ENCODED_LENGTH);
+    key_ptr = ITEM_key(it);
+    if (!cm->noencode) {
```
smukil, line 251: I primarily prefer #752 so that we can avoid this if() for every key.

dormando:
closing these out; we'll come up with something better probably.

PR #752, opened 2020-12-15
Allow metadump to output keys without URI encoding them (Method 2)

smukil:
This patch creates a new crawler module type called
CRAWLER_METADUMP_NOENCODE. When given the "noencode" option while
doing a metadump operation, this new module type will be evaluated
on every key which skips URI encoding them before outputting them.

TODO: Add test

### DIFF
--- a/crawler.c
+++ b/crawler.c
@@ -40,16 +40,16 @@ typedef void (*crawler_doneclass_func)(crawler_module_t *cm, int slab_cls);
 typedef void (*crawler_finalize_func)(crawler_module_t *cm);
 
 typedef struct {
-    crawler_init_func init; /* run before crawl starts */
-    crawler_eval_func eval; /* runs on an item. */
+    crawler_init_func init;           /* run before crawl starts */
+    crawler_eval_func eval;           /* runs on an item. */
     crawler_doneclass_func doneclass; /* runs once per sub-crawler completion. */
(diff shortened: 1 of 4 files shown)


smukil:
Alternate method here:
https://github.com/memcached/memcached/pull/751

dormando:
Thanks!

The module approach is probably the better one. We have some design to work through.

1) This would be the only non-deprecated case in the code where a binary key _could_ be printed into a text protocol.
2) Because of the above this would make more sense as a module that prints a different/condensed format. It's impossible to safely parse the output of metadump with simple code.
3) I was hoping to see some data to drive justification of adding more options/the new module here. Is this saving CPU/space/time/etc? How much?

Is the benefit worth devising a new dump format? What would that format be? :)

dormando:
Noting here: I'd thought this through and decided I'd prefer to add a "meta crawler" command (`mc` perhaps?), which lets the user pick what gets returned via tokens and use the same format as metaget. For non binary encoded keys this would dump straight text, but with the binary flag it would dump base64.

I don't want to add an entirely different protocol into the core code. If you are writing a client that would dump and replay/restore keys in this case, you would not need to decode any base64 encoded keys. You would simply need to ensure the 'b' flag gets copied before doing a data fetch or similar. This I believe solves the concern about encoding/decoding without breaking binary key support.

Will create a separate PR/issue for this though.

PR #766, opened 2021-01-29
The total number of UDP datagrams required for the message is calculated incorrectly.

theoryfate:
UDP_MAX_PAYLOAD_SIZE actually contains the length of the private UDP header, but resp->tosend only contains the length of the data part.

The number of required UDP packets calculated by the original code will be less than the actual need.

E.g:
1000000/1400 = 714.2 ceil 715
1000000/1392 = 718.3 ceil 719
Actually 719 datagrams are needed, and 715 is wrong.


dormando:
thanks!

dormando:
Out in 1.6.10

PR #767, opened 2021-01-30
Logger: customize max sleep

iadbbwarln:
allow to change logger max sleep time. Usefull to increase on
mostly idle cases like docker on dev machine.

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -510,7 +510,6 @@ static void logger_thread_sum_stats(struct logger_stats *ls) {
     STATS_UNLOCK();
 }
 
-#define MAX_LOGGER_SLEEP 1000000
 #define MIN_LOGGER_SLEEP 1000
 
 /* Primary logger thread routine */
@@ -541,10 +540,10 @@ static void *logger_thread(void *arg) {
 
         /* TODO: abstract into a function and share with lru_crawler */
         if (!found_logs) {
-            if (to_sleep < MAX_LOGGER_SLEEP)
+            if (to_sleep < settings.logger_max_sleep)
                 to_sleep += to_sleep / 8;
-            if (to_sleep > MAX_LOGGER_SLEEP)
-                to_sleep = MAX_LOGGER_SLEEP;
+            if (to_sleep > settings.logger_max_sleep)
+                to_sleep = settings.logger_max_sleep;
         } else {
             to_sleep /= 2;
             if (to_sleep < MIN_LOGGER_SLEEP)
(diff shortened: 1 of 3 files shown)


iadbbwarln:
This PR for discussion.

Problem: memcached on idle spins some busy loop and burns CPU cycles. This is part of problem on dev machines inside docker. Nowadays devs runs multiple docker machines with services like memcached in background.

This change allows to tune logger max sleep time, this allows memched to sleep longer and wake up not so often.

If this approach is viable I planning to use same strategy for `lru_maintainer_thread`.

dormando:
Hey, thanks!

I did one of these changes recently to the idle timeout thread: 2896cc6c49c2bcd064e00da8bdf0dcaf1a481e73 - since we now more cleanly shut down, having long idle timers makes shutdown take forever. So at least it needs to use this pattern. I would also prefer a "dev mode" setting but not suure how I would want it to work yet: it could simply crank all the max sleeps to 60s (avoiding adding 10+ timeout override options you'd have to set), but would be better if there was some way to ensure the threads actually do work within a reasonable wakeup timeframe. Else when people do development work in short bursts these threads won't end up doing anything.

It's probably impossible since the answer is likely "in dev mode things wait on cond wakeups and then an these get kicked off when activity happens" but we'd have to come up with a good way of doing that without adding branches to hot paths.

iadbbwarln:
> Hey, thanks!
> 
> I did one of these changes recently to the idle timeout thread: [2896cc6](https://github.com/memcached/memcached/commit/2896cc6c49c2bcd064e00da8bdf0dcaf1a481e73) - since we now more cleanly shut down, having long idle timers makes shutdown take forever. So at least it needs to use this pattern. I would also prefer a "dev mode" setting but not suure how I would want it to work yet: it could simply crank all the max sleeps to 60s (avoiding adding 10+ timeout override options you'd have to set), but would be better if there was some way to ensure the threads actually do work within a reasonable wakeup timeframe. Else when people do development work in short bursts these threads won't end up doing anything.
> 
> It's probably impossible since the answer is likely "in dev mode things wait on cond wakeups and then an these get kicked off when activity happens" but we'd have to come up with a good way of doing that without adding branches to hot paths.

Ok, "dev mode"! :) I think now logger does polling (aka busy loop with sleep).
Currently I hav such ideas, please take a look at them:
* Go into "push" mode: caller will signal logger thread when there is need to wake up logger thread.
* Do dynamic and after we reach max sleep go from poll to push.
* Don't use logger threads (log directly from call place) in "dev mode"

dormando:
Going to close this one out, because I think you're on a better approach in the newer PR.

PR #768, opened 2021-01-31
Logger: pause thread if no watchers available

iadbbwarln:
In case we don't have any watchers we pause logger thread
this will reduce context switches useful to spare cpu
cycles.

Inspired by: #767

Note: there can be race condition between:
```
    watcher_count++;
    pthread_cond_signal(&logger_pause_cond);
```
and
```
            if (watcher_count > 0) {
...
              pthread_cond_wait(&logger_pause_cond, &logger_pause_lock);
```
Possible to address using:
* atomic
* mutex
* ignore this problem

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -32,6 +32,8 @@ static unsigned int logger_count = 0;
 static volatile int do_run_logger_thread = 1;
 static pthread_t logger_tid;
 pthread_mutex_t logger_stack_lock = PTHREAD_MUTEX_INITIALIZER;
+static pthread_cond_t logger_pause_cond = PTHREAD_COND_INITIALIZER;
+static pthread_mutex_t logger_pause_lock = PTHREAD_MUTEX_INITIALIZER;
 
 pthread_key_t logger_key;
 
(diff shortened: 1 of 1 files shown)


iadbbwarln:
cc @dormando, any comment on this? Please let me know if I can improve?

dormando:
Sorry I'm a bit behind right now; I'll review this as soon as I can.

Pretty sure doing it this way means we'd have to barrier or mutex around every access to watcher_count...

dormando:
Hey @iadbbwarln sorry for the delay in looking into this. I like the idea of having the logger thread sit on a condition if the watcher count drops to 0. Think there are a few things that prevent this patch from working. The lock does need to protect `watcher_count`, for one.

The `logger_stack_lock` already protects watcher_count's increment portion and is used in the poll thread. in fact `watcher_count--` is only used in `close_watcher()`, which is called from `poll_watchers()`, while the stack lock is held. So all we really need to do is add a condition variable and move stack lock in the poll thread up a bit and sleep if the watcher count is 0?

I think?

The other complicating factor is the logs will be full if the poller isn't reading from them. Need to test what happens when going from 0 watchers to 1 watcher with a full log; instead of getting a ton of junk it should only get new data.

dormando:
Referenced a commit that fixes this. `watcher_count` and `watchers` were already protected by logger stack lock, so just had to add a condition to signal on. Also required setting a minimum GID so it wouldn't unload log backlog accidentally on new watchers.

dormando:
released in 1.6.11

PR #769, opened 2021-02-09
Fix memory leak

binho1991:
I observed a concern in this project with Memory Leak even when the program is ending or when has no memory to allocate. This pull request follows the same logic.

Some places I observed this concern:
https://github.com/memcached/memcached/blob/c472369fed5981ba8c004d426cee62d5165c47ca/extstore.c#L118-L121

https://github.com/memcached/memcached/blob/c472369fed5981ba8c004d426cee62d5165c47ca/items.c#L1272-L1281

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -986,6 +986,7 @@ void memcached_thread_init(int nthreads, void *arg) {
 
     threads = calloc(nthreads, sizeof(LIBEVENT_THREAD));
     if (! threads) {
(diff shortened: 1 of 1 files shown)


dormando:
Hey,

I don't see any reason to add these frees when the exit() call is right there. Usually I try to explicitly free when there's a chance the caller could change. Thanks for the submission!

PR #780, opened 2021-03-24
Adding an option to link jemalloc as the malloc implementation

tharanga:


configure.ac
```diff
@@ -114,6 +114,9 @@ AC_ARG_ENABLE(static,
 AC_ARG_ENABLE(unix_socket,
   [AS_HELP_STRING([--disable-unix-socket], [Disable unix domain socket])])
 
+AC_ARG_ENABLE(enable_linking_jemalloc,
+  [AS_HELP_STRING([--enable_linking_jemalloc], [Link jemalloc as the memory allocator implementation])])
```
dormando, line 118: help string should be dashes not underscores. Also maybe just --enable-jemalloc ?

configure.ac
```diff
@@ -217,6 +220,10 @@ if test "x$enable_unix_socket" = "xno"; then
     AC_DEFINE([DISABLE_UNIX_SOCKET],1,[Set to nonzero if you want to disable unix domain socket])
 fi
 
+if test "x$enable_linking_jemalloc" = "xyes"; then
+    LIBS="$LIBS -ljemalloc"
```
dormando, line 224: Are there any easy tests we can do here? pkg-config or something?

dormando (changes requested):
nits, thanks!

dormando:
@tharanga pinging again

dormando:
Test fails here :)

tharanga:
@dormando if everything is good, I will squash this.

configure.ac
tharanga: Added `t/jemalloc.t` which uses `ldd` to check the binary.

configure.ac
tharanga: Done

Makefile.am
```diff
@@ -124,9 +124,13 @@ test_basic_tls:
 	fi
 endif
 
+
 test:	memcached-debug sizes testapp
 	$(builddir)/sizes
 	$(builddir)/testapp
+if ENABLE_JEMALLOC
+	@JEMALLOC_TEST=1 prove $(srcdir)/t/jemalloc.t
```
dormando, line 132: Is there another runtime/startitme way to detect if jemalloc exists? This causes this one test to run not in parallel. I spent a lot of time running these tests :)

tharanga: We need to somehow wire the fact jemalloc is linked from configuration to tests. Though a bit of an overkill, a different make target would work I believe?

dormando: https://github.com/jemalloc/jemalloc/issues/388 - I see a couple decent approaches here. Should be easy to add a `supports_jemalloc()` test to the ones in t/lib/MemcachedTest.pm

dormando: Looking at the test again.. trying to figure out what the goal of it is. If jemalloc is enabled, it's checking that it's enabled via ldd?

So the supports() test could be looking for something that configure would dump (like a "HAVE_JEMALLOC" in config.h, would be pretty easy to add)

configure.ac
dormando: I can't remember what I was thinking at the time, but I bet what I meant was detecting if jemalloc existed on the system before trying to link it. So by "Tests" I meant autoconf tests. That would make more sense, I think? :)

Makefile.am
dormando: maybe ignore this comment and see the other one.

configure.ac
tharanga: Got it. `dpkg` seems to be providing both lib existence and the version.

tharanga: Let me try with `ldconfig`

PR #781, opened 2021-03-27
seccomp: extend allowed rules for extended usage

anthraxx:
This puts syscalls into the allow list that were triggered on latest
systems including extended usage tests, signal and systemd handling.


dormando:
Thanks!

dormando:
Out in 1.6.10

PR #785, opened 2021-05-04
hash: add XXH3 to list of hash algorithms.

dormando:
needs more testing. in a "pure miss" test to help show hashing overhead,
with four worker threads and large ascii multigets there is a measurable
performance boost for longer keys. Very short keys (10-15 bytes) are
about even.

for 150ish byte keys 12m rps -> 13m rps roughly. Not too bad!

need to test hash table distribution for large sets of keys before
merging. need to consider if worth switching the defaults now or later.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,5 @@
 bin_PROGRAMS = memcached
-pkginclude_HEADERS = protocol_binary.h
+pkginclude_HEADERS = protocol_binary.h xxhash.h
 noinst_PROGRAMS = memcached-debug sizes testapp timedrun
 
 BUILT_SOURCES=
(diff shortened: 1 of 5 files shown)


dormando:
Ok... Still need to test a bit more.. also wish there was an easy way to not pull in all 5,000 lines of it when we only need a couple functions :(

dormando:
think switching defaults should come later.

dormando:
Did some quick testing with bench-warmer loading 1m, 10m, etc key loads. Using a small patch to `associ_iterator` to print the depth if over 0. Distribution is nearly the same between old/new defaults, at a glance. Doesn't give me reason to do anything more thorough. Checked total depth, number of buckets over 1, and max depth.

Impressively close, though cringing to see a handful of buckets do end up at depths of 5+ :)

dormando:
Out in 1.6.10

PR #786, opened 2021-05-06
Update AUTHORS

BrianAker:
This should have been done a long time ago.


dormando:
Thanks :) my e-mail is dormando at rydia.net though, heh.

PR #790, opened 2021-05-30
logger: Implementation proposal for LOG_CONNEVENTS

LINKIWI:
This PR adds support for recognizing a new watcher command `connevents` which enables the existing but not yet implemented `LOG_CONNEVENTS` flag. It adds support for logging events on new connections and closed connections.

Consider this a draft implementation proposal. There's still some testing left to do (esp. with protocols other than TCP), and docs and tests need to be added as well. I'll work on those items if this change is in the general right direction, and something desirable to include upstream.

The current proposal shapes log lines something like this:

```
ts=1622360517.365343 gid=1 type=conn_new rip=::1 rport=56972 transport=tcp cfd=29
ts=1622360520.236421 gid=2 type=conn_close rip=::1 rport=56972 transport=tcp cfd=29
ts=1622360532.667576 gid=3 type=conn_new rip=127.0.0.1 rport=49332 transport=tcp cfd=29
ts=1622360534.714145 gid=4 type=conn_close rip=127.0.0.1 rport=49332 transport=tcp cfd=29
```

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -1,5 +1,6 @@
 /* -*- Mode: C; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 
+#include <arpa/inet.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
@@ -56,6 +57,8 @@ static const entry_details default_entries[] = {
     [LOGGER_SLAB_MOVE] = {LOGGER_TEXT_ENTRY, 512, LOG_SYSEVENTS,
         "type=slab_move src=%d dst=%d"
     },
+    [LOGGER_CONNECTION_NEW] = {LOGGER_CONNECTION_NEW_ENTRY, 512, LOG_CONNEVENTS, NULL},
+    [LOGGER_CONNECTION_CLOSE] = {LOGGER_CONNECTION_CLOSE_ENTRY, 512, LOG_CONNEVENTS, NULL},
 #ifdef EXTSTORE
     [LOGGER_EXTSTORE_WRITE] = {LOGGER_EXT_WRITE_ENTRY, 512, LOG_EVICTIONS, NULL},
     [LOGGER_COMPACT_START] = {LOGGER_TEXT_ENTRY, 512, LOG_SYSEVENTS,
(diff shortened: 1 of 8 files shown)


dormando:
Thanks! Definitely a thing. Just took a quick skim; info on closed by timeout/error/etc probably useful?

I think this week will be a PR/issue scrub so I'll get a chance to take a closer look. Took half a year to start to wash off 2020.

LINKIWI:
Thanks for taking a look. I'm thinking about adding an `enum close_reasons` in `memcached.h` and adding a new `close_reason` field on `struct conn`, optionally set at any callsite that sends the connection into `conn_closing` or similar. What do you think of that approach?

dormando:
hmm. thinking about it. Downside; more stuff on bloated enum. Upside: stats conns may be able to print reason why something's in its state as well..

There are a lot of places where `conn_closing` gets set. Debating if it's worth doing the enum, or if the places that transition is called should all have logger calls. I think that's what I was originally intending on doing; many of those have prints or verbose prints, which should change to logger_log and eventually get removed.

LINKIWI:
Another practical consideration is that it might be easy to forget a LOGGER_LOG at a conn_closing transition callsite in future code changes, could be a lot of duplication as well. But yeah I agree that `struct conn` is a bit bloated already today.

I'm happy to try either approach though and put up a patch to see how it looks in practice.

dormando:
All those transitions need to be marked with the reason anyway.. up to you I guess :) You can take a look at the conn struct packing and see if there's anywhere to just drop an enum without inflating it. See which one looks better overall, sure.

memcached.h
```diff
@@ -245,6 +245,12 @@ enum stop_reasons {
     EXIT_NORMALLY
 };
 
+enum close_reasons {
+    NORMAL_CLOSE,
+    ERROR_CLOSE,
```
LINKIWI, line 250: There might be value in making this more granular, maybe: server error, client/protocol error, others?

dormando: looks like without making it more granular, it should be the default? all of these close_reason = lines seem redundant.

dormando:
Cool. Mind giving this a squash/cleanup? I can't find think of any arguments with it.

logger.c
```diff
@@ -689,6 +728,39 @@ static void _logger_log_item_store(logentry *e, const enum store_item_type statu
...
 }
 
+static void _logger_log_conn_event(logentry *e, struct sockaddr *addr,
+        enum network_transport transport, enum close_reasons reason, int sfd) {
+    struct logentry_conn_event *le = (struct logentry_conn_event *) e->data;
+
+    memset(le->rip, 0, sizeof(le->rip));
+    le->sfd = sfd;
+
+    switch (addr->sa_family) {
```
dormando, line 738: Actually noodling this a bit; it's most optimal to do the binary->text conversion in the logger thread (parse-ise) and do a fast memcpy of the sockaddr stuff here? Any reason not to?

In general the inline log events should be as lean as possible and then we backload all parsing/decoding into the logger thread.

LINKIWI: Makes sense. I pushed up 8579cadf1258d62864971672fe3fcb2c34b5fe8d which moves this block of logic into the logger thread and keeps the hot path minimal. Let me know what you think (it's a bit more verbose).

logger.c
```diff
@@ -254,27 +254,75 @@ static int _logger_thread_parse_extw(logentry *e, char *scratch) {
...
 static int _logger_thread_parse_cce(logentry *e, char *scratch) {
     int total;
+    unsigned short rport = 0;
+    char rip[64];
+    struct logentry_conn_event *le = (struct logentry_conn_event *) e->data;
     const char * const transport_map[] = { "local", "tcp", "udp" };
     const char * const reason_map[] = { "error", "normal", "idle_timeout", "shutdown" };
 
-    struct logentry_conn_event *le = (struct logentry_conn_event *) e->data;
+    memset(rip, 0, sizeof(rip));
```
dormando, line 300: make a __logger_util_something to avoid the redundancy?

LINKIWI: Pushed up c722073b7189566e24c688415a4234c1f1a8280d; let me know what you think.

dormando: Looks good, squash it up and clean up anything else you want?

I need to change some things around in the logger code, so trying to get these in first so I don't invalidate your PR's :)

LINKIWI: Thanks! I don't think there's anything else I want to change so I've squashed everything down.

dormando:
Ah, duh. `doc/protocol.txt` - I think that's also missing sysevents but I'll get that one. add a bit for connevents?

PR #791, opened 2021-05-31
tls: Parameterize minimum supported TLS protocol version

LINKIWI:
This change proposes the addition of a new extended option `ssl_min_version` to parameterize the minimum supported TLS protocol version. Currently, the minimum supported version is hard-coded as TLS v1.2; to maintain backwards compatibility, this change sets TLS v1.2 is the default value when the option is unspecified.

`t/ssl_proto_version.t` likely only works in environments with a decently recent version of OpenSSL. Older versions don't support TLS v1.3, and newer versions explicitly disable support for EOL'd versions like 1.1 and below.

The primary motivation of this change is to provide the ability to enforce TLS v1.3 usage in hardened environments.


dormando:
Yup, this is a thing we need. Thanks!

dormando:
@tharanga guessing you won't have time to review this? :)

Sorry, it may take us a bit to get to this. I've been chipping at backlog.

LINKIWI:
No worries! Thanks for taking a look.

I've deployed a build with this patch to some of our production clusters and have verified the following permutations:

* OpenSSL 1.0.1f (older hardware) TLS v1.1 + `ssl_min_version=1` -> handshake success
* OpenSSL 1.0.1f TLS v1.2 + `ssl_min_version=1` -> handshake success
* OpenSSL 1.0.1f TLS v1.1 + `ssl_min_version=2` -> handshake fail
* OpenSSL 1.1.1+ TLS v1.2 + `ssl_min_version=1` -> handshake success
* OpenSSL 1.1.1+ TLS v1.3 + `ssl_min_version=1` -> handshake success
* OpenSSL 1.1.1+ TLS v1.2 + `ssl_min_version=2` -> handshake success
* OpenSSL 1.1.1+ TLS v1.3 + `ssl_min_version=2` -> handshake success
* OpenSSL 1.1.1+ TLS v1.2 + `ssl_min_version=3` -> handshake fail
* OpenSSL 1.1.1+ TLS v1.3 + `ssl_min_version=3` -> handshake success

t/ssl_proto_version.t
```diff
@@ -0,0 +1,32 @@
...
+
+my $server = new_memcached('-o ssl_min_version=2');
+
+# Minimum supported protocol version
+$sock = $server->new_sock(undef, 'TLSv1_2');
+print $sock "version\r\n";
+like(scalar <$sock>, qr/VERSION/, "handshake with minimum proto version");
+
+# Above minimum supported protocol version
+$sock = $server->new_sock(undef, 'TLSv1_3');
```
LINKIWI, line 22: There's a [convention](https://github.com/memcached/memcached/blob/c472369fed5981ba8c004d426cee62d5165c47ca/t/ssl_session_resumption.t#L43) for skipping tests to ensure compatibility with environments on older versions of OpenSSL that don't ship with TLS v1.3 support, not sure if we should do the same here.

dormando: tests shouldn't fail on older distro builds, so it should at least handle it properly.

LINKIWI:
I've made the following changes to ensure compatibility with older distro versions:

* Add a `HAVE_OPENSSL_111` macro, set at `./configure` time to compile-gate usage of `TLS1_3_VERSION`
* Skip some assertions in the test if TLS v1.3 isn't available

I tested this on a system with OpenSSL 1.1.0 (minimum required by configure.ac) and on 1.1.1. Let me know what you think.

dormando:
Probably won't be giving this a thorough review for this upcoming release, but we'll definitely get it in for the next one. I have some vacation time this month too, so sorry for any extra delays.

memcached.c
```diff
@@ -3959,9 +3961,16 @@ static void usage(void) {
            "   - ssl_wbuf_size:       size in kilobytes of per-connection SSL output buffer\n"
            "                          (default: %u)\n", settings.ssl_wbuf_size / (1 << 10));
     printf("   - ssl_session_cache:   enable server-side SSL session cache, to support session\n"
-           "                          resumption\n");
+           "                          resumption\n"
+           "   - ssl_min_version:     minimum protocol version to accept, default is 2(TLSv1.2).\n"
+#ifdef HAVE_OPENSSL_111
```
dormando, line 3966: Hmm.. another place with an external library that has random strings which need to be updated in two places. Any better options for this? Guessing we just have to update the code when new versions appear and someone has to manually notice this?

dormando: Oh; we have some asserts that've been added near the docs for when the changed default no longer matches what's in the docs. add one of those at least?

dormando:
Well I have no major complaints other than the one comment. Mind squashing up the history or anything else you'd want to do before merge?

memcached.c
LINKIWI: `HAVE_OPENSSL_111` is defined manually in `configure.ac` conditionally on the value of `OPENSSL_VERSION_NUMBER` (a constant whose name generally speaking shouldn't change across versions).

I've reverted the changes to `configure.ac` and replaced `HAVE_OPENSSL_111` with an evaluation against `OPENSSL_VERSION_NUMBER` in 1c5ced4b691580c90cfcb10ee10bff6276318a3f. I think this is a bit clearer, though comes at the expense of repeating the 1.1.1 version constants in a few places. Let me know if this addresses your concern?

dormando: Ah sorry, What I meant is the help (`-h`) text itself. For other help output with defaults we have asserts that check that the default hasn't changed without the documentation getting updated? That doesn't seem to be what you fixed just here.

LINKIWI: Ah my bad. In any case though I do feel like 1c5ced4b691580c90cfcb10ee10bff6276318a3f (the unrelated change) is good to have.

In 785c51064b5d3d910a4a406a8e55fbb3608bc87e I parameterized `settings.ssl_min_version` in the help text so that the documentation is always consistent with its default value. This looks a bit more similar to how it's done for the other options. Not entirely sure if this is what you were looking for but let me know.

I think the `verify_default("ssl_min_version", settings.ssl_min_version == TLS1_2_VERSION);` accomplishes the assertion you alluded to but correct me if I'm wrong.

This is what the text output looks like now when compiling on 1.1.1:

```
   - ssl_min_version:     minimum protocol version to accept (default: tlsv1.2)
                          valid values are 0(tlsv1.0), 1(tlsv1.1), 2(tlsv1.2), or 3(tlsv1.3).
```

And on 1.1.0:

```
   - ssl_min_version:     minimum protocol version to accept (default: tlsv1.2)
                          valid values are 0(tlsv1.0), 1(tlsv1.1), or 2(tlsv1.2).
```

dormando: Got it, looks good. If those were in before I'd missed them somehow.

dormando:
Looks okay to me. I think @tharanga might also take a look - but go ahead and squash it or do anything else you want before I merge it.

Thanks again!

PR #792, opened 2021-06-04
stats_prefix.c: Check for NDEBUG before using total_written variable

kraj:
When using NDEBUG assert macro is ineffective which is caught by latest
clang and reports that total_written is set but unused. Therefore check
for NDEBUG to make sure assert is used only when its effective

Fixes
error: variable 'total_written' set but not used [-Werror,-Wunused-but-set-variable]
    size_t size = 0, written = 0, total_written = 0;
                                  ^

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/stats_prefix.c
+++ b/stats_prefix.c
@@ -127,8 +127,10 @@ char *stats_prefix_dump(int *length) {
     PREFIX_STATS *pfs;
     char *buf;
     int i, pos;
-    size_t size = 0, written = 0, total_written = 0;
-
+    size_t size = 0, written = 0;
+#ifndef NDEBUG
+    size_t total_written = 0;
+#endif
     /*
      * Figure out how big the buffer needs to be. This is the sum of the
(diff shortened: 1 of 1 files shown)


dormando:
thanks!

dormando:
I think most/all cases in the code that check for a debug build state actually look for `MEMCACHED_DEBUG`?

kraj:
> MEMCACHED_DEBUG

how  do we define it ? if it is not based on NDEBUG then we can not use it because libc uses NDEBUG to decide the fate of assert macro, so we better use that

dormando:
Guess it's just a weirdness in the makefiles: if NDEBUG is not being set, MEMCACHED_DEBUG is set. so all you should need to do is switch ifndef to ifdef's. It's not a huge deal and I could take as-is though, it looks like this convention was set recently.

dormando:
Out in 1.6.10

PR #795, opened 2021-06-09
Meta command followups.

dormando:
PR follows the fixes for https://github.com/memcached/memcached/issues/546

Finally had a giant brain moment while wondering where to stash flag data for the set results: turns out wbuf has plenty of space! A small parser to replay some tokens was all that was needed. Now CAS can be atomically returned from a meta set.

There're a handful of fixes in here which should finalize the incompatible changes for meta.. anything new should be new.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -523,7 +523,7 @@ VA <size> <flags>*\r\n
 If the request did not ask for a value in the response (v) flag, the server
 response looks like:
 
-OK <flags>*\r\n
+HD <flags>*\r\n
 
 If the request resulted in a miss, the response looks like:
 
@@ -609,7 +609,7 @@ access time.
 - v: return item value in <data block>
 
 The data block for a metaget response is optional, requiring this flag to be
-passed in. The response code also changes from "OK" to "VA <size>"
+passed in. The response code also changes from "HD" to "VA <size>"
 
 These flags can modify the item:
 - N(token): vivify on miss, takes TTL as a argument
@@ -698,7 +698,7 @@ the reply, which is of the format:
 
 Where CD is one of:
 
-- "OK" (STORED), to indicate success.
+- "HD" (STORED), to indicate success.
 
 - "NS" (NOT_STORED), to indicate the data was not stored, but not
 because of an error.
@@ -746,7 +746,7 @@ See description under 'Meta Get'
 
 Noreply is a method of reducing the amount of data sent back by memcached to
 the client for normal responses. In the case of metaset, a response that
-would start with "OK" will not be sent. Any other code, such as "EX"
+would start with "HD" will not be sent. Any other code, such as "EX"
 (EXISTS) will still be returned.
 
 Errors are always returned.
@@ -782,7 +782,7 @@ The response is in the format:
 
 Where CD is one of:
 
-- "OK" (DELETED), to indicate success
+- "HD" (DELETED), to indicate success
 
 - "NF" (NOT_FOUND), to indicate that the item with this key was not found.
 
@@ -847,7 +847,7 @@ The response is in the format:
 
 Where CD is one of:
 
-- "OK" to indicate success
+- "HD" to indicate success
 
 - "NF" (NOT_FOUND), to indicate that the item with this key was not found.
 
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -96,16 +96,19 @@ noted in the 'Commands' section above.
 
 All meta commands follow a basic syntax:
 
-<cm> <key> <flag1> <flag2> <...>\r\n
+<cm> <key> <datalen*> <flag1> <flag2> <...>\r\n
 
 Where <cm> is a 2 character command code.
+<datalen> is only for commands with payloads, which is set.
 
 Responses look like:
 
-<RC> <flag1> <flag2> <...>\r\n
+<RC> <datalen*> <flag1> <flag2> <...>\r\n
 
 Where <RC> is a 2 character return code. The number of flags returned are
 based off of the flags supplied.
+
+<datalen> is only for responses with payloads, with the return code 'VA'.
 
 Flags are single character codes, ie 'q' or 'k' or 'I', which adjust the
 behavior of the command. The flags are reflected in the response. The order of
@@ -671,9 +674,11 @@ The meta set command a generic command for storing data to memcached. Based
 on the flags supplied, it can replace all storage commands (see token M) as
 well as adds new options.
 
-ms <key> <flags>*\r\n
+ms <key> <datalen> <flags>*\r\n
 
 - <key> means one key string.
+
+- <datalen> is the length of the payload data.
 
 - <flags> are a set of single character codes ended with a space or newline.
   flags may have strings after the initial character.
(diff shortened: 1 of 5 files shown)


dormando:
not doing a flag for the old set behavior, since this was always broken before this PR (extra space on OK response).

dormando:
only going to document the flag in the release notes... don't want people to accidentally use it.

doc/protocol.txt
```diff
@@ -742,7 +746,7 @@ See description under 'Meta Get'
 
 Noreply is a method of reducing the amount of data sent back by memcached to
 the client for normal responses. In the case of metaset, a response that
-would start with "OK" will not be sent. Any other code, such as "EX"
+would start with "HD" will not be sent. Any other code, such as "EX"
```
TysonAndre, line 749: Unrelated to the docs change, I haven't seen this before: Will this behavior for meta-set cause issues with the embedded proxy you're working on?
(I'm working on supporting meta commands in a different proxy)

E.g. 

1. if there are 2 clients sending a meta-set command to set a key to a value if it doesn't exist (`set key 5 q Madd`), which client would get the response.
2. If a single client of the proxy sends 2 pipelined requests `set key1 5 q Madd\r\nvalue\r\n` (noreply), then `md key2 C(token)\r\n`, and key1 and key2 go to different backend servers,

    Then the client (of the proxy) may receive the response for key2 before the optional response for key1 because the proxy can't tell whether or not it needs to await key1, and it may misbehave. (e.g. misinterpret `EX\r\n` for key2 as applying to key1 if `k` or `O(token)` isn't added).

--------

To reduce memory usage of proxies or of pipelining libraries,

- It would be useful to have a very quiet flag (instead of, or in addition to) the quiet flag to not send errors except for protocol errors or client/server errors (e.g. don't send response about expected conditions such as invalid cas token, missing key, already existing key, etc)

  This is my understanding of how `noreply` works in the original text protocol for set/add/delete/etc?

dormando: 1. I don't think there's a race? I've been implementing this via strict ordering in response parsing. There is, as you noticed, an issue with noreply, which I'll answer.
2. I have an option question about this; in the meantime noreply isn't supported. What's likely to happen is the proxy will have to inject `mn`'s to separate noreply requests. There isn't any real way to do it otherwise.

This plays out in a few stages:
1. Clients will be doing something like 'mg key1 q, mg key2 q, mn`, to confirm when all their stacked work is done.
2. Proxy ends up splitting these gets out to different backends, adding an `mn` after each.
3. The top level `mn` gets replayed once all sub-requests are completed.
4. This is effectively the same as what a client would do if it's sending quiet requests out to multiple backends in a pool directly; it has to tag an `mn` at the end of each set of keys. This is how the binprot operates.

This still saves syscalls on the network and a bit of effort on the client but it can't be as efficient as directly pinging a server.

I changed the semantics of noreply to add errors, because otherwise errors cause the protocol to completely desync. Which is now even worse in the case of a proxy; if you send a malformed request, it may get partway interpreted via the next request in the pipeline and give back crazy errors. It's also a security fault if people can inject newlines. Since noreply is allowed to surface bad errors the proxy can know when the protocol's desynced and close out safely.

TysonAndre: > I don't think there's a race? I've been implementing this via strict ordering in response parsing. There is, as you noticed, an issue with noreply, which I'll answer.

I was talking about `set key 5 q Madd`, where `q` is the `noreply` flag - but still, if you add `mn` requests between them, then it's unambiguous which request a response before `MN\r\n` is.

> I have an option question about this; in the meantime noreply isn't supported. What's likely to happen is the proxy will have to inject `mn`'s to separate noreply requests. There isn't any real way to do it otherwise.

Alternately, the proxy could remove or replace the `q`(`noreply`) flag with a no-op, record the fact the request was originally a noreply, wait for the response normally, then filter out the response if it was `HD`

What are your thoughts on a no-op flag (e.g. lowercase `n`) for meta commands - this would be ignored by proxies and memcache servers and have no special behavior, and the only difference is that memcached would stop sending `CLIENT_ERROR invalid flag` for `n`

- Removing `q`(noreply) would require potentially copying byte ranges, which is slower for long requests such as sets. Replacing `q` with `n` would allow a proxy to forward the command with barely any work, and check if the reply should be filtered out for the corresponding request when the reply is received
- Extra whitespace (e.g. trailing) is forbidden in the meta protocol

> I changed the semantics of noreply to add errors, because otherwise errors cause the protocol to completely desync. Which is now even worse in the case of a proxy; if you send a malformed request, it may get partway interpreted via the next request in the pipeline and give back crazy errors. It's also a security fault if people can inject newlines. Since noreply is allowed to surface bad errors the proxy can know when the protocol's desynced and close out safely.

I agree with that, but my wording was unclear. I was talking about continuing to send responses on real errors (protocol errors or client/server errors) but not key misses, keys already existing, mismatched CAS tokens, etc. (i.e. noreply for potential application errors, but continue to reply for protocol errors)

Closing on desync is what I'd expect

dormando: I misread the set example, sorry. Think I still answered it.

I'll double check the set noreply semantics; It should match the binary protocol. I don't believe there is as large an overlap with "add, but I don't care to know if it fails" vs "add, but I only care if it fails", which is why noreply tended in that direction.

The `n` flag idea is pretty interesting. I'll give it some more thought. What I have in mind (apologies for not documenting much of this clearly for the proxy; a lot changes quickly so I can't really waterfall it) are one or two proxy-specific "route" opaques. This gives me a couple reserved future optimizations:

- One client-fed router opaque string, which is where you would stick "/foo/" type prefixes or commands. Currently to route to pools you would prefix a key, then the router has to edit the command to remove the prefix if you don't want to waste bytes on the backend. This is probably showing up sooner than later since it's easy.
- One (possibly two or N) router-owned opaque strings. My proxy's internals allow for heavy syscall batching under higher loads. This should mean that you can add an opaque, enable noreply if not already, tag an `mn` after a long batch of commands, then match request/responses definitively. This works well for "common queries" (mostly gets I guess) where miss doesn't net you much information. The proxy can handle reflecting anything the client wanted (original opaque/etc) but doesn't need to parse every response or read as many bytes.
- This is especially useful since `mn` disables response batching on a memcached server, so if we're injecting them frequently it can reduce performance. Injecting them not frequently enough increases latency, however. So some balance will be necessary.

I'm not sure when I'll have time to experiment with this, but probably not until after the shortlist of "v1" features I'm working through on the proxy.

doc/protocol.txt
```diff
@@ -96,22 +97,24 @@ noted in the 'Commands' section above.
 
 All meta commands follow a basic syntax:
 
-<cm> <key> <flag1> <flag2> <...>\r\n
+<cm> <key> <datalen*> <flag1> <flag2> <...>\r\n
 
 Where <cm> is a 2 character command code.
+<datalen> is only for commands with payloads, which like set.
```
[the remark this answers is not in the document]
dormando: thanks!

dormando:
Didn't get any other major reviews going. I've looked it over again and pretty sure I'm happy with this as-is. `HD` is as good as anything.. it can be literally anything so long as it doesn't overlap with any old return codes, and it doesn't.

The mset fixes are all fine. It didn't work before due to the extra space.

Any further changes are easily extensions; new commands with new return codes, new flags to existing commands, and so on.

I'll merge this in to `next` and cut a release once I kick the buildbots back to life.

PR #797, opened 2021-06-11
File descriptor listener

### LINKED ISSUE #789: Listener on inherited socket file descriptor
Currently, memcached supports binding TCP/UDP and Unix domain sockets for listeners. I'm working on an experimental deployment environment in which a parent process manages/creates the listener socket, and simply passes the associated file descriptor directly to memcached to serve connections.

This is useful, for example, when:

* memcached runs behind a shared socket manager like [Einhorn](https://github.com/stripe-archive/einhorn) et al.
* memcached runs as a forked (or even embedded) process, and communicates with the parent over a shared IPC Unix pipe without binding to a network socket.

This is a bit of a narrow use case, but I'm wondering if there would be any interest in having this feature upstream. I have a proof-of-concept patch that would need to be cleaned up a bit, but in summary it exposes two new flags `--listen-fd` (initialized file descriptor) and `--listen-fd-transport` (choose from one of local/unix, TCP, UDP) that can be used instead of `--unix-socket/--port/--udpport`.

<details>
<summary>An example client program can probably describe the use case more succinctly</summary>

```go
package main

import (
	"fmt"
	"os"
	"os/exec"
	"syscall"
	"time"
)

func main() {
	fds, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
	if err != nil {
		panic(err)
	}

	fd0 := os.NewFile(uintptr(fds[0]), "local")
	fd1 := os.NewFile(uintptr(fds[1]), "mc")

	cmd := exec.Cmd{
		Path:       "./memcached",
		Args:       []string{"./memcached", "-vv"},
		// Assume memcached uses FD 3 as the listener and treats it as AF_UNIX
		ExtraFiles: []*os.File{fd1},
		Stdout:     os.Stdout,
		Stderr:     os.Stderr,
	}

	go cmd.Run()

	time.Sleep(1 * time.Second)

	fd0.Write([]byte("stats\r\n"))

	buf := make([]byte, 4096)
	fd0.Read(buf)
}
```
</details>

Feel free to close this issue if this proposal isn't needed.

LINKIWI:
This code change adds the ability to accept connections directly from a file descriptor passed by a socket manager parent process, like systemd et al.

It's experimental and largely untested, especially for UDP transports. Mostly putting up this PR for discussion.

If there's a path to having this in the main source tree, I'll clean it up a bit and write some tests and docs.

Closes #789.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3727,6 +3727,39 @@ static int server_socket_unix(const char *path, int access_mask) {
 #define server_socket_unix(path, access_mask)   -1
 #endif /* #ifndef DISABLE_UNIX_SOCKET */
 
+static int server_socket_fd(int sfd, enum network_transport transport) {
+    struct sockaddr_in addr;
+
+    memset(&addr, 0, sizeof(struct sockaddr_in));
+
+    if (settings.listen_fd_transport == local_transport) {
+        for (int i = 0; i < settings.num_threads; i++) {
+            int threadfd;
+
+            if (i == 0) {
+                threadfd = sfd;
+            } else {
+                threadfd = dup(sfd);
+            }
+
+            dispatch_conn_new(threadfd, conn_new_cmd,
+                              EV_READ | EV_PERSIST, READ_BUFFER_CACHED,
+                              settings.listen_fd_transport, NULL);
+        }
+
+        return 0;
+    }
+
+    if (!conn_new(sfd, conn_listening,
+                  EV_READ | EV_PERSIST, 1,
+                  settings.listen_fd_transport, main_base, NULL)) {
+        fprintf(stderr, "failed to create listening connection\n");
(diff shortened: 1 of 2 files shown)


dormando:
I personally don't care either way, but I wonder if there's a way to find out if more people would be interested in this? We're starting to get a proliferation of options nobody uses. Any user groups/etc or examples of other projects doing this that can provide feedback?

LINKIWI:
This change is currently only used in an experimental capacity for an embedded memcached deployment. I agree that adding unused options adds a lot of smell; perhaps let's close this PR for now and I can revisit it in the future if I'm able to find some compelling use cases within the wider community.

dormando:
I think that's a decent compromise.. thanks!

PR #802, opened 2021-07-01
Expose shutdown_command state in `stats settings`

LINKIWI:
Loosely related follow-up to some earlier changes around the shutdown command. This change proposes exposing the value of `settings.shutdown_command` (set with `-A` or `--enable-shutdown`) in the output for `stats settings`.

```
$ echo "stats settings" | nc -q0 localhost 11211 | grep shutdown
STAT shutdown_command no
```

```
$ ./memcached -A
...
$ echo "stats settings" | nc -q0 localhost 11211 | grep shutdown
STAT shutdown_command yes
```


dormando:
thanks!

4383 (approved):
These changes LGTM

dormando:
Out in 1.6.10

PR #803, opened 2021-07-07
[docker] Add a new user to the docker images

4383:
These changes aim to improve our memcached devtools images to allow us to reuse them during development stages.

These changes create a memcached user for all the images, switch to this user by default once memcached have been configured for building.

That allow to run the memcached server directly when the containers are started, else, without the user the run will fail as we try to execute memcached as root user.

Also we should notice that the alpine image has been fixed by adding git to the list of the requirements to install. Else the build will fails.

### DIFF
--- a/devtools/Dockerfile.alpine
+++ b/devtools/Dockerfile.alpine
@@ -2,13 +2,16 @@ FROM alpine:latest
 
 ARG CONFIGURE_OPTS="--enable-seccomp"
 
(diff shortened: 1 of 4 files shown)


dormando:
Thanks!

Given light of past security issues, I don't think we're able to ship something that will expose the network port by default here. The rest I will look into as soon as I can.

4383:
No problem concerning the expose. Anyway the ports can be published by using the command line `--publish 11211:11211` so I'll drop the related lines.

4383:
Done.

dormando:
Out in 1.6.10

4383:
Thanks

PR #809, opened 2021-07-28
Io queue flow work

dormando:
Tracking a set of core changes I need for the proxy. The deferred IO queue system requires all IO's batch out and in at once. This work will allow an asynchronous mode, where more IO's can go out than need to come back, etc.

This requires some cleanups around cache objects, some ancient code for worker thread notifications (the `CQ` stuff hasn't been touched since like... 2008!), then ideally some small changes to the deferred IO queue interface. Also brings `queue.h` into core which should make the proxy PR a bit smaller.

- [x] import queue.h
- [x] cleanup dead code in cache.c and use STAILQ freelist instead of a malloc'ed array
- [x] move "connection queue" code for new connection dispatch to use cache.c
- [x] move all pipe-write worker thread notifiers to use the client queue object instead
- [x] add eventfd mode for notifier (faster!)
- [x] `io_queue`: allow returning IO's individually, letting the worker thread to figure out when to re-attach and resume the connection.
- [x] `io_queue`: refactor to fix issues with an IO's connection reference not being usable once a connection has been resumed. This is bad for the orphaned IO's.


dormando:
I'd like cache.c even better if it could handle an object queue internally but it wasn't too much more code here. Gutted the `cq` stuff and left queue.h and cache.c code in its place. The freelist for connection queues was a global, now it's per-worker-thread, which will be necessary to scale the next parts.

dormando:
don't like the scary error handling sitting in this part of the code, though I doubt they're ever activated in practice. got the code rewound away from the pipe but still working on it.

dormando:
Nearly done... can now ship IO's back via the notifier queue and it'll call a `return_cb` function against the io (which should have its `io_queue` and `conn` references). In proxy this will immediately resume the lua thread to figure out what to do next.

last bit is working out changes in the submission API: in the full-batch case the `io_queue_t` ownership is passed entirely to the side thread. In the "flow" case the ownership stays on the worker thread, but that's technically implicitly done so I have to carefully check the code to ensure nobody's accessing the queue structure at the wrong time.

dormando:
pretty sure this is what I want... took entirely too much noodling to come up with the simplest interface. tested performance with `bench_event` and it's a little low but will scale with additional threads.
.
Already found one thing to maybe fix... might need a specific method of immediately dispatching redispatched requests...

dormando:
I'm a fucking moron. `return_io_pending` can't use `conn *c` to find the worker thread, it needs to be on `io_pending`. I knew this and I goofed it. hooray waste of time rebasing everything...

dormando:
This is back to being WIP-y since it didn't work once I went to integrate...

dormando:
Going back and forth between this and the proxy PR in an attempt to fix the API.

Lots of small objects that point at each other, and sorting this out has been a headache. Today's API changes have gotten me a lot closer, but I think the hangup now is re-dispatching brand new requests after a parent has been orphaned, which is something the lua interface can do.

I should revisit doing an API limitation of disallowing await()'s from being anything but hash selectors, if this can get things done sooner. It should be plausible to lift the limitation later but it does hamstring the API badly if you can only interject new routines above async routines. I've worked out how to do that but it's a lot more code and maybe another round of API changes, so we can try to avoid it for v1.

dormando:
Just found/fixed a quite serious bug in extstore. Astounding nobody's run into this! Or at least not complained about it...

dormando:
Removing WIP tag agin.. this is ready for review on upstream merge. The proxy stuff (an asynchronous wait interface) built on top of these changes is working so far.

thread.c
```diff
@@ -689,29 +711,36 @@ void dispatch_conn_new(int sfd, enum conn_states init_state, int event_flags,
...
-    char buf[REDISPATCH_MSG_SIZE];
-    LIBEVENT_THREAD *thread = c->thread;
+    notify_worker_fd(c->thread, c->sfd, queue_redispatch);
+}
 
-    buf[0] = 'r';
-    memcpy(&buf[1], &c->sfd, sizeof(int));
-    if (write(thread->notify_send_fd, buf, REDISPATCH_MSG_SIZE) != REDISPATCH_MSG_SIZE) {
-        perror("Writing redispatch to thread notify pipe");
+void timeout_conn(conn *c) {
```
dormando, line 726: Note: would prefer exposing the enum and having 1-2 public functions for notifying a worker of something but don't feel like I need to do this now.

dormando:
The thing that still sucks about the IO interface is the direct struct/variable access from users (q->count-- and such). I'm not sure I have the energy to encapsulate this properly right now but I'm going to give it 10-15 minutes today in case it can be done quickly. Feels bad leaving poor quality abstractions for "later".

Otherwise it's been working okay so far.

dormando:
Still working fine. Going to merge this unless someone else wants to review.

PR #813, opened 2021-08-18
logger: Expose current active watcher count in stats state

LINKIWI:
Proposing exposing `watcher_count` in `stats_state` so that clients can check the number of currently active watchers through stats.

Related: there exists a [TODO in `watcher.t`](https://github.com/memcached/memcached/blob/ddee3e27a031be22f5f28c160be18fd3cb9bc63d/t/watcher.t#L111) that this change could help with (though probably in a follow-up PR if the change proposed here makes sense).


dormando:
Thanks! The watcher count is normally protected by the logger_stack lock. It might be more correct to note the current watcher count before dropping the stack lock, then passing that to the sum/flush stats or tossing it directly?

LINKIWI:
Makes sense--let me know what you think of ee0d36bff0bf8e9096df457f15add1b3976fdeed?

dormando:
Looks good. Saves me from having to tell someone's PHD-thesis static analyzer that there's no harmful bug there :)

dormando:
released in 1.6.11

PR #817, opened 2021-09-08
Fix time-of-check time-of-use bugs

kokke:
Fixing 11 small 'bugs' of the pattern: 'assert(ptr != 0)' after 'ptr' was already dereferenced. 

E.g. changing something of the form:
```C
char* x = c->y; // dereferencing 'c'
assert(c != NULL); // checking 'c'
```
... into

```C
assert(c != NULL); // checking 'c'
char* x = c->y; // dereferencing 'c'
```

I couldn't find out how to advertise this onto the mailing list (as is suggested in the contribution-guide), so hopefully someone can help me there.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -2240,8 +2240,8 @@ enum delta_result_type do_add_delta(conn *c, const char *key, const size_t nkey,
 }
 
 static int try_read_command_negotiate(conn *c) {
-    assert(c->protocol == negotiating_prot);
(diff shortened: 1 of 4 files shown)


dormando:
Thanks!

You just join the mailing list (links from https://memcached.org) and send the link there. I guess it's optional since nobody does it :/

dormando:
out in 1.6.13

PR #825, opened 2021-10-02
Track store errors in thread stats

LINKIWI:
Currently, there's good visibility coverage for most of the [`enum store_item_type`](https://github.com/memcached/memcached/blob/56dc81db316a0b957415e371d20c683fea9d7d2f/memcached.h#L265) statuses, with the exception of the explicit error states `TOO_LARGE` and `NO_MEMORY`.

This change proposes tracking such occurrences in thread stats and exposing them under new stat keys `store_too_large` and `store_no_memory`.

The practical use case where this has been useful: gradually rolling out a more restrictive `--max-item-size` in a cluster and monitoring how much more frequently memcached is rejecting sets due to item sizes that are too large.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -1267,6 +1267,12 @@ integers separated by a colon (treat this as a floating point number).
 |                       |         | with a new expiration time                |
 | touch_misses          | 64u     | Number of items that have been touched    |
 |                       |         | and not found                             |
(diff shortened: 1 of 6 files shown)


dormando:
seems cool, thanks!

dormando:
out in 1.6.13

PR #828, opened 2021-10-14
Fix full test suite under TLS

LINKIWI:
`make test_tls` has some errors:

```
t/maxconns.t ................ 1/? 
#   Failed test 'rejected connections recorded'
#   at t/maxconns.t line 38.
#     '0'
#         >
#     '1'
t/maxconns.t ................ Dubious, test returned 1 (wstat 256, 0x100)

```

```
t/watcher.t ................. 3/30 
#   Failed test 'logged new connection'
#   at t/watcher.t line 92.
#                   'ts=1634230984.805284 gid=2 type=conn_new rip=127.0.0.1 rport=53462 transport=tcp cfd=34
# '
#     doesn't match '(?^:ts=\d+\.\d+\ gid=\d+ type=conn_new .+ transport=local)'

#   Failed test 'logged closed connection due to client disconnect'
#   at t/watcher.t line 95.
#                   'ts=1634230984.807135 gid=3 type=conn_close rip=127.0.0.1 rport=53462 transport=tcp reason=normal cfd=34
# '
#     doesn't match '(?^:ts=\d+\.\d+\ gid=\d+ type=conn_close .+ transport=local reason=normal)'

#   Failed test 'logged new connection'
#   at t/watcher.t line 102.
#                   'ts=1634230984.809056 gid=4 type=conn_new rip=127.0.0.1 rport=53464 transport=tcp cfd=34
# '
#     doesn't match '(?^:ts=\d+\.\d+\ gid=\d+ type=conn_new .+ transport=local)'

#   Failed test 'logged closed connection due to client protocol error'
#   at t/watcher.t line 105.
#                   'ts=1634230984.810899 gid=5 type=conn_close rip=127.0.0.1 rport=53464 transport=tcp reason=error cfd=34
# '
#     doesn't match '(?^:ts=\d+\.\d+\ gid=\d+ type=conn_close .+ transport=local reason=error)'
t/watcher.t ................. 29/30 # Looks like you failed 4 tests of 30.
t/watcher.t ................. Dubious, test returned 4 (wstat 1024, 0x400)
```

* `maxconns.t`: The connection rejection logic writes the rejection message directly to the client socket without wrapping with TLS. I suspect this caused some issues with the SSL socket client, because refreshing the stats right before the check lets the test pass since the server itself is doing the right thing.
* `watcher.t`: Over TLS, we use a TCP transport rather than a local (Unix) transport.

[CI currently only runs the basic TLS tests](https://github.com/memcached/memcached/blob/master/.github/workflows/ci.yml#L21) so I guess that's why these went unnoticed. Happy to add a `make test_tls` step there if that's desirable.

### DIFF
--- a/t/lib/MemcachedTest.pm
+++ b/t/lib/MemcachedTest.pm
@@ -311,7 +311,13 @@ sub new_memcached {
             $args .= " -U $udpport";
         }
         if ($ssl_enabled) {
-            $args .= " -Z -o ssl_chain_cert=$server_crt -o ssl_key=$server_key";
+            $args .= " -Z";
+            if ($args !~ /-o ssl_chain_cert=(\S+)/) {
+                $args .= " -o ssl_chain_cert=$server_crt";
+            }
+            if ($args !~ /-o ssl_key=(\S+)/) {
+                $args .= " -o ssl_key=$server_key";
+            }
         }
     } elsif ($args !~ /-s (\S+)/) {
         my $num = @unixsockets;
--- a/t/maxconns.t
+++ b/t/maxconns.t
@@ -47,6 +47,8 @@ sub test_maxconns {
     for my $s (@sockets) {
         $s->close();
     }
+
+    $stats = mem_stats($stat_sock);
     cmp_ok($stats->{rejected_connections}, '>', '1', 'rejected connections recorded');
     $server->stop;
     $stat_sock->close();
--- a/t/ssl_cert_refresh.t
+++ b/t/ssl_cert_refresh.t
@@ -3,6 +3,7 @@
 use strict;
 use warnings;
 use File::Copy;
+use File::Temp;
 use Test::More;
 use FindBin qw($Bin);
 use lib "$Bin/lib";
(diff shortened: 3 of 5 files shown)


LINKIWI:
That makes sense. Yeah `test_tls` probably takes too long to run in Github CI.

This patch allowed `make test_tls` to complete without errors at least once, but after doing some more testing it seems like there are either some race conditions or nondeterminism since I'm not able to get it to reliably pass the same as `make test`. I'll spend some more time to see if I can track down all the remaining issues and come back to this.

dormando:
feel like I worked around the tcp port issue by making it use unix sockets for most tests.... haven't given it any thought recently.

Should I review/merge up what you have so far?

LINKIWI:
Yeah, the non-TLS test path uses Unix sockets which avoids that race condition, but unfortunately it seems like the Perl SSL client library only supports originating TLS to TCP endpoints.

I just rebased against `next` and resolved conflicts. I think the change is ready for review. I don't plan to make any additional changes so feel free to merge if you think it's good to go.

dormando:
out in 1.6.13

PR #834, opened 2021-11-09
Improved TLS v1.3 eligibility check for compatibility with non-OpenSSL libraries

### LINKED ISSUE #833: 1.6.12 fails to build on FreeBSD when using libressl < 3.4
As per title, 1.6.12 fails to build with libressl < 3.4, as the checks for enabling TLS1_3 is checking the openssl version. For some reason, libressl has the OPENSSL_VERSION_NUMBER set so your check enables TLS1_3. By "cheating" a bit, and checking for LIBRESSL_VERSION_NUMBER >= 03040000fL (not 100% sure that's the right number to check for, as I don't know when they implemented the public interfaces for TLS1_3, but 3.4.1 does at least have them), the build works with a disabled TLS1_3..

Attached is my FreeBSD patches (remove .txt for paching to work automagically)
[patch-memcached.c.txt](https://github.com/memcached/memcached/files/7493179/patch-memcached.c.txt)
[patch-tls.c.txt](https://github.com/memcached/memcached/files/7493180/patch-tls.c.txt)
..

LINKIWI:
For compatibility with LibreSSL (and possibly others), change the preprocessor directive to check whether `TLS1_3_VERSION` is defined rather than checking for a specific minimum OpenSSL version.

Fixes #833.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -4047,7 +4047,7 @@ static void usage(void) {
     printf("   - ssl_session_cache:   enable server-side SSL session cache, to support session\n"
            "                          resumption\n"
            "   - ssl_min_version:     minimum protocol version to accept (default: %s)\n"
-#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+#if defined(TLS1_3_VERSION)
            "                          valid values are 0(%s), 1(%s), 2(%s), or 3(%s).\n",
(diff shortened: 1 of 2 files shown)


Nuadh (approved):

dormando:
Simple! Appreciate this, thanks.

dormando:
out in 1.6.13

PR #836, opened 2021-11-15
extstore: avoid looping IO queues on submission

dormando:
with low (ie; 1) IO threads, and a long IO depth (10k+) that might
suddenly appear during a disk hiccup, we can cause a slowdown as each
worker thread submission walks the entire queue.

This still walks queues under a lock, but now only what IO's were
submitted.

Should fix or improve the situation noted in #829

### DIFF
--- a/extstore.c
+++ b/extstore.c
@@ -78,6 +78,7 @@ typedef struct {
     pthread_mutex_t mutex;
     pthread_cond_t cond;
     obj_io *queue;
+    obj_io *queue_tail;
     store_engine *e;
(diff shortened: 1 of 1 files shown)


dormando:
hmmm... think there's actually no reason to do the walk while holding the io thread lock. it's almost always going to be a few low number but meh. just need to hold the depth in a temp var since the rest are already temp vars.

dormando:
made that update just now. it iterates the objects before even getting the thread to submit to... should be a decent speedup overall.

PR #838, opened 2021-12-04
proxy: fixes for liburing support

dormando:
Starting out by linking liburing's timeouts, which are hrtimers and shouldn't be infinitely scalable. I did some quick tests and it doesn't seem to be impacting performance at ~300krps, which is fine for a first pass.

This PR will track a few more changes:

- [x] all timeouts for liburing mode (note: double check connect timeouts?)
- [x] periodic timeout updater in uring mode (sigh)
~~restructure uring mode to avoid sqe/cqe starvation at high loads~~ - making this a separate branch once I do more experiments/research.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -167,6 +167,11 @@ struct proxy_timeouts {
     struct timeval connect;
     struct timeval retry; // wait time before retrying a dead backend
     struct timeval read;
+#ifdef HAVE_LIBURING
+    struct __kernel_timespec connect_ur;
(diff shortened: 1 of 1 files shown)


dormando:
uring vs non-uring in a basic mctester loadtest shows about 10% throughput improvement. haven't checked latency yet.

once I do general write combining I'll add uring support for writes, which should hopefully give a higher overall improvement as everything is batched at that point.

dormando:
out in 1.6.13

PR #843, opened 2021-12-14
little mem leak fixes.

devnexen:


dormando:
Thanks!

Now I already regret inline-vendoring mcmc :) The upstream for that is: https://github.com/dormando/mcmc

dormando:
I guess I'm just going to close this; that free happens right before the program exits and the other change was to a vendored file.

thanks though!

Issue #849, opened 2022-01-11
Multikey get sometimes fails with ERROR - 1.6.12


adam-ah:
**Describe the bug**
On 1.6.12 when I'm fetching values for a long list of keys, sometimes (on the _n_ th run, but typically not on the first run), it returns ERROR instead of the values. For a short list of keys or single key, the operation completes as normal.
Using docker latest image.

EDIT: 
**Affected server versions:** 1.6.12
**_NOT_ affected server versions:** 1.6.1, 1.5.22

**To Reproduce**

EDIT:
Minimal example to reproduce with pymemcache (NB., _several runs needed_, sometimes it will work. The larger the batch size, the more likely it will fail)
```
from pymemcache.client.base import Client

server = 'SERVER'

client = Client(server)
client.set(f'somekey', '012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789')
many_keys = [f'somekey' for i in range(10000)]
client.get_many(many_keys)

# Sometimes fails with 
# pymemcache.exceptions.MemcacheUnknownCommandError: b'get'
```
Digging into the source code of pymemcache, the server responds with an 'ERROR' in `line` on https://github.com/pinterest/pymemcache/blob/master/pymemcache/client/base.py#L1026 which is rethrown when processing the response on L1030.

*[Obsolete repro with telnet]* Minimal example using telnet as below (removed a part of the first response for brevity).
The first multikey get suceeds as normal, the seconds fails with an ERROR.
```
telnet SERVER 11211
Trying SERVER...
Connected to SERVER.
Escape character is '^]'.
set somekey 0 0 100
0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
STORED
[...]

PR #850, opened 2022-01-12
core: fix use-after-free for text multigets

dormando:
Reported in #849 - this fixes copying a read buffer after freeing the
original read buffer.

This didn't matter for years since the cache code didn't touch the
buffer, but recently it can reuse the first 8 bytes as a pointer to the
internal freelist. Thus in some situations where large reads happen the
command can get corrupted, returning an unhelpful "ERROR" to the end
user.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -440,8 +440,8 @@ bool rbuf_switch_to_malloc(conn *c) {
     if (!tmp)
         return false;
 
-    do_cache_free(c->thread->rbuf_cache, c->rbuf);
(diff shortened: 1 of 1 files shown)


dormando:
Note to self: also fix cache.c to use `STAILQ_INSERT_HEAD` since we pop from head and LIFO is better for these kinds of caches.

dormando:
More note to self: in the tested case here the code is switching to a malloc'ed read buffer while the default read buffer still has plenty of space in it (~10k left). Might be an easy fix.

Issue #852, opened 2022-01-19
Incorrect error message when using the -l option


smukil:
[Note: This is a minor bug which was misleading me during a dev session]

Memcached can be started with configurable addresses to listen on, using the "-l" option. Eg:
`$> ./memcached -l 127.0.2.1:22122`

If the port `22122` is available, things run smoothly. However, if the port is unavailable, we see the following error message:
`failed to listen on TCP port 11211`

Note that the port number is incorrect and is the default port number `11211` instead of the configured `22122`.

**To Reproduce**
To reproduce, start memcached on the same listen address twice. The second one will repro the error.

**System Information**
Reproducible on any supported Distro.

**Detail (please include!)**
This is because the `server_sockets()` function treats `settings.port` and `settings.inter` differently. If "-l" is provided, `settings.inter` contains the correct port number, whereas `settings.port` the default port, and `settings.port` is used to print the error info, hence printing the incorrect port number.

To fix, track the port information correctly. However, memcached allows listening on multiple interfaces (`./memcached -l <ip1>:<port1>,<ip2>:<port2>`), which means we must print the right conflicting port as part of the error.

dormando:
Nice catch, thanks!

smukil:
@dormando A quick question. When multiple interfaces are provided with the `-l` option, I'm curious why the `server_sockets` code attempts to continue binding with sockets if one of them hits a EADDRINUSE.

https://github.com/memcached/memcached/blob/master/memcached.c#L3702

If any one of them fails, memcached still won't startup as per:
https://github.com/memcached/memcached/blob/master/memcached.c#L6032

Shouldn't `server_sockets` return as soon as it hits the first error?

dormando:
Amazingly I didn't actually write that code so I don't remember. I remember the `server_socket` function itself being tricky and having to actually look through the returned addrinfo structures until one works. EADDRINUSE is specifically exempted while working through the addrinfo structures, which is why the behavior falls through like that. A lot of other errors there do an early return instead.

I don't remember the exact conditions that happens offhand, some ipv6 thing I'm forgetting. You would have to validate/understand exactly why EADDRINUSE is being exempted there before changing the code to bail early.

smukil:
Ok that makes sense. In order to not risk breaking a bunch of cases, I'll leave this code as is while sending in a fix.

[...]

PR #854, opened 2022-01-21
Fix error message on conflicting ports while using '-l'

smukil:
Fix for [https://github.com/memcached/memcached/issues/852]

Memcached can be started with configurable addresses to listen on,
using the "-l" option. Eg:
    $> ./memcached -l 127.0.2.1:22122

If the port 22122 is available, things run smoothly. However, if
the port is unavailable, we see the following error message:
    "failed to listen on TCP port 11211"

Note that the port number is incorrect and is the default port
number 11211 instead of the configured 22122.

This is because the server_sockets() function treats settings.port
and settings.inter differently. If "-l" is provided, settings.inter
contains the correct port number, whereas settings.port the default
port, and settings.port is used to print the error info, hence
printing the incorrect port number.

Since multiple interfaces can be provided with '-l', this patch prints
all of them on a bind error, to maintain code simplicity. The
server_sockets() function is also adjusted to remember the first bind
error's 'errno' that occurs, as the caller of server_sockets() relies
on the errno to print the error message.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -3643,6 +3643,8 @@ static int server_sockets(int port, enum network_transport transport,
             fprintf(stderr, "Failed to allocate memory for parsing server interface string\n");
             return 1;
         }
+        // If we encounter any failure, preserve the first errno for the caller.
+        int errno_save = 0;
         for (char *p = strtok_r(list, ";,", &b);
             p != NULL;
             p = strtok_r(NULL, ";,", &b)) {
@@ -3700,8 +3702,10 @@ static int server_sockets(int port, enum network_transport transport,
                 p = NULL;
             }
             ret |= server_socket(p, the_port, transport, portnumber_file, ssl_enabled);
+            if (ret != 0 && errno_save == 0) errno_save = errno;
         }
         free(list);
+        errno = errno_save;
         return ret;
     }
 }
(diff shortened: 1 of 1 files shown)


memcached.c
```diff
@@ -3643,6 +3643,8 @@ static int server_sockets(int port, enum network_transport transport,
             fprintf(stderr, "Failed to allocate memory for parsing server interface string\n");
             return 1;
         }
+        // If we encounter any failure, preserve the first errno for the caller.
```
smukil, line 3646: [Note] Preserving the first error here since the following case can occur:
Considering 2 interfaces:
./memcached  -l <IF1>,<IF2>

If memcached fails to bind to <IF1>, but succeeds binding to <IF2>, the error message will look like so:
"failed to listen on one of interface(s) <IF1>,<IF2>: Success"

If we instead preserve the first error seen, we'll always see a valid error:
"failed to listen on one of interface(s) <IF1>,<IF2>: address already in use"

dormando:
Thanks! I'll review this as soon as I can. At a glance it looks simple enough. I'll try to resist temptation of changing that loop further since it's probably a compat nightmare.

dormando:
Hey,

I think the patch itself is fine. I'm going to hold off the urge to print a nicer specific error, but I did notice your e-mail is wrong on the commit. You should probably fix that and re-push before I merge :)

smukil:
Good catch! I just re-pushed it with the right email.

Also, I also did consider printing a better error message as well, i.e. printing the specific port that had an error, but the trade off was that it would've increased the complexity of the code for a minor gain, and hence I arrived at this.

dormando:
Thanks! And yeah, exactly. It's a lot of work to validate a deeper change unfortunately, and I'm okay with this.

dormando:
released in 1.6.14

Issue #855, opened 2022-01-24
Runaway CPU usage on high memory servers (>1B items)


gvengel:
After upgrading from memcached 1.5.6 on Ubuntu 18.04 to memcached 1.5.22 on Ubuntu 20.04, I started having issues where everything would be fine during initial cache fill, but then CPU usage would go crazy around 750GB of cached data. After a bunch of debugging, I finally figured out the tipping point was actually when switching from hashpower=31 to 32. Armed with information, I looked through the changes from 1.5.6 to 1.5.22, and discovered that dae3191 introduced an integer overflow which breaks the hash table.

With dae3191 reverted, performance is restored to normal, even with `hashpower=32` and greater than two billion items. 

**To Reproduce**
* Run memcached >= 1.5.14 with `-o hashpower=31`
* Generate a moderate amount of load with memslap (or similar)
* Everything is fine
* Rerun memcached `-o hashpower=32`
* Generate a moderate amount of load with memslap
* memcached will consume all available CPU

**System Information**
 - OS/Distro: Ubuntu Server
 - Version of OS/distro: 20.04
 - Version of memcached: 1.5.22
 - Hardware detail: AMD 7702P, 1TB DDR4

dormando:
Thanks for the report and the fix/test! Golden! :)

Note mostly for me but maybe also for you: I thought I had capped the hashpower value to not go over 30 or 31 (and also tested it at 31), so I need to dig into why that failed. It's been long enough that I forgot what I did. There were other things that may start to fail or act odd when the hash table gets near or above 2^32 so I wanted to simply avoid it for a while. One major issue is surprise at runtime, as it allocates a new hashtable, and migrates before dropping the old one: that ends up being a lot of memory at the high end and causes people to OOM. That has been giving people a really bad time in production.

gvengel:
Yeah, while debugging I realized how expensive increasing the hashpower could be. All my servers now start with hashpower preset to its expected max value to prevent runtime expansion. 

For what it's worth, I was running at hashpower 31 on memcached 1.5.6 for many years without issue, and haven't had any problems on my patched 1.5.22 build. If there are more bugs lurking out there, they must be subtle. :)

Memcache is an absolute workhorse. Thanks for being the bedrock we build on. ❤️

dormando:
Great, thanks :)

I see the bug now, but haven't been able to reproduce the CPU runaway locally. A backtrace may have helped here. There's only one place where hashsize should actually cast through to a 64bit number.

[...]

PR #856, opened 2022-01-24
Fix integer overflow in hashsize calculation

gvengel:
Fix for #855. Rather than simply revert dae3191, I decided it might be safer to use `uint64_t` directly, since the bad comment on the ub4 typedef was probably what caused the bug to be written in the first place.

I tried to figure out a way to test the bad result from hashsize directly, but best I could come up with was to check the `hash_bytes` metric. The problem is I believe this may vary by CPU architecture. While the tests try to accommodate both 64 and 32 bit systems, I only actually validated all the tests pass on 64 bit. It's also possible the perl architecture doesn't match the memcached architecture, if it's being cross compiled, etc. Which is to say, I don't love these tests, but hopefully they are close enough.


gvengel:
I fired up a 32-bit VM and discovered that my hunch the tests were bad, was correct. 😅 That led me to poke around a bit more and discover memcached's pointer size is actually exposed as a stat, which can be used to properly calculate `hash_bytes`.

dormando:
Okay I understand the situation I think. Think this needs a few more tweaks. `expand_bucket` and `assoc_iterator->bucket` are both `unsigned int`. I think `expand_bucket` might be fine since that's looking for `hashsize(hashpower-1)` but the iterator would overflow? So probably safest to keep the ub4 type and make expand/bucket be ub4's? Either that or lower `HASHPOWER_MAX` to 31 :)

From looking at the code from the `HASHPOWER_MAX` commit, I probably intended to set that to 31, and also put the max check after the multiplier instead of before. Lots of weird mistakes actually, considering the next commit in the chain fixes a signed int bug labeled as "Crash in 31 -> 32 expansion". Weird.

dormando:
updated the original comment a few times. I'm not drunk I'm just tired :)

gvengel:
Your call. Since I run at hashpower=31, lowering `HASHPOWER_MAX` wouldn't make any difference to me. Given it took years for anyone to notice things were very broken at hashpower=31, I doubt there are many users running with 32, if any.

dormando:
That's kind of the odd thing. I know folks who use `hashpower=31`. Locally I loaded a few hundred million items and ran a get/set/delete test for 10 minutes. The bug doesn't look like it should cause it to spin out of control, it looks like it probably triggers an expansion to hashpower=32 early.

I might hold this a little more so I can note what the exact failure mode was with the fix. If you have a backtrace from a hung instance (or more detail about the repro case, the issue was vague) I could nail that down faster.

gvengel:
Oh, interesting. Okay, it will take me a bit to get my test box setup again, but I should be able to get your more details by tomorrow.

dormando:
Apologies if the crash/hang is super obvious and I don't see it :) I didn't spend a ton of time looking as there're a bunch of incoming PR's right now. I remember triggering a bug once a long time ago which required constantly trying to resize the hash table (failing to malloc the new one due to OOM) but I never had a repro so I never fixed it. It might end up being a real but super rare bug with hash table expansion so if it ends up being that a repro would be useful.

If not, the fix seems obvious enough as of this PR + my notes, I just like to fully understand a crash/hang before fixing it.

gvengel:
Okay, sorry, I mislead you a bit here. I did the triage on this a month ago, and mixed a couple things up in my head since then. Getting a server setup to reproduce the problem again jogged my memory. 

The run away CPU does not actually happen at hashpower=31, but rather at 32. I can reliably reproduce that on my test box now. I believe the mix up was that in production when there are enough items to trigger expansion that happens early, so the tipping point is really 31, since hitting it jumps you to 32. Does that make sense? If not, I can setup a test for making the expansion happen instead of setting `-o hashpower=32` on the command line.

This is what I'm running for the test:

```
$ memcached -V
memcached 1.5.22
$ memcached -o hashpower=32 &
$ memaslap -s 127.0.0.1:11211 -S 5s
```

Then looking at top or similar you'll see each memcached thread taking 100% of a CPU core. You can add a `-t` with more threads to make it obvious, but it should be visible with the default.

Let me know if you still can't reproduce it, and I'll send you whatever other traces you like.
```
Reading symbols from memcached...
Reading symbols from /usr/lib/debug/.build-id/24/877ba5dc837cc1579a13b3842708990a6a6559.debug...
(gdb) r
Starting program: /usr/bin/memcached -o hashpower=32 -u memcache
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff79ae700 (LWP 598028)]
[New Thread 0x7ffff7184700 (LWP 598029)]
[New Thread 0x7ffff6983700 (LWP 598030)]
[New Thread 0x7ffff6161700 (LWP 598031)]
[New Thread 0x7ffff593f700 (LWP 598032)]
[New Thread 0x7ffff50fc700 (LWP 598033)]
[New Thread 0x7ffff48fb700 (LWP 598034)]
[New Thread 0x7fffe7fff700 (LWP 598035)]
[New Thread 0x7fffe77fe700 (LWP 598036)]
^Z
Thread 1 "memcached" received signal SIGTSTP, Stopped (user).
0x00007ffff7dc65ce in epoll_wait (epfd=3, events=0x5555555a0300, maxevents=32, timeout=737) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
30      ../sysdeps/unix/sysv/linux/epoll_wait.c: No such file or directory.
(gdb) info threads
  Id   Target Id                                      Frame
* 1    Thread 0x7ffff79c3740 (LWP 598024) "memcached" 0x00007ffff7dc65ce in epoll_wait (epfd=3, events=0x5555555a0300, maxevents=32, timeout=737)
    at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
  2    Thread 0x7ffff79ae700 (LWP 598028) "memcached" 0x00007ffff7d843bf in __GI___clock_nanosleep (clock_id=clock_id@entry=0, flags=flags@entry=0, req=req@entry=0x7ffff79ade70,
    rem=rem@entry=0x0) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  3    Thread 0x7ffff7184700 (LWP 598029) "memcached" __memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:72
  4    Thread 0x7ffff6983700 (LWP 598030) "memcached" __memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:71
  5    Thread 0x7ffff6161700 (LWP 598031) "memcached" __memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:164
  6    Thread 0x7ffff593f700 (LWP 598032) "memcached" __memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:153
  7    Thread 0x7ffff50fc700 (LWP 598033) "memcached" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x5555555951a8 <maintenance_cond+40>)
    at ../sysdeps/nptl/futex-internal.h:183
  8    Thread 0x7ffff48fb700 (LWP 598034) "memcached" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555596748 <lru_crawler_cond+40>)
    at ../sysdeps/nptl/futex-internal.h:183
  9    Thread 0x7fffe7fff700 (LWP 598035) "memcached" 0x00007ffff7d843bf in __GI___clock_nanosleep (clock_id=clock_id@entry=0, flags=flags@entry=0, req=req@entry=0x7fffe7ffeb10,
    rem=rem@entry=0x0) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  10   Thread 0x7fffe77fe700 (LWP 598036) "memcached" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555587928 <slab_rebalance_cond+40>)
    at ../sysdeps/nptl/futex-internal.h:183
(gdb) thread 3
[Switching to thread 3 (Thread 0x7ffff7184700 (LWP 598029))]
#0  __memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:72
72      ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S: No such file or directory.
(gdb) bt
#0  __memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:72
#1  0x00005555555719f2 in assoc_find (key=key@entry=0x7ffff001ebc4 "\363V\030\020\020\020\020\020YpKMpkL8lQFi4xtuZfmJve.nX.9ROovxcDHPxdXg1mM5hE.FtkYmMWXt", nkey=nkey@entry=64,
    hv=hv@entry=2169723189) at assoc.c:88
#2  0x0000555555570f68 in do_item_get (key=key@entry=0x7ffff001ebc4 "\363V\030\020\020\020\020\020YpKMpkL8lQFi4xtuZfmJve.nX.9ROovxcDHPxdXg1mM5hE.FtkYmMWXt", nkey=nkey@entry=64,
    hv=hv@entry=2169723189, c=c@entry=0x7ffff001e980, do_update=do_update@entry=true) at items.c:956
#3  0x000055555557297c in item_get (key=0x7ffff001ebc4 "\363V\030\020\020\020\020\020YpKMpkL8lQFi4xtuZfmJve.nX.9ROovxcDHPxdXg1mM5hE.FtkYmMWXt", nkey=64, c=0x7ffff001e980,
    do_update=<optimized out>) at thread.c:659
#4  0x000055555556052b in limited_get (do_update=true, should_touch=false, exptime=0, c=0x7ffff001e980, nkey=64, key=<optimized out>) at memcached.c:3839
#5  process_get_command (c=0x7ffff001e980, tokens=0x7ffff7183bf0, return_cas=false, should_touch=false, ntokens=<optimized out>) at memcached.c:4148
#6  0x0000555555564010 in process_command (command=0x7ffff001ebc0 "get", c=0x7ffff001e980) at memcached.c:5793
#7  try_read_command_ascii (c=0x7ffff001e980) at memcached.c:6341
#8  0x0000555555567872 in drive_machine (c=<optimized out>) at memcached.c:6822
#9  0x00007ffff7ef713f in ?? () from /lib/x86_64-linux-gnu/libevent-2.1.so.7
#10 0x00007ffff7ef787f in event_base_loop () from /lib/x86_64-linux-gnu/libevent-2.1.so.7
#11 0x0000555555571d27 in worker_libevent (arg=0x5555555a2fb0) at thread.c:455
#12 0x00007ffff7e9f609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#13 0x00007ffff7dc6293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb)
```

dormando:
Awesome! That lines up with my understanding of the code audit that I noted earlier, with the overflow causing hashpower=32 to happen too soon. Since you're able to reproduce it starting from hashpower=32 that means the bug isn't in the expansion work too, which is good to know. I may have not actually tried =32 in my quick testing though it felt like I did.. so I'll start there.

I'll repro when I get a chance and make a decision on which fix I'd prefer here. Thanks for circling back so soon!

dormando:
Just eyeballing it.. it looks like the 32bit overflow is causing `hashmask()` to return 0, which is the obvious bug now at hashpower=32. I'll take a quick stab at checking if we can't leave that mask value at 32bit for the sake of 32bit systems, but expanding it or lowering the MAX to 31 are probably both fine.

dormando:
I added a few more repairs for bucket walking/expansion and removed the tests. The tests weren't actually testing the failure condition and allocated tons of memory so none of the builders can run it successfully :) Not sure how to properly test this if at all..

HASHPOWER_MAX stays at 32 since the actual key hash value is 32bit so it won't be able to address anything larger. This is probably fine for a few more years...

Thanks!

gvengel:
Great, thanks. 👍

dormando:
released in 1.6.14

PR #857, opened 2022-01-25
proxy: backend write combining

dormando:
previously individual requests for the same backend would use multiple
syscalls. This batches them.

This is also prep work for write syscall offloading for uring mode.


dormando:
skipping the io_uring part for now.. better to focus on stabilizing the code.

PR #858, opened 2022-01-25
large pages support update the detection.

devnexen:


dormando:
Thanks! Any more detail on what RH-based means? Any particular version this started in/etc?

memcached.c
```diff
@@ -4242,13 +4242,32 @@ static int enable_large_pages(void) {
...
-    struct stat st;
-    int ret = stat("/sys/kernel/mm/transparent_hugepage/enabled", &st);
-    if (ret || !(st.st_mode & S_IFREG)) {
+    /* RH based systems possibly uses a different path */
+    static const char *mm_thp_paths[2] = {
+        "/sys/kernel/mm/transparent_hugepage/enabled",
+        "/sys/kernel/mm/redhat_transparent_hugepage/enabled",
+    };
+
+    size_t pathlen = sizeof(mm_thp_paths) / sizeof(mm_thp_paths[0]);
```
dormando, line 4251: line has a confusing smell to it. add terminating NULL/empty string and loop until empty string found instead of pathlen?

memcached.c
```diff
@@ -4242,13 +4242,32 @@ static int enable_large_pages(void) {
...
+    }
+
+    if (pfd == -1) {
         fprintf(stderr, "Transparent huge pages support not detected.\n");
         fprintf(stderr, "Will use default page size.\n");
         return -1;
     }
+    ssize_t rd = read(pfd, thpb, sizeof(thpb));
+    close(pfd);
+    if (rd <= 0 || strstr(thpb, "[never]")) {
```
dormando, line 4266: Lines like this tend to set off an army of student static analyzers. Need to use a string function with bounding, or at least manually insert a null terminator at the end of thpb before running strstr against it.

memcached.c
devnexen: it works here if I print out I have 2 entries.

memcached.c
devnexen: the buffer is initialized but I can change it a bit.

memcached.c
dormando: I believe that it works, but the common pattern in memcached code is until-null-entry. Then you also don't need the tracking variable. Or at least all the new stuff I write is.

memcached.c
```diff
@@ -4242,13 +4242,38 @@ static int enable_large_pages(void) {
...
         return -1;
     }
+    ssize_t rd = read(pfd, thpb, sizeof(thpb));
+    close(pfd);
+    if (rd <= 0) {
+        fprintf(stderr, "Transparent huge pages could not read the configuration.\n");
+        fprintf(stderr, "Will use default page size.\n");
+        return -1;
+    }
+    thpb[rd] = 0;
```
dormando, line 4271: Good. Think that fixes it. sizeof(thpb)-1 would be similar, fwiw. You could generate a 128 byte read and end up without a null zero without your fix or this.

memcached.c
devnexen: alright let s go for consistency.

memcached.c
```diff
@@ -4242,13 +4242,38 @@ static int enable_large_pages(void) {
...
+    /* RH based systems possibly uses a different path */
+    static const char *mm_thp_paths[] = {
+        "/sys/kernel/mm/transparent_hugepage/enabled",
+        "/sys/kernel/mm/redhat_transparent_hugepage/enabled",
+        NULL
+    };
+
+    char thpb[128] = {0};
+    int pfd = -1;
+    for (const char **p = mm_thp_paths; *p; p ++) {
```
dormando, line 4254: thanks! hopefully last style nitpick: `p ++` - shouldn't be a space before ++

dormando:
Pushed to next, thanks!

dormando:
released in 1.6.14

PR #862, opened 2022-01-28
proxy: audit TODO/FIXME lines. mark some for v2

dormando:
Marks TODO/FIXME items that can be done after MVP/V1 status is achieved. This PR also attempts to address most of the non-v2 TODO/FIXME items.

"v2" just means "after v1". Not at all tied to a specific version or any
order.

This cuts us down from 170 items to... 80...

Most of what was marked v2 are optimizations or require some refactoring for future features. Most of what's left are missing error handling and obvious bugs/limitations. I'll fix all of these before doing more serious edge case testing, since after looking through them a lot of these are pretty obvious.


dormando:
only 63 more TODO/FIXME lines to go... not much time for code yesterday/today but at least today fixes a fairly bad bug.

dormando:
42 items left. long slog is long.

dormando:
31 bottles of bugs on the wall. 31 bottles of bugs. take one down, pass it around... and everyone's grossed out cuz it's bugs.

PR #869, opened 2022-02-11
Proxy observability improvements

dormando:
More MVP grind:

- adds a bunch more stats to `stats proxy` - tracking how many of each type of command was received.
```
stats proxy
STAT user_example 0
STAT user_another 0
STAT cmd_mg 0
STAT cmd_ms 0
STAT cmd_md 0
STAT cmd_mn 0
STAT cmd_ma 0
STAT cmd_me 0
STAT cmd_get 0
STAT cmd_gat 0
STAT cmd_set 0
STAT cmd_add 0
STAT cmd_cas 0
STAT cmd_gets 0
STAT cmd_gats 0
STAT cmd_incr 0
STAT cmd_decr 0
STAT cmd_touch 0
STAT cmd_append 0
STAT cmd_prepend 0
STAT cmd_delete 0
STAT cmd_replace 0
END
```
- adds `watch proxycmds|proxyevents|proxyuser` watcher endpoints.
- adds `mcp.log()` lua interface to logging (lands in proxyuser)
- adds some logs to backend errors, these look like:
```
ts=1644553677.231725 gid=5 type=proxy_backend error=connecting ip=127.2.1.3 port=11213
ts=1644553677.231738 gid=6 type=proxy_backend error=connecting ip=127.1.1.3 port=11212
ts=1644553677.231746 gid=7 type=proxy_backend error=connecting ip=127.3.1.3 port=11214
```

More specific errors are possible but will take more effort to get the interfaces to align. the errno should be added to the existing line at some point.

Before merging, also looking at:
- [ ] `mcp.slow_log_threshold(0.5)` - set a timing threshold for sending logs to `proxycmds` log stream.
- [ ] might be able to add the backend ip:port to the slow log stream
- [ ] also looking at a short string or number to "tag" a request with (which can be zone/route/anything), which would be sent in the slow/error log stream.
- don't think I can get detailed errors into the slow log stream (which we could also trigger on error) without significant work... but I tried.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -53,6 +53,11 @@
     c->thread->stats.stat += amount; \
     pthread_mutex_unlock(&c->thread->stats.mutex); \
 }
+#define WSTAT_DECR(c, stat, amount) { \
+    pthread_mutex_lock(&c->thread->stats.mutex); \
+    c->thread->stats.stat -= amount; \
+    pthread_mutex_unlock(&c->thread->stats.mutex); \
+}
 #define STAT_L(ctx) pthread_mutex_lock(&ctx->stats_lock);
 #define STAT_UL(ctx) pthread_mutex_unlock(&ctx->stats_lock);
 #define STAT_INCR(ctx, stat, amount) { \
@@ -1212,31 +1217,7 @@ void proxy_submit_cb(io_queue_t *q) {
 }
 
 void proxy_complete_cb(io_queue_t *q) {
-    /*
-    io_pending_proxy_t *p = q->stack_ctx;
-    q->stack_ctx = NULL;
-
-    while (p) {
-        io_pending_proxy_t *next = p->next;
-        mc_resp *resp = p->resp;
-        lua_State *Lc = p->coro;
-
-        // in order to resume we need to remove the objects that were
-        // originally returned
-        // what's currently on the top of the stack is what we want to keep.
-        lua_rotate(Lc, 1, 1);
-        // We kept the original results from the yield so lua would not
-        // collect them in the meantime. We can drop those now.
-        lua_settop(Lc, 1);
-
-        proxy_run_coroutine(Lc, resp, p, NULL);
-
-        // don't need to flatten main thread here, since the coro is gone.
-
-        p = next;
-    }
-    return;
-    */
+    // empty/unused.
 }
 
 // called from worker thread after an individual IO has been returned back to
@@ -1259,7 +1240,7 @@ void proxy_return_cb(io_pending_t *pending) {
         // p can be freed/changed from the call below, so fetch the queue now.
         io_queue_t *q = conn_io_queue_get(p->c, p->io_queue_type);
         conn *c = p->c;
-        proxy_run_coroutine(Lc, p->resp, p, NULL);
+        proxy_run_coroutine(Lc, p->resp, p, c);
 
         q->count--;
         if (q->count == 0) {
@@ -1347,6 +1328,7 @@ void proxy_cleanup_conn(conn *c) {
     lua_State *L = thr->L;
     luaL_unref(L, LUA_REGISTRYINDEX, c->proxy_coro_ref);
     c->proxy_coro_ref = 0;
+    WSTAT_DECR(c, proxy_req_active, 1);
 }
 
 // we buffered a SET of some kind.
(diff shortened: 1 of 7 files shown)


proto_proxy.c
```diff
@@ -1041,17 +1099,22 @@ static int proxy_thread_loadconf(LIBEVENT_THREAD *thr) {
         return -1;
     }
 
+    // seed per-thread stats.
+    if (thr->proxy_int_stats == NULL) {
+        thr->proxy_int_stats = calloc(1, sizeof(struct proxy_int_stats));
```
smukil, line 1104: Probably very unlikely to happen since this code runs at initialization time, but should we be checking for failed malloc/calloc's here and fail explicitly?

proto_proxy.c
```diff
@@ -2448,18 +2494,18 @@ static void proxy_backend_handler(const int fd, const short which, void *arg) {
...
                         break; // sit on epoll again.
                     } else {
-                        _reset_bad_backend(be);
+                        _reset_bad_backend(be, P_BE_FAIL_READING);
                         return;
                     }
                 }
             } else if (res == -1) {
-                _reset_bad_backend(be);
+                _reset_bad_backend(be, P_BE_FAIL_PARSING);
```
smukil, line 2508: Does this mean the command was improperly formatted?

I'm guessing this will be useful if there's a command received by the proxy and then transformed into a different command(s) via a Lua hook function, and this transformed command is incorrectly formatted. And this could only happen if there's a bug in the Lua function. Am I understanding it correctly?

proto_proxy.c
```diff
@@ -2605,6 +2648,16 @@ static void proxy_process_command(conn *c, char *command, size_t cmdlen, bool mu
         return;
     }
 
+    // Count requests handled by proxy vs local.
+    // Also batch the counts down this far so we can lock once for the active
+    // counter instead of twice.
+    struct proxy_int_stats *istats = c->thread->proxy_int_stats;
+    WSTAT_L(c->thread);
```
smukil, line 2655: You mentioned in another commit that atomics are slower than locks in your benchmarking.
But curious if this causes a regression since we're adding a new lock/unlock on every command.

proto_proxy.c
```diff
@@ -2605,6 +2648,16 @@ static void proxy_process_command(conn *c, char *command, size_t cmdlen, bool mu
...
+    istats->counters[pr.command]++;
+    c->thread->stats.proxy_conn_requests++;
+    c->thread->stats.proxy_req_active++;
```
smukil, line 2658: We're increasing `proxy_req_active` here, but we increase it again in `proxy_run_coroutine()`. Wouldn't that count the same command twice?

dormando:
Might merge this down as-is and do a separate PR or come back later for slow log improvements. Not liking most of the ideas I have for the slow log.

proto_proxy.c
```diff
@@ -4053,6 +4111,15 @@ static int mcplib_open_dist_jump_hash(lua_State *L) {
 
 /*** END jump consistent hash library ***/
 
+/*** START lua interface to logger ***/
+
+static int mcplib_log(lua_State *L) {
+    LIBEVENT_THREAD *t = lua_touserdata(L, lua_upvalueindex(MCP_THREAD_UPVALUE));
+    const char *msg = luaL_checkstring(L, -1);
+    LOGGER_LOG(t->l, LOG_PROXYUSER, LOGGER_PROXY_USER, NULL, msg);
```
smukil, line 4119: I'm just confirming that the timestamp here is the timestamp at the time of the log being put onto the buffer, and not the time that the log is consumed.
Im 99% sure the former is the case, but just confirming.

proto_proxy.c
dormando: Yeah; I can fix that. I'm just lazy since it's nearly impossible for these to fail + at start time.

proto_proxy.c
dormando: This would be because the remote server sent a response it didn't understand for some reason (any reason why that would happen; lua sent junk or the backend server is broken and sent garbage)

proto_proxy.c
dormando: There're a lot of details, but the main difference between global atomic counters and the per-worker thread mutexes are contended shared memory vs typically uncontended differential locks. These mutexes will typically never collide (only writer is that worker thread, only other time it's locked is when it's being observed), and the memory is unique to each worker thread so there's no false-sharing or shared cacheline level contention. It's complicated :)

This also isn't a new lock for this code, it was already there.

proto_proxy.c
dormando: That's a DECR in `proxy_run_coroutine()`

proto_proxy.c
dormando: Log line's timestamp is when the log is inserted into the buffer. The elapsed time for this log entry is between the initial request generation and the result being returned to the client.

PR #870, opened 2022-02-14
proxy: 'stats proxy' crashes on invocation

smukil:
I noticed running 'stats proxy' caused a crash. On observing it under
GDB, it crashed while attempting to access the worker thread level
proxy_user_stats in process_proxy_stats().
This showed that the thread level 'proxy_user_stats' wasnt properly
initialized. On checking it's initialization code in
proxy_worker_reload(), a typo was noticed that caused the bug.
This patch fixes it.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -1103,7 +1103,7 @@ static int proxy_thread_loadconf(LIBEVENT_THREAD *thr) {
     STAT_L(ctx);
     struct proxy_user_stats *us = &ctx->user_stats;
     struct proxy_user_stats *tus = NULL;
-    if (us->num_stats != 0) {
+    if (us->num_stats == 0) {
         pthread_mutex_lock(&thr->stats.mutex);
         if (thr->proxy_user_stats == NULL) {
             tus = calloc(1, sizeof(struct proxy_user_stats));


smukil:
@dormando Re-thinking this fix. Might not be correct.

The issue is that if we don't add a Lua stat (that calls `mcplib_add_stat()`), calling `stats proxy` crashes since `us->num_stats` is 0.

I believe this issue comes down to what's the contract of a thread-level `proxy_user_stat` struct? Is it supposed to be initialized if there are no user stats? If not, then we should add checks for whether it's inited before accessing it at `process_proxy_stats()`

If it's supposed to be initialized even if there are no user stats, then we need to make sure we're initing it all the time be default.

dormando:
I probably just broke this recently since I'm pretty sure I tested running with/without user stats... but like one of my TODO items the unit test coverage isn't great so these'll pop up until it is better.

Was just coming to write that I don't think that's the right fix :) That routine is supposed to skip the user stats update code if they don't exist. I'll take a closer look in a minute.

dormando:
Ah, no. This never worked. I had another bug with it failing if it ran without any user stats initialized, but probably never tested the stats command without stats.

smukil:
Thanks for the quick turnaround! So the contract is that it's not initialized when we don't specify user stats. This needs to be kept in mind if more code needs to touch `tus->num_stats`.

PR #872, opened 2022-02-18
Split proxy into multiple files

dormando:
Probably my grossest habit... I tend to build things out then reorgnize them into functional units slowly. Since this project went on for so long the file separation didn't happen at all.

Code was still generally organized, so this pulls out most of the major sections into their own files. Can clean up the interfaces over time since that's not as urgent as making the different components more obvious to humans.

Core file's down below 1kloc! Yay.


dormando:
Merged up... I definitely want to keep working on the interfaces as these files get revised, but I'm happy with them at least being split up by general type.

Might move some of the smaller lua files up into proxy_lua.c, depending on if some of these other things can be isolated enough to get their own .h's.

PR #874, opened 2022-02-26
sock ip filtering tagging support for FreeBSD

devnexen:


dormando:
Thanks!

Can you add more detail about what this is for? The help argument doesn't say anything about FreeBSD or what it's for.

devnexen:
with ipfw (ip filtering) you can redirect some packets with a ID of your choosing.

devnexen:
so what do you think ? I could not find anything close on Linux.

dormando:
Looks like you figured out linux/openbsd. I'm still not sure of the use case. I understand what it does but nobody's ever asked for this before. was this something you actually make use of?

devnexen:
In some corner cases it might be useful, in my case I have a box with particular firewall rules.

dormando:
released in 1.6.17

PR #875, opened 2022-02-26
whitespaces test getting perl interpreter from env instead

devnexen:


dormando:
Thanks!

What does this fix, exactly? None of the other scripts use the env perl either.

devnexen:
because the perl interpreter is not necessarily in /usr/bin :-)

dormando:
I think I'm going to give this a rare not merge? There're like a hundred+ places with /usr/bin/perl and I kinda don't want just a few of them to be env perl. that's asking for confusion :)

devnexen:
ah ok was just fixing what was necessary maybe other places are more accessory anyway are they external tools ?

dormando:
all of t/*, version.pl, etc. used in build and test stuff.

dormando:
what platform do you have that this even broke on, btw? I have a bunch of buildbots and never seen this, not even on mac

devnexen:
freebsd

dormando:
13? I have a 12 builder and it's never complained (but maybe I'm not running the scripts in the same way?)

devnexen:
most likely because on freebsd the perl interpreter is in /usr/local/bin (from ports)

dormando:
indeed.. my 12 is like that. I guess the only places perl is actually called in the normal build routine is using env perl. Though I thought ages ago version.pl was being called on fbsd builders.

either way, I get it now. Sorry for the extra poking :)

dormando:
released in 1.6.17

PR #877, opened 2022-03-02
Proxy feature roundup

dormando:
There're a few usability features that bother me too much to go without in "v1"

- a hacky method of supporting noreply/quiet commands; I can see what would be bullet proof but will take time to get there.
- allow creating set requests from get responses. this is at least partly in now.
- _maybe_: get meta binary keys to work. Unfortunately this requires some fiddling as the key needs to be unpacked/repacked. The parser operates on const strings (as it must) so we would have to deal with binary keys from a layer higher.
- _maybe_: some extra functions for finding/editing meta tokens/flags

The meta routing tokens means we wouldn't necessarily have to examine binary keys at all, so for people trying to get every ounce of performance they can still avoid the transforms easily.

Everything past these should be stability fixes instead of feature updates until v1 is out.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -228,10 +228,15 @@ void proxy_submit_cb(io_queue_t *q) {
     while (p) {
         // insert into tail so head is oldest request.
         STAILQ_INSERT_TAIL(&head, p, io_next);
-        if (!p->is_await) {
+        if (p->is_await) {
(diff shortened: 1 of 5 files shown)


dormando:
not going to have a lot of work hours this week... just going to merge this down and try super hard to punt on anymore features and work on bugs/stabilization/testing/unit tests instead.

PR #879, opened 2022-03-03
Fix buffer overflow and intermittent backend failure due to recv() of size 0

feihu-stripe:
This PR aims to fix two issues:

## issue1: read buffer overflow, illustrated below.
corruption

## issue2: intermittent backend failure
This happens when `toread` is not _prepared_ in drive machine in some cases, which results in recv() call with 0 byte followed by `_reset_bad_backend()`, [link](https://github.com/feihu-stripe/memcached/blob/85c4e59bd9cdfa1bdf718c14fbc51d62f7886759/proxy_network.c#L990) to code

### memcached commit
[83d36ff7c6bbdb7783d352bb056554c83efcd](https://github.com/memcached/memcached/commit/a7b83d36ff7c6bbdb7783d352bb056554c83efcd)

###memcached-proxylibs 
[fa1fb8145f5167a682603afebb019e4e4df8d41](https://github.com/memcached/memcached-proxylibs/commit/7fa1fb8145f5167a682603afebb019e4e4df8d41)

### System
Linux qa-mydev--09780a313110af27e.northwest.stripe.io 5.11.0-1028-aws #31~20.04.1-Ubuntu SMP Fri Jan 14 14:37:50 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### mctester command
./basic -pipelines 10 -keyprefix /foo/ -conncount 10 -server 127.0.0.1:11211 -stripkeyprefix
see this [PR](https://github.com/memcached/mctester/pull/3) for extra options

### topology
```
s.my_zone("az1")
s.pool {
	name = "foo",
	zones = {
		az1 = {
			"127.0.2.1:11212",
		},
	}
}
```

### start commands
`./memcached -l *:11212 -d`
`./memcached -o proxy_config=/pay/src/memcached-proxylibs/lib/topo.lua -l *:11211`

### Without this fix
proxy will crash in seconds
### With `bread = 0` only
proxy will not crash but spit backend error from time to time in mctester
### With this fix
proxy will not emit any error in mctester


feihu-stripe:
cc @smukil

dormando:
Sweet, thanks! I'll look over this carefully since the flow is complicated.

re: process, can you squash those commits? I normally do this for folks on accepting but it saves some time if I don't have to. Basically; you can have multiple commits but they need to work independently so the main tree history can be bisected/inspected.

feihu-stripe:
Thanks @dormando, commits squashed.

To trace down the bugs, I have LLDB debugger setup in vscode on my mac. If that's useful, I can share the configurations.

dormando:
Thanks. Probably not helpful; myself I'm on Linux and am an old school gdb'er. I'm thinking now about integration/unit tests and code structure in particular. Or rather I would be but I'm juggling a lot of parallel tasks this week.

feihu-stripe:
Sounds good, and thank again @dormando for your patience and guidance to the first time contributor like me.

Please keep us posted about your testing and ideas.

dormando:
You're doing great, thanks :) You landed in what's probably the hardest to follow part of the code right now and sussed out a few bugs.

Sure; lets move over to the mctester changes actually. I'd love to get those integrated so I can utilize them myself. I'll ping that issue in a minute.

dormando:
This has been pushed into `next`, thanks!

PR #884, opened 2022-03-21
Improve Slab Automove behavior

vnsavage:
- Skip using crawler items when calculating the automover age stats as
they can severely skew the ages in the stats to the point of completely
starving particular slabs
- Include the current window data in the window sum so we don't free
pages that are actually needed - this also matches the python script
behavior
- Reset young / old when interrupting the automove decision loop so we
don't accidentally move things which we didn't mean to

### DIFF
--- a/items.c
+++ b/items.c
@@ -660,10 +660,17 @@ void fill_item_stats_automove(item_stats_automove *am) {
         i = n | COLD_LRU;
         pthread_mutex_lock(&lru_locks[i]);
         cur->evicted = itemstats[i].evicted;
-        if (tails[i]) {
-            cur->age = current_time - tails[i]->time;
-        } else {
+        if (!tails[i]) {
             cur->age = 0;
+        } else if (tails[i]->nbytes == 0 && tails[i]->nkey == 0 && tails[i]->it_flags == 1) {
+            /* it's a crawler, check previous entry */
+            if (tails[i]->prev) {
+               cur->age = current_time - tails[i]->prev->time;
+            } else {
+               cur->age = 0;
+            }
+        } else {
+            cur->age = current_time - tails[i]->time;
         }
         pthread_mutex_unlock(&lru_locks[i]);
      }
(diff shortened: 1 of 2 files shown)


dormando:
Thanks!  Just as a warning it might take a while to get this reviewed/merged. I try to be extra conservative with the slab rebalancer since it's tricky. Good finds!

I'm curious though: how are you ending up with crawler items at the bottom of LRU's often enough for the rebalancer algorithm to notice them? LRU crawlers move pretty quickly, often have long time gaps before re-running, and should only be the bottom instance for a brief moment as they start there. Hoping there isn't a second bug here.

Points 2/3 look like good finds, thanks!

vnsavage:
Thanks for looking. It was actually a fun one to debug, because it's a race condition between the threads and when you try to catch it in `fill_item_stats_automove()` with gdb the bug magically disappears.

The LRU crawlers do move pretty quickly, but the crawler items are added with `lru_maintainer_crawler_check()` right before `slab_automove_run()` is called which allows it to happen often enough. 

The preconditions we have are:

1. Memcached instances which run tons of evictions all the time.
2. A particular slab that has significantly lower max age and eviction_time than the rest of the slabs due to the data it contains and its usage pattern.
3. Instances that have significantly longer uptime than the max age of all items in Memcached.

With these preconditions, once we get a single crawler item in a window, the age of that window becomes current_time, because the crawler item has a time of 0. This significantly skews the automove algorithm to think that based on the average age, this slab is the `oldest` instead of the `youngest` and it starts reassigning pages out of it until it starves it down to 2 pages.

vnsavage:
I only kind of scanned of see if something can lead to more serious exploits that we should be worried about. It seemed like the worst impact would be unexpected behavior, like possibly in `lru_maintainer_juggle()` but I haven't spent too much time on it to verify etc.

dormando:
note to self: need to apply this to `slab_automove_extstore.c` as well.

dormando:
Spent a while working on simplifying the automover for extstore, so I got the code loaded well enough into my brain to review this finally :)

All the changes make total sense. The crawler fix is common to both movers and I ported the latest-window-sum change to the extstore automover as well. For the oldest/youngest change apparently I had a `too_free` flag in the extstore mover already.

This should be out in the next release, thanks again!

dormando:
released in 1.6.17

PR #887, opened 2022-04-02
proxy: mcp.log_req* API interface

dormando:
Been experimenting with different logging interfaces. The `watch proxycmds` logger has a very limited view into the request context (mainly, it only has the response object and not the original request to compare to), but adds some other hard to get information (total length of a request cycle, etc). To get better logging with full context and some control lets hoist this into lua. It's also a crazy firehose so lets add some conditionals/sampling.

This adds `mcp.log_req(request, response, "detail")` and `mcp_log_reqsample(milliseconds, rate, allerrors, rq, rs, "detail")`. The idea is to add context from passing through a route function. See `t/startfile.lua`'s usage. Note that in startfile the reqlog is buried pretty low, but it could be the top level route as well. This means you can change logging granularity (or if it should log at all) per route if desired.

TODO:
- [x] Still figuring the interface I prefer: lots of funcs or funcs with options. Don't want to create/dump an options table so we're stuck with arguments. (done: went with two functions)
- [x] This adds a way to only log if it took over some elapsed time, or a sample rate (ie; one in 1000 reqs on average). Also want to add a way of saying "if we're sampling or checking elapsed, also log 100% of request errors (ie; not `MCMC_OK`) (done)
- [x] Needs to encode request and detail for the log output (yes yes yes, it's fine! you can't parse the log lines unless this happens!) (snipping newline from the request so we can just drop the whole line in as the final argument here... will make parsing log lines a bit harder but not sure it matters). if people put spaces in the detail section that's their own fault)
- [x] Not convinced I need the random generator vs a modulo. Modulo isn't exactly fast but it'll still beat this random generator by a bit. (meh, fine.)
- [x] rewire t/startfile.lua to add log detail from the zone functions. (got one example up)
- [x] Need to extend these a bit so the "response" can be a string, so we can log plain responses created by lua. Currently `watch proxycmds` will only fire log lines that create full response objects.
- [x] Need to get backend host/port info into the string. Sadly doesn't look like any better options other than copying the whole strings into the response object on every request :(
- [x] Update logger to reserve a min 1k space for these potentially huge lines :(
- [x] Maybe: delete proxycmds entirely. it's half useless and can minus out a decent chunk of code. Could reuse the proxycmds flag as proxyreqs specific to `log_req`
- [ ] Maybe: Store the length of the backend host strings in `mcp_backend_s` so we can remove all of the strncpy's.

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -304,33 +304,6 @@ static int _logger_parse_cce(logentry *e, char *scratch) {
 }
 
 #ifdef PROXY
-static void _logger_log_proxy_raw(logentry *e, const entry_details *d, const void *entry, va_list ap) {
-    struct timeval start = va_arg(ap, struct timeval);
-    char *cmd = va_arg(ap, char *);
-    unsigned short type = va_arg(ap, int);
-    unsigned short code = va_arg(ap, int);
-
-    struct logentry_proxy_raw *le = (void *)e->data;
-    struct timeval end;
-    gettimeofday(&end, NULL);
-    le->type = type;
-    le->code = code;
-    le->elapsed = (end.tv_sec - start.tv_sec) * 1000000 + (end.tv_usec - start.tv_usec);
-    memcpy(le->cmd, cmd, 9);
-    e->size = sizeof(struct logentry_proxy_raw);
-}
-
-static int _logger_parse_prx_raw(logentry *e, char *scratch) {
-    int total;
-    struct logentry_proxy_raw *le = (void *)e->data;
-
-    total = snprintf(scratch, LOGGER_PARSE_SCRATCH,
-            "ts=%d.%d gid=%llu type=proxy_raw elapsed=%lu cmd=%s type=%d code=%d\n",
-            (int) e->tv.tv_sec, (int) e->tv.tv_usec, (unsigned long long) e->gid,
-            le->elapsed, le->cmd, le->type, le->code);
-    return total;
-}
-
 // TODO (v2): the length caps here are all magic numbers. Haven't thought of
 // something yet that I like better.
 // Should at least make a define to the max log len (1024) and do some math
@@ -383,7 +356,6 @@ static int _logger_parse_prx_req(logentry *e, char *scratch) {
     int total;
     struct logentry_proxy_req *le = (void *)e->data;
 
-    // TODO: encode detail / req
     total = snprintf(scratch, LOGGER_PARSE_SCRATCH,
             "ts=%d.%d gid=%llu type=proxy_req elapsed=%lu type=%d code=%d status=%d be=%.*s:%.*s detail=%.*s req=%.*s\n",
             (int) e->tv.tv_sec, (int) e->tv.tv_usec, (unsigned long long) e->gid,
@@ -435,8 +407,7 @@ static const entry_details default_entries[] = {
     [LOGGER_PROXY_CONFIG] = {512, LOG_PROXYEVENTS, _logger_log_text, _logger_parse_text,
         "type=proxy_conf status=%s"
     },
-    [LOGGER_PROXY_RAW] = {512, LOG_PROXYCMDS, _logger_log_proxy_raw, _logger_parse_prx_raw, NULL},
-    [LOGGER_PROXY_REQ] = {1024, LOG_PROXYUSER, _logger_log_proxy_req, _logger_parse_prx_req, NULL},
+    [LOGGER_PROXY_REQ] = {1024, LOG_PROXYREQS, _logger_log_proxy_req, _logger_parse_prx_req, NULL},
     [LOGGER_PROXY_ERROR] = {512, LOG_PROXYEVENTS, _logger_log_text, _logger_parse_text,
         "type=proxy_error msg=%s"
     },
--- a/logger.c
+++ b/logger.c
@@ -330,6 +330,70 @@ static int _logger_parse_prx_raw(logentry *e, char *scratch) {
             le->elapsed, le->cmd, le->type, le->code);
     return total;
 }
+
+// TODO (v2): the length caps here are all magic numbers. Haven't thought of
+// something yet that I like better.
+// Should at least make a define to the max log len (1024) and do some math
+// here.
+static void _logger_log_proxy_req(logentry *e, const entry_details *d, const void *entry, va_list ap) {
+    char *req = va_arg(ap, char *);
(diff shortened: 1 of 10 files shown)


dormando:
Why we have to copy the backend name/port strings:
- the actual backend objects are shared global objects. the lua objects only exist in the central "config VM". they are referenced through the pool objects that are proxied into each worker thread. aka; the ownership is complicated.
- once a response is actually received we can't hold a reference to the backend object anymore (since we only could because we held a reference to the proxy pool object).

so we need to copy the string on every single request if we want to occasionally return the string in a log entry... optimization can/should come later but it does pain me. Could use the backend reference if the pool proxy object reference lived as long as the response object, but that could end up being a wash because `luaL_ref()` and friends aren't exactly free either.

t/startfile.lua
```diff
@@ -130,19 +139,19 @@ function failover_factory(zones, local_zone)
...
                     --break
-                    return res
+                    return res, "failover_backup_hit"
                 end
             end
-            return restable[1]
+            return restable[1], "failover_backup_miss"
         end
-        return res -- send result back to client
+        return res, "failover_hit" -- send result back to client
```
smukil, line 154: Fyi for the startfile.lua.

CMD_GET actually relies on the `failover_factory` since it's not mapped to the `walkall_factory`. Hence the log_req() ends up printing for every sampled GET call.

logger.c
```diff
@@ -330,6 +330,70 @@ static int _logger_parse_prx_raw(logentry *e, char *scratch) {
...
+    e->size = sizeof(struct logentry_proxy_req) + reqlen + dlen;
+}
+
+static int _logger_parse_prx_req(logentry *e, char *scratch) {
+    int total;
+    struct logentry_proxy_req *le = (void *)e->data;
+
+    // TODO: encode detail / req
+    total = snprintf(scratch, LOGGER_PARSE_SCRATCH,
+            "ts=%d.%d gid=%llu type=proxy_req elapsed=%lu type=%d code=%d status=%d be=%.*s:%.*s detail=%.*s req=%.*s\n",
```
smukil, line 388: Thanks for putting this patch out @dormando !

Sharing some information I found while testing this out, I notice a lot of info getting truncated:
1. The BE ips and ports are getting truncated for many of the calls. 
2. The `req` doesn't seem to print the entire command. It sometimes prints some of the information, but it doesn't contain the prefix and key.

Eg logs:
`ts=1649199095.50545 gid=489377 type=proxy_req elapsed=132 type=100 code=15 status=0 be=10.1:w detail=failover_hit req=get       2321742

ts=1649199095.50701 gid=489378 type=proxy_req elapsed=136 type=100 code=15 status=0 be=10.1:w detail=failover_hit req=get       2321824

ts=1649199095.51151 gid=489379 type=proxy_req elapsed=125 type=100 code=15 status=0 be=10.1:w detail=failover_hit req=get       2321743

ts=1649199095.51330 gid=489380 type=proxy_req elapsed=157 type=100 code=15 status=0 be=<valid-ip>:11211 detail=failover_hit req=get       2321825`

Note, the last logline did have a valid IP, but I just replaced it with <valid-ip>.

Note also that I also am not using startfile.lua. We've been using proxylibs as a starting point for our Lua files and I ported the use of log_req there.

smukil: I'm going to play around with the Lua to not print on every sampled GET but one of the things I suspect is: Due to this logline being printed so frequently, maybe we're overwriting our circular logger buffer.

dormando: log issue should be fixed. forgot to update the entry size when I added the backend ip/port.

smukil: Oh wait, my bad, the key does get printed sometimes. It sometimes gets truncated.
2321742, 2321824, etc. in the logs above are the key names.

t/startfile.lua
dormando: Not sure I'm following the issue here? If you want to sample you need to use `log_reqsample`, else my example there will print every GET call. Or am I missunderstanding?

smukil: Ah, I meant every GET call. I mistakenly thought log_req() was sampling.

In any case, we only would want to print a reqlog for GETs that failover. We aren't too interested in getting all this extra info for GETs from the nearzone.

I'll just make the necessary changes on our end.

dormando: This change adds two API calls, `log_req` which logs unconditionally and `log_reqsample`, which has several options for how to conditionally log (check the e-mail thread for an example for reqsample). Then I added an example on how to log via wrapping routes. They're just API calls so you can use either of them however you see fit.

logger.c
dormando: keep an eye out for an off-by-one-or-two bug still. insomnia has left my ability to math out to lunch.

dormando:
Decided that I will remove the proxycmds log entry, but leave proxycmds watcher channel for `log_req*`, though will likely rename it to proxyreqs. Please let me know if it looks good otherwise and I'll merge it in today.

smukil (commented):
Hey @dormando. Thanks for the updates. We did more testing with your change after fixing the memory corruption and things look very good now. I'm yet to test it internally after `proxycmds` and `proxyuser` are now merged into `proxyreqs` but I went over the commit and it looks straightforward and good to merge.

A couple of minor questions:
1. We don't use `type`, `code` and `status` very much right now. Could you share any insight on how useful those fields could be?
2. We'd want to print the prefix as well, but from my reading of the code, I can pass it as a detail to log_req(). Is my understanding correct?

dormando:
proxycmds and proxyuser weren't merged, proxycmds became proxyreqs. user is still a separate stream just in case that's useful to anyone.

type/code have info on what the parser thinks of the request and response.. so you have the request text but you don't have the response text, and type/code will tell you something there. status is how you know if the response was an error or normal response (ie; backend failed, etc) with further detail in type/code. Those're lightweight anyway.

If you're stripping the prefix from the request before passing it along you would add it back in via 'detail', yes. Wish I could avoid having to modify the underlying request object but that's how it is at the moment.

smukil:
ACK. Thanks for the explanation. I'll pay more attention to those type/code/status fields while debugging.

And to confirm, my understanding of proxyuser, cmds, and reqs matches what you explained as well.

Passing the prefix via the detail should suffice for us.

We're good with merging this if you are too. Thanks again!

dormando:
squashed and pushed to `next`

PR #889, opened 2022-04-03
use define for magic filedescriptor value

ComputerScienceMasterStudent:
added asserts and replaced hard coded value with a macro


dormando:
Hard to argue with but not a high value patch. Also please give PR's useful names.

Thanks, though!

PR #890, opened 2022-04-07
tls: Add switch to opt-in to kernel TLS on OpenSSL 3.0.0+

LINKIWI:
This patch enables the ability to opt-in to kernel TLS offload via extended option `ssl_kernel_tls`. kTLS implements the TLS record layer directly in the kernel, allowing for in-kernel software encryption/decryption or hardware offload to a supported NIC. OpenSSL 3.0.0+ offers support for transparent underlying usage of kTLS by setting the SSL context option `SSL_OP_ENABLE_KTLS`. In some cases, especially when dedicated HW is available, kTLS can offer a reduction in memcached resource usage when serving high throughput.

I tested this on Linux 5.13.0 with statically linked OpenSSL at 7c78932b9a4330fb7c8db72b3fb37cbff1401f8b, verifying counters at `/proc/net/tls_stat`.

(I've been able to verify this end-to-end with closed-environment benchmarking, but am still in the process of evaluating against some real production traffic.)

### DIFF
--- a/doc/tls.txt
+++ b/doc/tls.txt
@@ -18,6 +18,12 @@ will use new certificates without a need of re-starting the server process.
 # Multiple ports with and without TLS : by default all TCP ports are secured. Optionally we can setup
 the server to secure a specific TCP port.
 
+# Kernel TLS offload (kTLS): when compiled with OpenSSL 3.0.0+ with the `enable-ktls` build-time
+option set, and running on Linux with kernel version 4.17+ with `CONFIG_TLS` enabled, memcached can
+transparently offload cryptography operations to the kernel. Depending on a variety of factors
+(including the workload, NIC support for HW-accelerated cryptography, ciphers in use, etc.),
+kTLS may offer improved throughput and/or reduced memcached resource consumption.
+
 Note that initial implementation does not support session renegotiation.
 
 Design
--- a/memcached.c
+++ b/memcached.c
@@ -233,6 +233,7 @@ static void settings_init(void) {
     settings.ssl_last_cert_refresh_time = current_time;
     settings.ssl_wbuf_size = 16 * 1024; // default is 16KB (SSL max frame size is 17KB)
     settings.ssl_session_cache = false;
+    settings.ssl_kernel_tls = false;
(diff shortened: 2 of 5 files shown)


dormando:
sorry for the lag, looks useful! thanks!

any updates on performance differences?

LINKIWI:
Yeah, I've been canarying an OpenSSL 3.1.0-dev and kTLS build on one of our dark traffic clusters this past week. Unfortunately, for our environment and workloads, there's a fairly reliable (though minor) performance regression on the order of single-digit percentage points. Our cloud vendor doesn't supply hardware with NICs capable of inline TLS offload, so I'm only able to test SW kTLS.

Nonetheless, this patch might be useful for other workloads or those who have dedicated NICs with HW offload support. I'm happy to hold off on this patch since personally I'm not able to extensively test/benchmark--up to you.

neelpatelbiz:
I am wondering how to build against a version of openssl 3.0.0+.
Configuring the memcached build via: 
./configure --enable-tls --with-libssl=/home/openssl-3.0.0 --with-libevent=/home/libevent_build
and building:
gives a binary that links against older versions of openssl's libcrypto/libssl:
/home/memcached/memcached
        linux-vdso.so.1 (0x00007fffb53c7000)
        libevent-2.1.so.7 => /home/libevent_build/lib/libevent-2.1.so.7 (0x00007fb814ec5000)
        libssl.so.1.1 => /usr/local/lib64/libssl.so.1.1 (0x00007fb814e08000)
        libcrypto.so.1.1 => /usr/local/lib64/libcrypto.so.1.1 (0x00007fb814b1b000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fb814af8000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb814906000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fb8148fe000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fb814f82000)

PR #891, opened 2022-04-12
'proxyreqs' does not work unless 'proxyuser' also provided

smukil:
Ideally we should be able to see the requst logs only after:
"watch proxyreqs"


dormando:
Argh, sorry! Thanks for the fix

dormando:
released in 1.6.16

PR #899, opened 2022-05-13
Fix undefined behavior and warning with clang

tstellar:
This fixes a warning from clang about undefined behavior:

error: embedding a directive within macro arguments has undefined behavior [-Werror,-Wembedded-directive]


dormando:
thanks!

dormando:
released in 1.6.16

PR #901, opened 2022-06-01
Proxy parser rework

dormando:
There are some potential protocol desync bugs due to mcmc handling its
own buffers and the newer event handler handling its own syscalls.

this change should have better separation of code for the buffer
tracking. if this change works I will add some optimizations to reduce
memmove's.

this also needs an extension to allow reading large values directly into the value buffer again. right now it'll double copy large values and run excessive recv calls. I've not made this too hard but I want to validate correctness before optimizations.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -5,6 +5,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <errno.h>
 
 #include <lua.h>
 #include <lualib.h>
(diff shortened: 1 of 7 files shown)


dormando:
Since this has a few critical bugs in it now, I'm auditing the code and probably merging into next before doing the optimization pass. It looks like a bit of work to do the optimizations and I want to invest more time in the testing rig to validate the work.

It should still be fast enough for most people's use cases and will be improved soon.

dormando:
released in 1.6.16

PR #904, opened 2022-06-20
Fix the hashstring size for evcache ketama string

pkarumanchi9:


pkarumanchi9:
This patch fixes the following -

1. Debug builds will hit the assert when using proxy mode
2. evcache ketama string size is getting truncated. So counted the parts[0] twice to fit in the string.

(gdb) p hashstring
$8 = 0x55555589b5b0 "xx.xx.xx.xx/xx.xx.xx.xx:11211" 

211 static void _add_server_evcache(ketama_t *kt, size_t hashstring_size, const char **parts,
212         lua_Integer bucket_size, lua_Integer id, unsigned int *cont) {
213     char *hashstring = malloc(hashstring_size);
214
215     for (int k = 0; k < bucket_size / 4; k++) {
216         unsigned char digest[16];
217
218         // - create hashing string for ketama
219         snprintf(hashstring, hashstring_size, "%s/%s:%s-%d", parts[0], parts[0], parts[1], k);

dormando:
Thanks! I should try to add more tests so that stops getting broken.

Sorry for the late response; also could you configure git to rebase instead of merge upstream? I don't mind but I have to fiddle with your branches a bit to get a clean merge (or I usually just cherry pick, heh)

pkarumanchi9:
Fixed.

dormando:
thanks!

dormando:
released in 1.6.16

PR #911, opened 2022-07-25
proxy: mcplib_request_token() doesn't delimit the final token in a request

smukil:
mcplib_request_token() allows us to parse each token in a request string. The
implementation delimits tokens using <whitespace>. This approach works for
every token but the last one which will be followed by \r\n.

### DIFF
--- a/proxy_request.c
+++ b/proxy_request.c
@@ -653,15 +653,12 @@ int mcplib_request_token(lua_State *L) {
             lua_pop(L, 1); // got a nil, drop it.
 
             // token not uploaded yet. find the len.
-            char *s = (char *) &rq->pr.request[rq->pr.tokens[token-1]];
-            char *e = s;
-            while (*e != ' ') {
-                e++;
-            }
-            vlen = e - s;
+            char *start = (char *) &rq->pr.request[rq->pr.tokens[token-1]];
+            char *end = (char *) &rq->pr.request[rq->pr.tokens[token]];
+            vlen = end - start;
 
             P_DEBUG("%s: pushing token of len: %lu\n", __func__, vlen);
-            lua_pushlstring(L, s, vlen);
+            lua_pushlstring(L, start, vlen);
             lua_pushvalue(L, -1); // copy
 
             lua_rawseti(L, -3, token); // pops copy.


smukil:
Seems like SSL tests are broken?

```
IO::Socket::SSL not installed or failed to load, cannot run SSL tests as requested at ./t/binary.t line 10.
./t/binary.t .................. 
Dubious, test returned 2 (wstat 512, 0x200)
No subtests run 
IO::Socket::SSL not installed or failed to load, cannot run SSL tests as requested at ./t/getset.t line 10.
# Looks like your test exited with 2 before it could output anything.
./t/getset.t .................. 
```
https://github.com/memcached/memcached/runs/7510542134?check_suite_focus=true

dormando:
> Seems like SSL tests are broken?
> 
> ```
> IO::Socket::SSL not installed or failed to load, cannot run SSL tests as requested at ./t/binary.t line 10.
> ./t/binary.t .................. 
> Dubious, test returned 2 (wstat 512, 0x200)
> No subtests run 
> IO::Socket::SSL not installed or failed to load, cannot run SSL tests as requested at ./t/getset.t line 10.
> # Looks like your test exited with 2 before it could output anything.
> ./t/getset.t .................. 
> ```
> 
> https://github.com/memcached/memcached/runs/7510542134?check_suite_focus=true

The github auto tests are busted right now yeah. It's on the backlog list :( Tests still all pass fine if you run them yourself.

dormando:
Neat, thanks!

I feel like I actually added the finalizer token to the tokenizer... that means you can always get the length by checking the distance to the next token and the loop can be removed entirely. I know I designed this but can't fully remember if I added all the code for it.

I can do this though since I know you're pretty busy (just yell at me if you want to try yourself tho :P)

smukil:
> Neat, thanks!
> 
> I feel like I actually added the finalizer token to the tokenizer... that means you can always get the length by checking the distance to the next token and the loop can be removed entirely. I know I designed this but can't fully remember if I added all the code for it.
> 
> I can do this though since I know you're pretty busy (just yell at me if you want to try yourself tho :P)

Good call! Let me give it a quick go in that case. It might help me understand the parser better (mcp_parser_t and surrounding code) :)

dormando:
Go for it! The tokenizer's a lot simpler than it used to be, but I got a little clever with the 16-bit offset thing as an optimization. It got a lot simpler again when I started adding the finalizer/end cap token.

Just checked the code and right there in `process_tokenize` is the commented endcap, so this should be an easy fix!

smukil:
Yup, it was easy to figure out and the cap token was there. Thanks for the pointer :)

dormando:
released in 1.6.16

PR #912, opened 2022-07-30
proxy: mcp.request improvements

dormando:
- errors if a string value is missing the "\r\n" terminator
- properly uses a value from a response object
- allows passing in a request object for the value as well

- also adds r:vlen() for recovering the value length of a response

think this still needs r:flags() or similar?

TODO: a few more functions for this particular PR:
- [x] `r:has_flag("C")`
- [x] `r:flag_token("C")` find and returns the flag token if exists. bonus: replace if second arg
- [x] `resp:code()` for retrieving meta status code (HD vs EX vs etc)
- [x] `resp:line()` for getting the raw response meta line, which can be inspected in lua while the lower level API is missing.

edit: note to self the characters are flags, arguments tokens, lol.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -457,6 +457,7 @@ int mcplib_request_ltrimkey(lua_State *L);
 int mcplib_request_rtrimkey(lua_State *L);
 int mcplib_request_token(lua_State *L);
 int mcplib_request_ntokens(lua_State *L);
+int mcplib_request_has_flag(lua_State *L);
 int mcplib_request_gc(lua_State *L);
 
 int mcplib_open_dist_jump_hash(lua_State *L);
--- a/proxy.h
+++ b/proxy.h
@@ -458,6 +458,7 @@ int mcplib_request_rtrimkey(lua_State *L);
 int mcplib_request_token(lua_State *L);
 int mcplib_request_ntokens(lua_State *L);
 int mcplib_request_has_flag(lua_State *L);
+int mcplib_request_flag_token(lua_State *L);
 int mcplib_request_gc(lua_State *L);
 
 int mcplib_open_dist_jump_hash(lua_State *L);
--- a/proxy_lua.c
+++ b/proxy_lua.c
@@ -32,6 +32,23 @@ static int mcplib_response_hit(lua_State *L) {
     return 1;
 }
 
+// Caller needs to discern if a vlen is 0 because of a failed response or an
+// OK response that was actually zero. So we always return an integer value
+// here.
+static int mcplib_response_vlen(lua_State *L) {
+    mcp_resp_t *r = luaL_checkudata(L, -1, "mcp.response");
+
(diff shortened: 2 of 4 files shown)


dormando:
After poking at this a few times I now remember where I left off... I wanted to pull the revised tokenizer out into mcmc, or at least abstracted enough to use for responses as well as requests. There's no way to look at the flags of a `mcp.response` for meta as well. I had also not fully converted the tokenizer to use the endcap token instead of re-scanning so I'll do that.

I also have some user reported glitches in the meta protocol (extra spaces) which I should probably fix... it _shouldn't_ cause a compatibility issue but it might. It looks like a bug in the tokenizer made meta delete not work the way I expected.

dormando:
waiting a day for a review (if not I'll self review and cut a release)

smukil:
Quick note: Thanks for writing this up :) I went through the code and it looks pretty straight forward. I'm trying this out shortly. Will get back on this by EOD.

dormando:
@smukil thanks! I'm doing a self-review now and found a few bugs. currently fixing them, but I've force-pushed a couple times with updates in case you need to pull again

dormando:
@smukil ok think I'm out of bugs. Give it a shot? I might still merge it early so I can cut a release in the next few hours.

smukil:
@dormando Got around to trying it out. I tried every call with regular get/set and meta read/write commands and I was able to access the flags, vlen, line.

This LGTM!

smukil:
r:flags() will be useful too if you were planning on adding it.
It will help us figure out what flags are present in a request without having to try each has_flag(<flag>) before fetching its value. If it were returned as a map to Lua, that would be ideal (if that's possible).

dormando:
Thanks!

I figured in most cases you're looking for specific flags so I didn't do that. Now I'm forgetting if there's a way to get the full request line in lua? I did definitely add that for responses, and if you need to iterate the keys it might be better to just walk that string. Creating a table with entries for each flag is a lot of memory churn if you're doing it in a high usage route.

What's the use case for iterating all the flags all the time?

dormando:
I think an iterator might be an easier sell for walking the flags or flag/token pairs. not quite as many allocs

smukil:
You're right, iterating should just be easier. We wouldn't need all the flags all the time in any case I can think of for now, but more than 1; and that should be easy to iterate on.

Thanks again!

dormando:
re-reading your question: if you're looking for a flag with a token (like CAS) you don't have to `has_flag` beforehand. `flag_token()` returns (bool, str), with bool stating if the flag exists and if it has a token the second arg will be a string instead of nil.

`has_flag` is only separate since it can fast path without looking at the response line and most flags don't have tokens anyway.

smukil:
Ack, that should work well for us for now. flag_token() returning (bool, str) can simplify the Lua for sure.

PR #916, opened 2022-08-08
proxy: allow selecting top level routes based on listener tag

dormando:
`-l tag[asdfasdf]:0.0.0.0:11211`

in proxy:

`mcp.attach(mcp.CMD_ANY_STORAGE, routetop, "asdfasdf")`

If a third argument is given to mcp.attach, it will only route there if the client socket came from the matching listener socket's tag.

TODO
- [x] in this same PR I want to add core: `-l noproxy:etc` for skipping the proxy.
- [ ] `req:listener_tag()` if a request was created directly by a tagged client, this should also return the string tag. in cases where you don't want to use the attach method or have many tags and with use a lua table.

I _think_ I want a `noproxy` tag, but `mcp.attach(mcp.CMD_ANY, nil, "tag")` might be cleaner to implement so I'll have to check how hard it is first.

@smukil

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -647,7 +647,7 @@ void conn_io_queue_return(io_pending_t *io) {
 conn *conn_new(const int sfd, enum conn_states init_state,
                 const int event_flags,
                 const int read_buffer_size, enum network_transport transport,
-                struct event_base *base, void *ssl) {
+                struct event_base *base, void *ssl, uint64_t conntag) {
     conn *c;
 
     assert(sfd >= 0 && sfd < max_fds);
@@ -694,6 +694,7 @@ conn *conn_new(const int sfd, enum conn_states init_state,
 
     c->transport = transport;
     c->protocol = settings.binding_protocol;
+    c->tag = conntag;
 
     /* unix socket mode doesn't need this, so zeroed out.  but why
      * is this done for every command?  presumably for UDP
@@ -3072,7 +3073,7 @@ static void drive_machine(conn *c) {
 #endif
 
                 dispatch_conn_new(sfd, conn_new_cmd, EV_READ | EV_PERSIST,
-                                     READ_BUFFER_CACHED, c->transport, ssl_v);
+                                     READ_BUFFER_CACHED, c->transport, ssl_v, c->tag);
             }
 
             stop = true;
@@ -3472,7 +3473,8 @@ static void maximize_sndbuf(const int sfd) {
 static int server_socket(const char *interface,
                          int port,
                          enum network_transport transport,
-                         FILE *portnumber_file, bool ssl_enabled) {
+                         FILE *portnumber_file, bool ssl_enabled,
+                         uint64_t conntag) {
     int sfd;
     struct linger ling = {0, 0};
     struct addrinfo *ai;
@@ -3614,12 +3616,12 @@ static int server_socket(const char *interface,
                 }
                 dispatch_conn_new(per_thread_fd, conn_read,
                                   EV_READ | EV_PERSIST,
-                                  UDP_READ_BUFFER_SIZE, transport, NULL);
+                                  UDP_READ_BUFFER_SIZE, transport, NULL, conntag);
             }
         } else {
             if (!(listen_conn_add = conn_new(sfd, conn_listening,
                                              EV_READ | EV_PERSIST, 1,
-                                             transport, main_base, NULL))) {
+                                             transport, main_base, NULL, conntag))) {
                 fprintf(stderr, "failed to create listening connection\n");
                 exit(EXIT_FAILURE);
             }
@@ -3642,6 +3644,7 @@ static int server_socket(const char *interface,
 static int server_sockets(int port, enum network_transport transport,
                           FILE *portnumber_file) {
     bool ssl_enabled = false;
+    uint64_t conntag = 0;
 
 #ifdef TLS
     const char *notls = "notls";
@@ -3649,7 +3652,7 @@ static int server_sockets(int port, enum network_transport transport,
 #endif
 
     if (settings.inter == NULL) {
-        return server_socket(settings.inter, port, transport, portnumber_file, ssl_enabled);
+        return server_socket(settings.inter, port, transport, portnumber_file, ssl_enabled, conntag);
     } else {
         // tokenize them and bind to each one of them..
         char *b;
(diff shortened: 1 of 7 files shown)


dormando:
Instead of noproxy: I went with `proto[ascii]` to allow forcing specific protocols on specific ports, which is less .. specific.

not sure if I'll do the `req` bits now. it can probably wait.

dormando:
merged the flag_token change into `next` and rebased this branch on top of `next` for ease of use.

smukil:
fyi: reviewing + trying this out now. Will send in my review by EOD

smukil:
This works perfectly for me.

I tried tagging 2 different top level routes- one default top-level route attached to the default listening port (11211), and one top-level route listening on (tag[trimprefix]:11221)

`trimprefix` trims the prefix of the key, and the default top level route doesn't.

I confirmed in the backend, that based on the port used to send the `set` command, the prefix is either there or not.

--------
> noproxy tag, but mcp.attach(mcp.CMD_ANY, nil, "tag")

Could you remind me what the use-case was for skipping the proxy? It seems to me that tagging a listener port and then adding that port as a backend does the trick for what I need. So I can't recall what we discussed for `noproxy`.

--------


> req:listener_tag()

This might make our implementation much simpler. Would it cause a performance penalty to do the runtime check everytime though?

Eg This is much simpler to implement:
```
    if req:listener_tag() == "prefixtrim" then
        r:ltrimkey(j)
    end
```
But would it be any slower than just having a completely different top-level tag that doesn't have to do this runtime check on every request?

dormando:
> Could you remind me what the use-case was for skipping the proxy? It seems to me that tagging a listener port and then adding that port as a backend does the trick for what I need. So I can't recall what we discussed for `noproxy`.

`noproxy` was for adding a backend that accesses the local cache embedded in the proxy instance. I guess what you're saying is that might accidentally work if you set a tag on a listener but don't set a proxy hook on them? In that case it's similar, but the noproxy mode will be faster since it bypasses the proxy subsystem entirely.

> This might make our implementation much simpler. Would it cause a performance penalty to do the runtime check everytime though?

That's exactly the problem. The way it is here is much faster. I figure you won't have many of these (maybe literally one or two) so it won't be a huge burden to have a special top level assignment for them.

If you have tons of listeners it would make sense to do the req:listener_tag() mode and use a lua hash table for the route lookup, but if you have just a couple you can bypass that extra hash table fiddling here.

dormando:
you're usually pretty worried about the performance implications of this and that, so this patch ensures there are optimized paths :)

smukil:
> noproxy was for adding a backend that accesses the local cache embedded in the proxy instance

Ah yes, this is a pretty strong requirement. It completely slipped my mind we discussed this.
I haven't tested this path yet, but the code makes sense, as long as it skips the proxy subsystem and uses the 'ascii' protocol, we would expect it to behave like a memcached backend.

I guess while trying to run this, I would be thinking of it as a "memcached-backend that can route", rather than a "router that can cache data", if that makes sense.
To be a bit more clear, I'd want to make sure that a memcached-backend can do very limited and specific routing on every request, but its main purpose is to store data.
Eg: A memcached backend can perform its own replication by receiving the request as a proxy, which routes the request back to itself and stores it as an item and also sends it to another memcached replica.

Now since the same process is behaving as a router and as a backend, the number of worker threads would double right?Should I be reducing the number of Lua worker threads to make room for the ascii-prot worker threads (Say on a 4 core machine)? Since the Lua threads are doing the network calls themselves anyway and they're instead done by the single proxy thread.

--------

> That's exactly the problem. The way it is here is much faster. 

Ack, then we'll stick with the top level route way.

smukil (approved):
LGTM, I'll test out the `forced ascii` mode (or bypass proxy mode) soon, but I don't want to hold up the patch.

dormando:
> I guess while trying to run this, I would be thinking of it as a "memcached-backend that can route", rather than a "router that can cache data", if that makes sense.
> To be a bit more clear, I'd want to make sure that a memcached-backend can do very limited and specific routing on every request, but its main purpose is to store data.
> Eg: A memcached backend can perform its own replication by receiving the request as a proxy, which routes the request back to itself and stores it as an item and also sends it to another memcached replica.

That is a core intent of the way I designed it, that it could work either way. I just haven't gotten that part hooked up as seamlessly as it will be yet. Basically I've been slowly working toward a unified protocol parser, and when that works the lua system can just say "execute this locally" and get a response.

If you have a "router that can cache", that just means you probably have a lot of lua and a lot of backends. If you have "A cache that can sometimes route", that just means a different setup.

For example, if you wanted to have a server that ansers `get` commands directly, but will run some lua and replicate `set` commands, you can do that today by adjusting `mcp.attach` - if you don't set ANY or ANY_STORAGE and split out the commands, any command not overridden is handled by the cache directly. The only difference coming in future plans is that you could have that decision in lua instead of only at the top level.



> Now since the same process is behaving as a router and as a backend, the number of worker threads would double right?Should I be reducing the number of Lua worker threads to make room for the ascii-prot worker threads (Say on a 4 core machine)? Since the Lua threads are doing the network calls themselves anyway and they're instead done by the single proxy thread.

They're the same threads, you don't have to do anything. A normal worker thread will answer cache requests or process lua requests.

dormando:
To reiterate, if you have a backend where you only want special behavior on SET, it looks like:

`mcp.attach(mcp.CMD_SET, function (r) return "override me!\r\n" end)`
... and any other specific command you want to override. Then commands like "GET" or any other not specifically overridden command, runs the get as though the proxy code didn't exist.

To reiterate the problem, if you wanted to do this there's no way to execute that SET locally _and_ remotely because local execute is missing :) You could use this backend + `noproxy` trick to get around it until that feature is done.

In any case again the difference between "a router that can cache" and "a backend that can route" is nothing. this system is supposed to do both.

smukil:
> To reiterate, if you have a backend where you only want special behavior on SET, it looks like:
> 
> mcp.attach(mcp.CMD_SET, function (r) return "override me!\r\n" end)
> ... and any other specific command you want to override. Then commands like "GET" or any other not specifically overridden command, runs the get as though the proxy code didn't exist.

This is good for now. As long as gets don't have to go through the proxy subsystem, we're good. SETs can be every so slightly slower until we can do local executes from Lua.

> the difference between "a router that can cache" and "a backend that can route"
Yea I incorrectly thought the worker threads were different and was thinking of how to split the resources but since they're the same, it doesn't matter.

PR #917, opened 2022-08-09
proxy: req:flag_token("F", "Freplacement")

dormando:
returns (exists, previous_token)
optional second argument will replace the flag/token with supplied
flag/token, or nothing if "" is passed.

Done some limited testing and seems to work right. Apologies for not doing this in the original PR :)

@smukil

### DIFF
--- a/proxy_request.c
+++ b/proxy_request.c
@@ -728,10 +728,13 @@ int mcplib_request_has_flag(lua_State *L) {
     return 1;
 }
 
+// req:flag_token("F") -> (bool, nil|token)
+// req:flag_token("O", "Onewopauqe") -> (bool, oldtoken)
 int mcplib_request_flag_token(lua_State *L) {
     mcp_request_t *rq = luaL_checkudata(L, 1, "mcp.request");
     size_t len = 0;
     const char *flagstr = luaL_checklstring(L, 2, &len);
+    bool replace = false;
     if (len != 1) {
         proxy_lua_error(L, "has_flag(): meta flag must be a single character");
         return 0;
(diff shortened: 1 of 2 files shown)


dormando:
Debating if I should move more of the behavior of `r:token()` over where it can return updated tokens, or if I document it as a "fast" function that only handles the original request string.

proxy_request.c
```diff
@@ -753,7 +767,19 @@ int mcplib_request_flag_token(lua_State *L) {
                 if (vlen > 1) {
                     // strip the flag off the token and return.
                     lua_pushlstring(L, s+1, vlen-1);
-                    return 2;
+                    ret = 2;
+                }
+
+                // Have something to replace the flag/token with.
+                if (replace) {
```
smukil, line 774: Thanks for the quick drop @dormando !

One of the things I noticed in both token() and flag_token() is that after changing the value of either of them, we set `r->was_modified = true`

However, we don't change the actual metadata of the original request itself.

For example, if I do the following:
```
      print(r:flag_token("C"))
      r:flag_token("C","")   -- blank out the CAS token
      print(r:flag_token("C"))
```

I get the following output:
```
true	6
true	6
```

For our purposes, we don't need to modify the original request more than once (yet).

But from an API guarantees POV, what do you think would be the right approach here?

smukil: > Debating if I should move more of the behavior of r:token() over where it can return updated tokens, or if I document it as a "fast" function that only handles the original request string.

Is this what you were referring to here? ^

dormando: Yes, that's what I'm referring to. The token accessors are optimized so you can make a bunch of modifications without having to re-write the request string over and over. `was_modified` is a hint for the request attach function to flatten the new request query.

I'm not really sure what the right answer is; I'm tempted to leave it like this for now and change it soonish. I had some ideas for how to speed up request rewriting such that it might not matter anymore. ie; now that the endcap works you can fast-copy before/after a token, so it ends up being extra memcpy's instead of all the lua table fiddling. That'll either come out break even or slightly ahead due to fewer allocations.

but I'd like to avoid doing that this week.

dormando: So; yeah the difference here is `r:token()` is always checking the lua table to see if a token's been overwritten. I can add that but we lose some "fast-pathy-ness".

Probably _should_ do it for expectations but not super sure it's necessary.

`local (exists, old) = r:flag_token("C", "Cnew")`

in this case old is not "new", but is whatever the previous token was. If a use case needs the previous token you have it there, and if you need the new one a second time you already have it as you just set it.

If you issue a request, IIRC it flattens everything. So if you modify a request with `r:flag_token`, issue it to a backend, then modify it again, you'll get the new data.

From there it gets kinda complex unfortunately.

smukil: Ack, that's fair. For our purposes, this works for now.

The only thing I'd add here is to make sure that the APIs are documented with their appropriate design choices/limitations, so that it's clear to other folks adopting it.

smukil (approved):

proxy_request.c
dormando: I'll update the docs on release and make a note about a potential behavior change, thanks!

PR #920, opened 2022-08-11
Race leads to deadlock during shutdown (sigterm/sigusr1)

smukil:
We've noticed in some rare cases, that sending a sigterm to a memcached process
could lead to a race condition which leads to a deadlock of the process.

The worker threads are stuck on 'worker_hang_lock' because the shutdown
process hasn't completed yet.

The shudown process hasn't completed since the logger thread signals
'logger_stack_cond' without locking its corresponding 'logger_stack_lock'.
This could cause the signal to get lost, causing the logger thread to
wait indefinitely on 'logger_stack_cond' in logger_thread().

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -826,8 +826,12 @@ static int start_logger_thread(void) {
 }
 
 static int stop_logger_thread(void) {
+    // Guarantees that the logger thread is waiting on 'logger_stack_cond'
+    // before we signal it.
+    pthread_mutex_lock(&logger_stack_lock);
     do_run_logger_thread = 0;
     pthread_cond_signal(&logger_stack_cond);
+    pthread_mutex_unlock(&logger_stack_lock);
     pthread_join(logger_tid, NULL);
     return 0;
 }


smukil:
For more context, here is the hung process under GDB:

Memcached version: https://github.com/memcached/memcached/commit/8d356ae1523f4fec72ade529cf74ef5ddc15e4aa

All Worker threads stuck in (4 threads):
```
Thread 7 (Thread 0x7f7ce35fb700 (LWP 1303)):
#0  0x00007f7ce6117170 in __lll_lock_wait () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00007f7ce610f0a3 in pthread_mutex_lock () from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x000055759d86f40b in register_thread_initialized () at thread.c:151
#3  0x000055759d86f46c in worker_libevent (arg=0x55759efd5400) at thread.c:521
#4  0x00007f7ce610c609 in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#5  0x00007f7ce602f133 in clone () from /lib/x86_64-linux-gnu/libc.so.6
```

The main thread waiting to join the worker threads (Essentially in `stop_logger_thread()`):
```
Thread 1 (Thread 0x7f7ce5f0d740 (LWP 748)):
#0  0x00007f7ce610dcd7 in __pthread_clockjoin_ex () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x000055759d86fb9a in stop_threads () at thread.c:248
#2  0x000055759d85e1a1 in main (argc=<optimized out>, argv=<optimized out>) at memcached.c:6159
```

dormando:
released in 1.6.17

PR #921, opened 2022-08-12
log: fix obscure crashes due to size_t promotion

dormando:
It dawns on me now that I should've been hinting or forcing the type at
all of the LOGGER_LOG endpoints to ensure they match the va_arg side on
logger.c. This patch initially swaps size_t's for int's so at least it's
using less complicated promotion rules.

Fixes a crash on centos 7 when logging sets via `watch mutations`

TODO
- [ ] audit all `LOGGER_LOG` callers and ensure the passed arguments match the va_arg side.


smukil (approved):
Ouch what a way to catch a compiler bug. The fix looks like it's needed regardless.

dormando:
took a quickish look at all the va_arg calls and logger calls. they seem good enough, though if I ever make a serious pass at it I should line them up more carefully.

PR #924, opened 2022-08-25
core: make large item storage more reliable

dormando:
When allocating sub-max chunks for the tail end of a large item the
allocator would only look at the exact slab class. If items in a cache
are all exclusively large, these slab classes could be empty. Now as a
fallback it will also check and evict from the largest slab class even
if it doeesn't necessarily want the largest chunk.

### DIFF
--- a/items.c
+++ b/items.c
@@ -226,8 +226,23 @@ item_chunk *do_item_alloc_chunk(item_chunk *ch, const size_t bytes_remain) {
     unsigned int id = slabs_clsid(size);
 
     item_chunk *nch = (item_chunk *) do_item_alloc_pull(size, id);
-    if (nch == NULL)
-        return NULL;
+    if (nch == NULL) {
+        // The final chunk in a large item will attempt to be a more
+        // appropriately sized chunk to minimize memory overhead. However, if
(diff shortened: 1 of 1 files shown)


dormando:
gave this some more testing: loaded an `-I 2m -m 4000`  instance with all-same large items (1.3 megs), then hit it with a set of random length 1m+ items. pre-patch this instantly threw the EOM error repeatedly. post-patch this doesn't throw any error.

it doesn't appear to be leaking memory or anything; hammering it with sets and the current items and bytes stored seems roughly static.

dormando:
released in 1.6.17

haydenme:
I pulled the 1.6.17 release and ran the same application that showed the problem originally. When running against 1.6.14 this application could not run more than 30s with encountering repeated "Out of memory during read" errors when performing writes that should have evicted older entries. (The application is a video processing pipeline that writes 1-1.5MB values to memcached 30 times a second.) With the 1.6.17 release it ran for ~2hr without any errors.

dormando:
@haydenme awesome! Thanks for confirming. Sucks I didn't think to try that years ago :)

PR #926, opened 2022-08-28
configure.ac: add --disable-werror

ffontaine:
Allow the user to disable -Werror to avoid the following build failure:

```
In file included from hash.c:7:
xxhash.h:2667:5: error: #warning is a GCC extension [-Werror]
 2667 | #   warning "XXH3 is highly inefficient without ARM or Thumb-2."
      |     ^~~~~~~
xxhash.h:2667:5: error: #warning "XXH3 is highly inefficient without ARM or Thumb-2." [-Werror=cpp]
```

Fixes:
 - http://autobuild.buildroot.org/results/3124bae73c207f1a118e57e41e222ef464ccb297

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,6 +130,9 @@ AC_ARG_ENABLE(proxy,
 AC_ARG_ENABLE(proxy-uring,
   [AS_HELP_STRING([--enable-proxy-uring], [Enable proxy io_uring code EXPERIMENTAL])])
 
+AC_ARG_ENABLE(werror,
+  [AS_HELP_STRING([--disable-werror], [Disable -Werror])])
+
 dnl **********************************************************************
 dnl DETECT_SASL_CB_GETCONF
 dnl
@@ -829,6 +832,10 @@ AM_CONDITIONAL([BUILD_SPECIFICATIONS],
                [test "x$enable_docs" != "xno" -a "x$XML2RFC" != "xno" -a "x$XSLTPROC" != "xno"])
 
 
+if test "x$enable_werror" != "xno"; then
+   CFLAGS="$CFLAGS -Werror"
+fi
+
 dnl Let the compiler be a bit more picky. Please note that you cannot
 dnl specify these flags to the compiler before AC_CHECK_FUNCS, because
 dnl the test program will generate a compilation warning and hence fail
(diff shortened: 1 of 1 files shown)


dormando:
Thanks!

What platform is that?

Any reason why this is being papered over instead of addressing the warning?

ffontaine:
The build failure is raised by buildroot autobuilders on arm926ej-s. 
IMHO, we can't address this warning: `__thumb2__` and `__ARM_ARCH_ISA_ARM` are just not defined by the toolchain on this architecture.

dormando:
There are other hash functions, so we could elect to not use/default to xxhash3 if it would be specifically unhappy. xxhash3 was only relatively recently added, even.

I don't really like the idea of giving people the ability to mask off Werror since that's my early warning system.

yann-morin-1998:
@dormando 

> Any reason why this is being papered over instead of addressing the warning?
> I don't really like the idea of giving people the ability to mask off Werror since that's my early warning system.

Using -Werror during development is very good, it helps very much improve the quality if the code.

However, consumers of your code will have a variety of situations that will trigger various warnings you could not have met. For example, a newer compiler will introduce new warnings, or have an improved detection of existing warnings. Different compilers (e.g. clang vs gcc) may also generate different sets of warnings. Even the same compiler version on different build machines may generate different sets of warnings. And then, even the same compiler on the same machine may generate different sets of warning, depending on how it was configured at build
time.

For example, let's assume some code is warning-free with, say, gcc-10. Then, gcc-11 gets out, with new warnings, and the code which was previously warning-free now hits this new warning. Is the code now worse than it was previously? Surely, no, it is the exact same code and was deemed to be correct then, and it is no longer less correct, so forcing -Werror in that case does not help at all.

See [-Werror is Not Your Friend](https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/) for an more in-depth explanation.

Of course, consumers may very well be enticed to provide fixes for those warnings, but -Werror is not a help in that situation.

Thanks you! :-)

dormando:
Not sure I really buy that. Yes, new compilers/different compilers have different warnings. Most of my "users" aren't actually building memcached, they're packagers; often on bleeding edge distro's. The packages are downstream from there.

Since I don't have bleeding edge versions of everything I often get reports as things are found, then they get fixed quickly, then anyone who does happen to build from tarballs on odd platforms don't get warnings. I feel like that's worked pretty well so far?

In this case we'd be running a potentially glacially slow hash algo on some platform, which seems like a real bug to me. There's no way I would ever build the software on this ARM platform myself (even with all of my buildbots that include a few ARM platforms), so I will never see this warning, and it will never get fixed, and any user who happens to use the software may be disappointed.

It'd be _nice_ if the `next` branch is the thing that got tested against everyone's bleeding edge platform so I don't release things with warnings, but I take what I can get :)

The issue is more subtle than that of course, "what does a user do if they do hit a warning?" but the answer is almost always holding off on an upgrade of some kind and almost nobody ever upgrades off of an upstream packaged version, which has already resolved these issues.

dormando:
@ffontaine can you share more info about what this platform/toolchain is? I'm curious if it has thumb2+ but is just missing the defines or if it's actually going to be glacially slow.

yann-morin-1998:
The failing build was for an arm926t (a [variant](https://elixir.bootlin.com/linux/v6.0-rc7/source/arch/arm/mm/Kconfig#L134) of an [arm920t](https://en.wikipedia.org/wiki/ARM9#ARM9TDMI), so it's an armv4 with Thumb, but not Thumb2.

dormando:
Sorry I was pretty behind again.. since september apparently.

Thought about this more and decided to flip it; Werror is now off by default and this change adds a `--enable-werror` flag I can use in buildbots + local dev (or anyone who cares enough I guess). I don't want to disable pedantic and it seems like pedantic errors trigger some people.

Thanks again!

PR #929, opened 2022-09-01
proxy: update mcmc and calling conventions

dormando:
upstream fixes: mcmc would return OK to garbage responses, which was
probably causing issues in the past.

This does remove the MCMC_CODE_MISS and replace it with MCMC_CODE_END.

Updated the commit a bit more: status is now either `MCMC_OK` or `MCMC_ERR` and resp.code always has the detail. I feel this is slightly cleaner overall.

I also didn't pull in the tests/* dir from mcmc since that's not really necessary yet.

@smukil - tagging for interest

### DIFF
--- a/vendor/mcmc/mcmc.c
+++ b/vendor/mcmc/mcmc.c
@@ -1,15 +1,15 @@
 #include <stdlib.h>
-#include <sys/types.h>
 #include <sys/socket.h>
-#include <sys/uio.h>
+#include <sys/types.h>
 #include <netdb.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <string.h>
-#include <stdint.h>
 #include <errno.h>
 #include <stdio.h>
 
+// TODO: move these structs into mcmc.h, but only expose them if
+// MCMC_EXPOSE_INTERNALS is defined... for tests and this thing.
 #include "mcmc.h"
 
 // TODO: if there's a parse error or unknown status code, we likely have a
@@ -24,8 +24,6 @@
 // at least doubled for wiggle room.
 #define MIN_BUFFER_SIZE 2048
 
-#define FLAG_BUF_IS_ERROR 0x1
-#define FLAG_BUF_IS_NUMERIC 0x2
 #define FLAG_BUF_WANTED_READ 0x4
 
 #define STATE_DEFAULT 0 // looking for any kind of response
@@ -62,9 +60,7 @@ static int _mcmc_parse_value_line(mcmc_ctx_t *ctx, mcmc_resp_t *r) {
     int keylen;
     p = memchr(p, ' ', l - 6);
     if (p == NULL) {
-        // FIXME: these should return MCMC_ERR and set the internal parse
-        // error code.
-        return MCMC_PARSE_ERROR;
+        return -MCMC_ERR_VALUE;
     }
 
     keylen = p - key;
@@ -76,14 +72,14 @@ static int _mcmc_parse_value_line(mcmc_ctx_t *ctx, mcmc_resp_t *r) {
     errno = 0;
     uint32_t flags = strtoul(p, &n, 10);
     if ((errno == ERANGE) || (p == n) || (*n != ' ')) {
-        return MCMC_PARSE_ERROR;
+        return -MCMC_ERR_VALUE;
     }
     p = n;
 
     errno = 0;
     uint32_t bytes = strtoul(p, &n, 10);
     if ((errno == ERANGE) || (p == n)) {
-        return MCMC_PARSE_ERROR;
+        return -MCMC_ERR_VALUE;
     }
     p = n;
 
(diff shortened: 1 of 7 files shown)


dormando:
starting to lean toward resp.code is the detail and top level function returns 0 or -1.

code could always be positive then? not sure. the parser's a lot cleaner with the single status code; could internally use -CODE for errors and then use that for the rv test and invert it.

dormando:
Planning on merging this today or tomorrow.. it should drastically improve bug safety.

smukil:
@dormando Just got back from vacation. I'll submit a review on this by tomorrow 3 PM PST if that works!

dormando:
@smukil wb! A review would be appreciated even if I merge this ahead of time. See the mcmc repo for more details and the tests.

vendor/mcmc/mcmc.c
```diff
@@ -173,8 +168,7 @@ static int _mcmc_parse_response(mcmc_ctx_t *ctx, mcmc_resp_t *r) {
             switch (buf[0]) {
             case 'E':
                 if (buf[1] == 'N') {
-                    code = MCMC_CODE_MISS;
-                    // TODO: RESP type
+                    code = MCMC_CODE_END;
```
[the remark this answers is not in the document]
dormando: It doesn't; see the changes to the :hit() code. The client code updates align the internal codes with what they are in the actual protocol. Considering it a hit or a miss should be at a higher level.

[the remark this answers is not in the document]
dormando: previously the return codes were a weird mix of "what the protocol says" vs "what a user thinks it means" and it was leading to bugs

dormando:
I think a major thing here is that there are 3-4 different perfectly fine approaches to the calling convention, but whatever it is just needs to be consistent. It can easily be changed if the new approach ends up too hard to work with. So far it seems better and was easier to write tests for.

dormando:
I'm also running real slow today so looks like you're winning the race :|

vendor/mcmc/mcmc.h
```diff
@@ -1,14 +1,16 @@
 #ifndef MCMC_HEADER
 #define MCMC_HEADER
 
+#include <sys/uio.h>
+#include <stdint.h>
+
 #define MCMC_OK 0
```
smukil, line 7: nit: For readability, it might be useful to separate these out with some code comments?
(OK, ERR) -> Status

(NOT_CONN .. HAS_RESULT) -> Connection codes

(CODE_STORED .. CODE_END) -> Successful protocol response codes

(ERR_SHORT .. ERR_VALUE) -> Error protocol response codes

vendor/mcmc/mcmc.h
```diff
@@ -1,14 +1,16 @@
...
 #define MCMC_ERR -1
 #define MCMC_NOT_CONNECTED 1
```
smukil, line 9: Part of me wonders if it would be useful to separate out the connection codes from the protocol codes as well?
Without knowing the code all that much in depth yet, it might get tricky to track them as part of the same response code as the code grows, even if they're checked at different layers?

vendor/mcmc/mcmc.h
dormando: they already kind of are, as you noted here CODE_* ERR_* etc. I deleted a few others that were unused and violated these principals already. Not sure it's worth splitting the rest out as they're entirely unused at the moment (see the commented out code)

vendor/mcmc/mcmc.h
dormando: Possibly, but with the read/write functions commented out nothing uses them outside of connect/disconnect. The parser uses WANT_READ but that's just communicated as a normal error code.

If/when the other functions are restored it'll be easier to reason about their return codes.

vendor/mcmc/mcmc.h
dormando: oh with comments. I guess so?

vendor/mcmc/mcmc.h
dormando: Thinking more; potential return codes is kind of a function-by-function (for client facing API calls) thing anyway and if we're documenting things properly it should be at those levels. I think?

smukil: > If/when the other functions are restored it'll be easier to reason about their return codes.
Ack

smukil: > Thinking more; potential return codes is kind of a function-by-function (for client facing API calls) thing anyway and if we're documenting things properly it should be at those levels. I think?

Yea, as long as the contract is in an easily readable format, it may make the codebase easier to navigate. But that's just a nit.

smukil (approved):

dormando:
Think I hear what you're getting at and yes the client is pretty barebones. It needs a doxygen/etc pass for its API. When I first wrote this I thought I'd be returning to it every week and fleshing it out into a full standalone client and I got stuck with just enough time to work on the proxy instead. In this change I removed some of those flags I was fiddling with for as-a-client and added the start of unit tests. As I get time to work on it it'll get some more normal docs, I imagine.

In the meantime it's being used as just a response parser so I haven't been strict. will merge this soon and come back again for more.

PR #930, opened 2022-09-02
proxy: addressing source TODO/FIXME's

dormando:
Doing another pass against all of the TODO/FIXME's in the source files. They should either be addressed now or renamed as "v3". Note that there is no v3, this is just how I'm making the work more greppable.

Not doing these in any _particular_ order. Focus is on making code more testable and potential bugs/crashes instead of optimizations/features.

Might also rename or change the tag for "feature-y" todo/fixme's vs bug/flaw.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1887,7 +1887,7 @@ void server_stats(ADD_STAT add_stats, conn *c) {
     storage_stats(add_stats, c);
 #endif
 #ifdef PROXY
-    proxy_stats(add_stats, c);
+    proxy_stats(settings.proxy_ctx, add_stats, c);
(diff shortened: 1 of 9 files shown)


dormando:
preparing for heat wave and then being not prepared enough for heat wave killed my whole week...

going to probably do a quick skim here and fix one or two other things but I'll have to come back to do a todo/fixme scrub in a few weeks.

dormando:
didn't have time... need to come back on the next cycle. going to merge this as-is. it touches a bunch of places but it's all the same pattern.

dormando:
released in 1.6.18

PR #935, opened 2022-09-15
hacks to help figure out extstore bug

dormando:
if a key gets overwritten many times in a row while constantly flushing to extstore, and then stops being written, the compactor can possibly over-delete from a page and cause a page to be reclaimed in the middle of compaction. This is because the header in memory still matches the page-id and page of the now dead entries.

need to come back to resolve this bug sharing the debugging I did.


dormando:
#1047 closing in favor of fix + less hacked up test.

PR #936, opened 2022-09-23
proxy: add mcp.await FASTGOOD flag

dormando:
returns early on a hit, else waits for N non-error responses.

### DIFF
--- a/proxy_await.c
+++ b/proxy_await.c
@@ -49,6 +49,7 @@ int mcplib_await(lua_State *L) {
             case AWAIT_ANY:
             case AWAIT_OK:
             case AWAIT_FIRST:
+            case AWAIT_FASTGOOD:
                 break;
             default:
                 proxy_lua_error(L, "invalid type argument tp mcp.await");
@@ -284,6 +285,16 @@ int mcplib_await_return(io_pending_proxy_t *p) {
                     } else {
                         // user only wants the first pool's result.
                         valid = false;
+                    }
+                    break;
+                case AWAIT_FASTGOOD:
+                    if (p->client_resp->status == MCMC_OK) {
+                        // End early on a hit.
+                        if (p->client_resp->resp.code != MCMC_CODE_END) {
+                            aw->wait_for = 0;
+                        } else {
+                            is_good = true;
+                        }
                     }
                     break;
             }
--- a/proxy.h
+++ b/proxy.h
@@ -444,6 +444,7 @@ enum mcp_await_e {
     AWAIT_ANY, // any response, including errors,
     AWAIT_OK, // any non-error response
     AWAIT_FIRST, // return the result from the first pool
+    AWAIT_FASTGOOD, // returns on first hit or majority non-error
 };
 int mcplib_await(lua_State *L);
 int mcplib_await_run(conn *c, mc_resp *resp, lua_State *L, int coro_ref);
--- a/proxy_lua.c
+++ b/proxy_lua.c
@@ -859,6 +859,7 @@ static void proxy_register_defines(lua_State *L) {
     X(AWAIT_ANY);
     X(AWAIT_OK);
     X(AWAIT_FIRST);
+    X(AWAIT_FASTGOOD);
     CMD_FIELDS
 #undef X
 }


proxy_await.c
```diff
@@ -286,6 +286,16 @@ int mcplib_await_return(io_pending_proxy_t *p) {
                         valid = false;
                     }
                     break;
+                case AWAIT_FASTGOOD:
+                    if (p->client_resp->status == MCMC_OK) {
+                        // End early on a hit.
+                        if (p->client_resp->resp.code != MCMC_CODE_END) {
+                            aw->wait_for = 0;
```
smukil, line 293: Curious if `cleanup` will always eventually end up being `true` for a FASTGOOD request even if we return early?
i.e., Is it guaranteed that every `io_pending_proxy_t` calls mcplib_await_return(), and so every request's state will be eventually cleaned up?

smukil:
To confirm my understanding of the above:
mcp.await(pool, 1, mcp.FASTGOOD) is the same as mcp.await(pool, 1, mcp.GOOD)

However:
mcp.await(pool, 2, mcp.FASTGOOD) will return after one HIT response, whereas mcp.await(pool, 2, mcp.GOOD) will wait for the second response even if the first one is a HIT.

Is my understanding correct?

proxy_await.c
dormando: cleanup is enforced by `aw->pending`. `aw->wait_for` is only for early return scenarios.

dormando:
Your understanding is correct, and that was the request as I understood it.

"GOOD" means "a non error response". So GOOD is N non-errors and FASTGOOD is 1 hit or N non-errors.

Remember that await always returns a table of all responses so far; so even if you're using FASTGOOD you can get N responses (ie; a miss, then a hit) so you have to check the table for the hit instead just looking at the first entry. This might seem odd but it's necessary when using await as a building block.

proxy_await.c
smukil: Ack. I was just checking to see if `mcplib_await_return()` will always be called if `pending > 0` and `wait_for == 0`, but it sounds like it will.

dormando: So far as I know you use it like that all the time so you'd notice if it were leaking.

`mcplib_await_return` is called for each pending IO that was dispatched unconditionally. internally it decides when to throw its associated client object back to work (ie `wait_for == 0`), and when all pending IO's are done the await object is reaped. The code is laid out a bit weird (it got refactored three times) but it's a state machine. I could rewrite it again to use a switch statement and be more clear but I don't like touching working code :)

It's already pending a refactor to reduce/remove the lua bits so I'll maybe get it then.

dormando: To back up a bit, when calling `mcp.await()` in lua, it creates a high level await object and immediately dispatches all requests, each wrapped with their own IO objects. As those requests complete they are shipped back to the worker thread one at a time where the return code is executed. The return code checks the conditions of the responses for building the lua result table, when to dispatch the client back to work, and when to complete and reap the parent await object.

smukil: > it creates a high level await object and immediately dispatches all requests, each wrapped with their own IO objects

Ack, thanks for that explanation- so to reiterate my understanding:-
- There are resources at 2 levels, the context for a request is at the `await` layer (or a `mcp_await_t`)
- The resources pertaining to a single IO is represented by `io_pending_proxy_t`

`io_pending_proxy_t` is guaranteed to call the callbacks once the IO is either complete, errors out or after some timeout. Once all the pending IOs call the callbacks, the await resources are cleaned up

dormando: yup.

dormando:
Ok. Let me know if/when you're good to have this merged (or if there're any other cases we should consider adding)

proxy_await.c
```diff
@@ -286,6 +286,16 @@ int mcplib_await_return(io_pending_proxy_t *p) {
...
+                        } else {
+                            is_good = true;
+                        }
+                    }
+                    break;
```
smukil, line 298: I think you need to also whitelist AWAIT_FASTGOOD in mcplib_await() (around L49) to have this working.

dormando: fixed, thanks!

smukil:
Fyi, testing this at scale today. Will get back by tomorrow at the latest

dormando:
how'd it go?

hemal-stripe:
Hey @dormando 
We tested this patch out and it worked well (with @smukil's patch included of allowlisting the FASTGOOD mode which I see you have patched as well). 
We tested it under heavy load under normal circumstances as well by adding artificial tcp delay of 10 ms in the backend. 

Out of the 3 backend nodes, we added 10ms packet delay to 2 backends and the latencies were still low due to the last backend being available and quick to respond. 

With this change, only time we would see high latency is when all 3 backends have higher latencies.

PR #937, opened 2022-09-27
proxy: backend connection improvement

dormando:
Improvements to handling of new and failed backend socket connections.

Previously connections were initiated immediately, and initially from the config thread, yet completion of opening sockets wouldn't happen until a request tried to use that backend.

Now we open connections via the IO thread, as well as validate new connections with a "version\r\n" command.

Also fixes a couple of error conditions (parsing, backend disconnect) where clients could hang waiting for a retry time in certain conditions. Now connections should re-establish immediately and dead backends should flip into a bad fast-fail state quicker.

NOTES:

- There is some duplicated code now, which I am okay with. I want a third use case or some more time to pass. I looked at de-duplicating but after adding some required structures the code was roughly the same length. As is it should let me move faster while the backend conn code is tuned.
- A big goal here is to avoid adding a ton more tunables (how many times to retry, timeout lengths, etc) so where possible any algorithms should have reasonable defaults or be dynamically scaled.

TODO:

- [x] Should validation step be optional? I was about to do this but not sure if it's necessary. No, not right now.
- [x] Need to update the "bad backend" algorithm to hold them as bad for longer under more conditions. IE; the number of timeouts over time affects when something goes into fast fail state and for how long. Diff PR.
- [x] Should we run multiple probes before marking a backend as okay? I was planning on doing this but I'm not sure what scenarios this would capture. No, skipping this.
- [x] Probably not now, but in the future, we need a way of distinguishing which connections do connect-ahead. This is a decent default now, but not for sidecar use cases where you want to load the full configuration but clients tend only use a subset of pools. Yes, later.
- [x] For io_uring; leave the compile option in but gate the start option. Leave it disabled until I can make it production worthy.
- [x] Any cases where we can ship errno into the error log? Later: need to translate errno to internal errors to avoid long strings.
- [x] remove PROXY_DEBUG flag I left in.

I want to make the validator have a lua override but there's no clean way of doing this yet. The lua request API only works under workers, so I can't simply ship requests to the config thread or similar. I may try again before merging this. For example, it might make sense for some folks to request a specific key or do a set/get combo and check the timing.

@smukil

### DIFF
--- a/proxy_network.c
+++ b/proxy_network.c
@@ -3,11 +3,37 @@
 
 #include "proxy.h"
 
+enum proxy_be_failures {
+    P_BE_FAIL_TIMEOUT = 0,
+    P_BE_FAIL_DISCONNECTED,
+    P_BE_FAIL_CONNECTING,
+    P_BE_FAIL_READVALIDATE,
+    P_BE_FAIL_BADVALIDATE,
+    P_BE_FAIL_WRITING,
+    P_BE_FAIL_READING,
+    P_BE_FAIL_PARSING,
+};
+
+const char *proxy_be_failure_text[] = {
+    [P_BE_FAIL_TIMEOUT] = "timeout",
+    [P_BE_FAIL_DISCONNECTED] = "disconnected",
+    [P_BE_FAIL_CONNECTING] = "connecting",
+    [P_BE_FAIL_READVALIDATE] = "readvalidate",
+    [P_BE_FAIL_BADVALIDATE] = "badvalidate",
+    [P_BE_FAIL_WRITING] = "writing",
+    [P_BE_FAIL_READING] = "reading",
+    [P_BE_FAIL_PARSING] = "parsing",
+    NULL
+};
+
 static void proxy_backend_handler(const int fd, const short which, void *arg);
+static void proxy_beconn_handler(const int fd, const short which, void *arg);
 static void proxy_event_handler(evutil_socket_t fd, short which, void *arg);
+static void proxy_event_beconn(evutil_socket_t fd, short which, void *arg);
 static void proxy_event_updater(evutil_socket_t fd, short which, void *arg);
 static int _flush_pending_write(mcp_backend_t *be);
 static int _reset_bad_backend(mcp_backend_t *be, enum proxy_be_failures err);
+static void _backend_failed(mcp_backend_t *be);
 static void _set_event(mcp_backend_t *be, struct event_base *base, int flags, struct timeval t, event_callback_fn callback);
 static int proxy_backend_drive_machine(mcp_backend_t *be);
 
@@ -163,10 +189,10 @@ static void proxy_backend_wrhandler_ur(void *udata, struct io_uring_cqe *cqe) {
             // should be safe to hold up until their timeout.
             _reset_bad_backend(be, P_BE_FAIL_CONNECTING);
             _backend_failed_ur(be);
-            P_DEBUG("%s: backend failed to connect\n", __func__);
+            P_DEBUG("%s: backend failed to connect (%s:%s)\n", __func__, be->name, be->port);
             return;
         }
-        P_DEBUG("%s: backend connected\n", __func__);
+        P_DEBUG("%s: backend connected (%s:%s)\n", __func__, be->name, be->port);
         be->connecting = false;
         be->state = mcp_backend_read;
         be->bad = false;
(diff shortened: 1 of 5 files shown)


smukil:
Thanks @dormando, I'm going to try pulling this in and running it, but I'll likely get the time to review it only by early next week. I'll keep you updated!

dormando:
doing a self-review on this so I can merge it and move on. A few of the things I intended for this will just be separate changes.

proxy_network.c
```diff
@@ -961,20 +1046,126 @@ static void proxy_backend_handler(const int fd, const short which, void *arg) {
...
         if (res == -1) {
             _reset_bad_backend(be, P_BE_FAIL_WRITING);
+            _backend_failed(be);
             return;
         }
+        flags |= res;
+    }
+
+    // Still pending requests to read or write.
+    if (!STAILQ_EMPTY(&be->io_head)) {
```
dormando, line 1139: note: gate on !validating here.

dormando:
Going to look at bundling errno into logs again before merge, but I haven't seen any blockers. Just fixed a few things as noted in the TODO list above + comment and pushed it up again.

dormando:
think I'll do the errno stuff separately. got some weird test results and just printing back numbers isn't super useful in portable software... need to dedicate a little more time to it and either: ERRNAME, full string, or add more internal error codes and map the errors. Probably latter is more likely?

smukil:
@dormando I tried this PR against our regular load test infra. From what I've seen so far, it seems to work pretty much the same with or without the patch in terms of backend errors.

For us: We don't see too many "SERVER_ERROR backend failed" when a router is running against in-memory backends.

However, when running against `extstore` backends, we. constantly see `SERVER_ERROR backend failed`- and that's the same with or without this patch. `proxyevents` spits out the following events: `timeout`, `writing` and `parsing`.

What specific error scenarios does the patch address? I'm seeing if I could do more targeted testing to validate the patch better.

dormando:
@smukil I don't think your load tests will encounter a typical backend failure unless you're running scenarios, so I guess it's good that it didn't change anything?

re: extstore what was the stream log for that? I'm guessing timeout since your'e still tuning your extstore backends? There's no other reason for extstore to cause specific backend failures.

In my own testing it fixed some lags that happened when backends were completely killed and replaced. It also fixes lags that would happen during parsing errors and other more rare events.

dormando:
It also fixes a particularly gnarly crash scenario when new backends are added during reload but the connection attempt fails immediately (which I've reproduced via local stuff but would probably require a bad dns lookup normally). that was propagating an error up to lua, but no longer does as the initial connection attempt is async.

dormando:
ah I misread your first comment somehow. not sure why you'd be getting `parsing` errors there; could be timing out halfway through a request being read back.

smukil:
Ack. I think we noticed that some of these disk backed clusters had idle_timeout set. The clusters were also receiving sparse periodic traffic. We're removing the idle_timeout and testing. Our intuition is that the router <> memcached connections become idle by the time they;re used again.

So when we have sustained traffic, it _seems_ to not keep erroring out anymore. But we're still confirming.

We also noticed that when a connection is used after a long time, it errors out very quickly. Eg: The proxyreqs line looks something like:
```
ts=1658967915.755996 gid=90180 type=proxy_req elapsed=41 type=0 code=0 status=-1 be=[10.100.67.128:11211](http://10.100.67.128:11211/) detail=write_failed:testnamespace req=set /foo/bar/83161 0 2592000 23
```

Notice that `elapsed` is only around ~40 micros, which means that it never attempted to use the connection before returning an error.

I'll get back with the results of more of the testing.

dormando:
Thanks! That clears things up I think.

The proxy currently has no way of immediately knowing if a backend socket was closed; I need to rewire a few more things to have a standing READ event added. So when your extstore servers were closing the connection, the proxy would notice the next time it does either a write, or waits on read for a write. it would get an immediate error (40 micros) as the socket is already dead and re-open the connection. So it is trying to use the socket it's just known to be closed already so it errors out fast.

Not fixed in this particular changeset :) I've had a few complaints about that so I've been working toward fixing it and this change does unblock some of that.

dormando:
Sorry I keep hitting submit too fast: this change _does_ fix cases where if a backend socket was closed but the proxy hasn't noticed yet, clients could sometimes wait up to 3 seconds before it re-establishes. Now in all error cases it should immediately try to re-establish the connection instead of waiting for the retry timeout. Clients can still end up waiting for the socket to re-establish but that's all.

Once a backend is marked "bad" it will fast fail client requests and retry a few seconds later, as it did before.

dormando:
Going to merge this down today, for real. keep getting distracted...

I've been toying with ways of improving the test infra to make unit/integration tests of these kinds of changes easier. I've run through this enough by hand and reviewed the code a bunch of times that I feel it's safe.

smukil:
@dormando Fyi, we're doing our chaos testing on it as we speak. We'll be done within the next couple of hours. But so far we don't see any regressions yet.

smukil:
@dormando No regressions to report on this patch. We tried a bunch of manual chaos tasks:
- Killing one backend
- Killing multiple backends of the same replica
- Killing multiple backends of different replicas
- Our regular testing

All looks good so far. We're still running it a bit cautiously since the changes are fairly deep, but I think it's good to merge. We'll let it permeate through our fleet over the next couple weeks.

dormando:
Did find one bug via testing this, but it was from an earlier commit: 5d2da18808d50705f5038f304974d99f75c30f25
(I can't speel, sorry). Was from a TODO/FIXME cleanup commit that I merged a while ago; could cause a crash if a backend is completely removed and a number of restarts happen.

Finally identified a lifecycle issue with backends that was asked about a while ago. Will fix separately.

Both bugs aren't related to this PR though.

dormando:
released in 1.6.18

PR #938, opened 2022-09-28
Meta protocol cleanup adjustments

dormando:
Cleanup fixes for meta protocol:

- Removes some excess spaces for md, ma commands.
- Allows bare "mg key" requests to work
- Reflects O and k flags for mg on miss (EN)
- Removes old `meta_response_old` start flag that was only supposed to be there for 3 months anyway.

Need to double check that I didn't miss any other report that's come up since. There're still some 'things to think about' with meta but I think as separate changes (ie; a "truly quiet" flag)

@smukil - tagging you since you folks are using meta and I'm paranoid of this change breaking something. All of the parsers I've written ignore N spaces before/after things and the protocol is supposed to allow for that. So it should be fine...

I'm going to have to just take an assumption/hope that nobody's coded super specifically to it and take the pain here. I was using regex prefex tests exclusively in the unit tests so I never noticed the bugs (and I don't think anyone ever reviewed meta deeply while it was under development)

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -4830,7 +4830,6 @@ int main (int argc, char **argv) {
         DROP_PRIVILEGES,
         RESP_OBJ_MEM_LIMIT,
         READ_BUF_MEM_LIMIT,
-        META_RESPONSE_OLD,
 #ifdef TLS
         SSL_CERT,
         SSL_KEY,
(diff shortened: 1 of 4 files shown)


dormando:
@zenbones @tdyas - I know you folks were recently working with the meta protocol. Any chance of taking a look at this? I'm concerned about accidental compatibility issues but properly written clients _should_ be fine if they're ignoring spaces instead of looking for specific spacing.

tdyas:
> I know you folks were recently working with the meta protocol. Any chance of taking a look at this? I'm concerned about accidental compatibility issues but properly written clients _should_ be fine if they're ignoring spaces instead of looking for specific spacing.

The particular project of mine which had the need was deferred, so I won't be able to check.

dormando:
going to merge this today too... it does match the doc/protocol.txt spec now so I can rely on that at least.

dormando:
released in 1.6.18

PR #945, opened 2022-10-20
Add watch deletions

hemal-shah:
Adds a `watch deletions` support in memcached:

```bash
watch deletions
OK
ts=1666303388.226398 gid=2 type=deleted key=hello cmd=na clsid=1 size=5
ts=1666303404.25740 gid=3 type=deleted key=nice cmd=na clsid=1 size=10
```

the corresponding sets and deletes: 
```bash
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
set hello 0 0 5
hello
STORED
delete hello
DELETED
set nice 0 0 10
1234567890
STORED
delete nice
DELETED
```

### DIFF
--- a/items.c
+++ b/items.c
@@ -226,8 +226,23 @@ item_chunk *do_item_alloc_chunk(item_chunk *ch, const size_t bytes_remain) {
     unsigned int id = slabs_clsid(size);
 
     item_chunk *nch = (item_chunk *) do_item_alloc_pull(size, id);
-    if (nch == NULL)
-        return NULL;
+    if (nch == NULL) {
+        // The final chunk in a large item will attempt to be a more
+        // appropriately sized chunk to minimize memory overhead. However, if
+        // there's no memory available in the lower slab classes we fail the
+        // SET. In these cases as a fallback we ensure we attempt to evict a
+        // max-size item and reuse a large chunk.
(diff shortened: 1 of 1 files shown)


logger.c
```diff
@@ -251,6 +262,20 @@ static int _logger_parse_ee(logentry *e, char *scratch) {
     return total;
 }
 
+static int _logger_parse_ide(logentry *e, char *scratch) {
+    int total;
+    const char *cmd = "na";
```
smukil, line 267:
```suggestion
```

logger.h
```diff
@@ -108,6 +109,14 @@ struct logentry_item_store {
     char key[];
 };
 
+struct logentry_deletion {
+    int nbytes;
```
smukil, line 113: We might want to track the `cmd`, because deletes can be triggered from different commands.

dormando:
Can you ensure the branch history is clean? ie; rebases on top of `next` and compare to `next`, generally.

- Think this is tied too deeply; will overlap with evictions and other cases where deletions aren't deliberate. I assume the intent of this is to find intentional removals? (ie; done by a command). This already kind of bad with fetchers/mutations.
- Ideally you add some tests to t/watcher.t as well.

PR #947, opened 2022-11-01
core: give threads unique names

dormando:
allow users to differentiate thread functions externally to memcached. Useful for setting priorities or pinning threads to CPU's.

Pretty trivial outside of needing to duplicate the function for extstore. I don't have instances handy to test BSD/OBSD/etc so I've left those forms out for now so I don't accidentally break builds. Any user with such a platform should feel free to submit a tested patch too.

I may or may not move the function to a new header file before merging. I might wait because I've been doing some slow audits and might want to structure it differently if I can find more common code to pull out.

### DIFF
--- a/proxy_config.c
+++ b/proxy_config.c
@@ -167,16 +167,18 @@ int _start_proxy_config_threads(proxy_ctx_t *ctx) {
         pthread_mutex_unlock(&ctx->config_lock);
         return -1;
     }
+    thread_setname(ctx->config_tid, "mc-prx-config");
     pthread_mutex_unlock(&ctx->config_lock);
 
     pthread_mutex_lock(&ctx->manager_lock);
     if ((ret = pthread_create(&ctx->manager_tid, NULL,
                     _proxy_manager_thread, ctx)) != 0) {
-        fprintf(stderr, "Failed to start proxy configuration thread: %s\n",
+        fprintf(stderr, "Failed to start proxy manager thread: %s\n",
                 strerror(ret));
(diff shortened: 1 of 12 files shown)


thread.c
```diff
@@ -629,6 +631,17 @@ static void thread_libevent_process(evutil_socket_t fd, short which, void *arg)
     }
 }
 
+// Interface is slightly different on various platforms.
+// On linux, at least, the len limit is 16 bytes.
+#define THR_NAME_MAXLEN 16
+void thread_setname(pthread_t thread, const char *name) {
+assert(strlen(name) < THR_NAME_MAXLEN);
```
smukil, line 638: Might want to copy the assert to the implementation in extstore.c as well?

proxy_config.c
```diff
@@ -177,6 +178,7 @@ int _start_proxy_config_threads(proxy_ctx_t *ctx) {
         pthread_mutex_unlock(&ctx->manager_lock);
```
smukil, line 178: nit: In line 176 above, the log line should say:
"Failed to start proxy manager thread"

thread.c
dormando: Thanks. I'll re-copy-paste it when I decide if I'm going to move the function or not

dormando:
I tried turning it into a macro but not all of the callers have pthread locally and I was making too big of a change trying to refactor things. I don't want a whole object file for a single function wrapping a single function call for now :( so keeping the dupe code until I get a better opportunity later.

This is merged into `next`

PR #951, opened 2022-11-07
Add opaque data for KEY_TOKEN token

ton31337:


dormando:
This is a pretty clever idea. Not sure I can take the patch but I definitely give it points for creativity. I'll think it over when I get some time.

Thanks!

ton31337:
@dormando thanks for a quick response. We considered using LiteSpeed Memcached implementation, or filtering using eBPF (regex), or writing our own implementation of memcached (aka proxy), but decided it's much better to stick with the performance, lots of perfect work in vanilla memcached, and make a patch that some other people could use as well.

dormando:
The proxy mode I've been working on might be a better place for this sort of thing, though it will still be a while before I have it wired properly. That would add a small abstraction layer you can use to do these sorts of things without having to hook it so deeply into the main code.

The missing piece of the proxy right now is local query resolution (this works via a hack but it's not as fast/native), as it can only proxy requests to others over the network. I would also need to think about where to expose the ipv6 client token at the API level, which would be pretty cool to support.

ton31337:
@dormando a quick update for you, we tested this in a shared hosting environment with lots of WordPress installs, and so far, the tests are promising (as expected).

dormando:
Just tossing an update in here so you don't think I forgot; proxy has come a long way in the last few months. I'm working on a major API revision at the moment, so I'm hoping within a month or two we'll see a good way to hook this in. Doesn't take long to write the code but validating/baking takes a few weeks.

It will definitely be a little slower but probably barely measurable unless your request rate is millions/sec per machine.

ton31337:
> Just tossing an update in here so you don't think I forgot; proxy has come a long way in the last few months. I'm working on a major API revision at the moment, so I'm hoping within a month or two we'll see a good way to hook this in. Doesn't take long to write the code but validating/baking takes a few weeks.
> 
> It will definitely be a little slower but probably barely measurable unless your request rate is millions/sec per machine.

Cool, thanks.

PR #952, opened 2022-11-08
adding an independant portability layer, starting with the debug feature

devnexen:
as naming threads and two platforms.

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -34,6 +34,8 @@
 # pragma ident "$NetBSD: daemon.c,v 1.9 2003/08/07 16:42:46 agc Exp $"
 #endif
 
+#include "config.h"
+
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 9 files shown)


daemon.c
```diff
@@ -34,6 +34,8 @@
 # pragma ident "$NetBSD: daemon.c,v 1.9 2003/08/07 16:42:46 agc Exp $"
 #endif
 
+#include "config.h"
```
sapairsun, line 37: good

dormando:
Thanks!

It's weird I swear I tried to do exactly this but it was failing, but now I can't remember what the error was.

This is missing Makefile.am changes to include the new portability.h file in `make dist`

Did the current `next` branch throw errors anywhere? I see it passing tests on my freebsd12 builder.

devnexen:
> 
> Did the current `next` branch throw errors anywhere? I see it passing tests on my freebsd12 builder.

Hmm nothing particular no.

PR #955, opened 2022-11-23
proxy: add mcp.AWAIT_BACKGROUND

dormando:
mcp.await(request, pools, 0, mcp.AWAIT_BACKGROUND) will, instead of waiting on any request to return, simply return an empty table as soon as the background requests are dispatched.

- [x] Basic testing. Using various log statements I can see the time increasing when using this option vs no option at all.
- [x] Stress testing. Not done any serious stress testing.
- [ ] Bug audit. Standard.

This works by creating a dummy IO object that immediately gets returned to the worker thread once the rest of the IO's are queued properly.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -401,6 +401,7 @@ struct _io_pending_proxy_t {
     bool ascii_multiget; // passed on from mcp_r_t
     bool is_await; // are we an await object?
     bool await_first; // are we the main route for an await object?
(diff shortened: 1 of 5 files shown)


dormando:
Don't think I see anything concerning in the code on a second read through; the change is relatively simple.

Done some stress testing too; no issues so far.

PR #958, opened 2022-12-05
proxy: IO thread performance improvements

dormando:
1) more IOV's per syscall
2) if a backend got a large stack of pending IO's + continual writes the
   CPU usage of the IO thread would bloat while looping past already
   flushed IO objects.


dormando:
Made change I was thinking of; should be safe to merge but I'll let it sit a little while.

dormando:
fixed a bug from needing to reset the write offset when the connection is reset.

dormando:
still want to test this a bit more before cutting a release with it, but it's mergeable.

PR #959, opened 2022-12-08
proxy: fix lifecycle of backend connections

dormando:
A backend's connection object is technically owned by the IO thread
    after it has been created. An error in how this was done lead to invalid
    backends being infinitely retried despite the underlying object being
    collected.
    
    This change adds an extra indirection to backend objects: a backend_wrap
    object, which just turns the backend connection into an arbitrary
    pointer instead of lua memory owned by the config VM.
    
    - When backend connections are created, this pointer is shipped to the
      IO thread to have its connection instantiated.
    - When the wrap object is garbage collected (ie; no longer referenced by
      any pool object), the be conn. pointer is again shipped to the IO
      thread, which then removes any pending events, closes the sock, and
      frees data.

TODO:
- [x] More thorough testing. This passes the basic tests, and adding/removing completely dead backends now stops the retry loop.

There may be some conditions lurking where a backend gets collected while requests are outstanding, so this needs to be audited and if found call a backend reset before freeing the data.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -242,6 +242,8 @@ enum mcp_backend_states {
     mcp_backend_next, // advance to the next IO
 };
 
+typedef struct mcp_backend_wrap_s mcp_backend_wrap_t;
+typedef struct mcp_backend_label_s mcp_backend_label_t;
 typedef struct mcp_backend_s mcp_backend_t;
 typedef struct mcp_request_s mcp_request_t;
 typedef struct mcp_parser_s mcp_parser_t;
@@ -287,6 +289,7 @@ struct mcp_request_s {
 };
 
 typedef STAILQ_HEAD(io_head_s, _io_pending_proxy_t) io_head_t;
+#define MAX_LABELLEN 512
 #define MAX_NAMELEN 255
 #define MAX_PORTLEN 6
 // TODO (v2): IOV_MAX tends to be 1000+ which would allow for more batching but we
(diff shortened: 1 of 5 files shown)


dormando:
Gave this a second read-through and it still looks good. Does need some battle testing as I'm not 100% confident in the cleanup routine yet.

If I run the logic it shouldn't be possible for a backend object to exist with any outstanding requests as the pool object is typically held with them. The pool object has to GC before this ever happens.

Need to triple check this case for mcp.await() requests now that they can be backgrounded.

dormando:
fixed a strncpy compile warning (should probably try harder to kill the few remaining calls of this...) - now idea how that wasn't being set off until just now. I re-ran configure and suddenly it wasn't happy anymore :)

dormando:
Been developing on top of this for a couple weeks and not encountered any bugs... I'm worried about the reset handler kicking in while the backend has active requests queued, but I don't think it's possible. Going to audit it again carefully and merge; see if there are any other tests I can do.

dormando:
I'm reasonably convinced the cleanup code is correct. There cannot be pending requests when a backend is cleaned up, because there are only two code paths where backends get requests stacked on them:
- Calling `pool(r)`, which yields the coroutine with request object and pool object on the stack. they cannot be cleaned up until the request returns.
- Calling `mcp.await()`, which is called with a request and table of pools. The whole table of pools is referenced until the entire await completes, even if it's doing background operations.
- Backends can only be cleaned up if no pool references them.

PR #960, opened 2022-12-12
remove useless function declaration do_assoc_move_next_bucket()

liangxs:
do_assoc_move_next_bucket in assoc.h is useless, it should be removed.


dormando:
Out in 1.6.19

PR #961, opened 2022-12-13
declare item_lock_hashpower as a static variable

liangxs:
The only reference of extern `item_lock_hashpower` was removed by 
[commit 6af7aa0b1581b3b](https://github.com/memcached/memcached/commit/6af7aa0b1581b3bfac98e4fd7c67801cd1f8f1fb#diff-01a5f0312bb8368e0e99f956b269866fbeadfed74fb86597972be641c8fe637f), then it should be a static variable.


dormando:
Out in 1.6.19

PR #962, opened 2022-12-13
remove useless function item_link

liangxs:
Function `mt_item_link ` was renamed to `item_link` in  [commit a9dcd9aceb17ed50](https://github.com/memcached/memcached/commit/a9dcd9aceb17ed502345742e77fd5d2cd89d9869#diff-161b2a279f4c67a1ab075a7890ecf6f3f1d483d910910fa52b3715e25cfdcbd7R430), and then it is never used.


dormando:
Think I'm going to keep this function for API completeness. Don't want to have to put it back if something needs it in the future.

Thanks!

PR #963, opened 2022-12-15
Proxy: update io_uring code

dormando:
Builds on another PR so ignore the first two commits.

Updated the io_uring code to match the libevent code. Pushed a little further to add writes into uring, but performance is not improved over the libevent code.

Not giving up on it yet but it's good to periodically update the code or we'll never get there anyway.

Since the bulk of it is ifdef'd uring code that isn't stable anyway, all we need to review here is the simplification to the flush_pending_write code.


dormando:
been benchmarking code on top of this for weeks. just fixed a bug with uring compilation that snuck in and merged.

PR #968, opened 2022-12-21
bugfix: size.c: struct size error

liangxs:
The size of `Thread stats` is shown as a negative number,  because `MAX_NUMBER_OF_SLAB_CLASSES` was changed to 64.
```
$ ./sizes
Slab Stats	64
Thread stats	-6352
Global stats	208
...
```

after fix:
```
$ ./sizes
Slab Stats	64
Thread stats	2352
Global stats	208
...
```


dormando:
Out in 1.6.19

PR #969, opened 2022-12-22
Proxy: mcp.internal(r) for accessing local cache.

dormando:
Still pending some work. Change has been minimized by upstreaming pre-requisite work.

This change adds a lua command:
```
local res = mcp.internal(r)
```
... which will dispatch the request using the embedded local cache in the proxy. Works with extstore enabled.

This was always the other half of the big plan for embedding the proxy into memcached: one was code reuse, the other was "Swiss army chainsaw" for cache flexibility. I'll get into all the wonderful benefits when this is closer to merge. Some ideas:
- Convert any existing pool of memcached's into proxies. Serve gets from local cache, replicate sets to redundant caches. Backhaul misses, etc.
- Convert legacy client requests into new or adjusted commands. IE: `incr foo 1` could change into a meta arithmetic command with the initializers set.
- Hot caches in larger border routers
- Hot caches in sidecar proxy configurations.

Notes:
- Will not support all of the protocol commands right now (like config commands, etc).
- Supports all the standard cache storage commands (get/set/meta cmds/etc)
- This is a lot of code duplication for the parser. This is fine; I consider it another step towards unifying the parser later. As the commands get rewritten for the proxy local parser the connection object is removed, and later I should be able to port this back to the text protocol handler.

TODO:
- [x] Finalizing support of ascii get. Making this work all the way through with extstore before moving on to the other commands, but this is largely done pending bugs and optimizations.
- [x] Restore `watch fetchers` log entry: this is why t/watcher.t fails to pass, because the log entry is missing.
- [x] Restore `watch mutations` which is now removed as well.
- [x] Remove the temporary mcmc client from the get path (optimization)
- [x] Add SET support; this is the only other odd codepath as it has to fill from the proxy request object.
- [x] Add the rest of the GET/SET type ascii commands that use the same get/set funcs (gets/gat/add/etc)
- [x] Add the remaining meta commands (delete/arithmetic)
- [x] Fix various known bugs or issues (meta cas, set-stale, noreply handling)
- [x] Properly render modified request objects (will currently send pre-modified req string) (bit involved)
- [x] Test building with extstore disabled
- [x] Some misc TODO/FIXME items
- [x] Quick performance analysis to find anything egregious
- [x] Chunked item support (copy functions didn't quite align with what I need)
- [x] Do some easy code cleanups: fix `MCP_PARSER_KEY()` so it can be used, remove the key length check, etc. (not going to do this until I merge the new parsing code with the text handler, so the code has minimal drift during the transitional period)

### DIFF
--- a/proxy_internal.c
+++ b/proxy_internal.c
@@ -1,0 +1,1694 @@
+/* -*- Mode: C; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+// Functions related to local command execution.
+
+#include "proxy.h"
+#include "storage.h"
+
+#define PROXY_STORAGE_GET 0
+#define PROXY_STORAGE_MG 1
+#define _DO_CAS true
+#define _NO_CAS false
+#define _DO_TOUCH true
+#define _NO_TOUCH false
+
+static int _store_item_copy_from_buf(item *d_it, char *buf, const int len) {
+    if (d_it->it_flags & ITEM_CHUNKED) {
+        item_chunk *dch = (item_chunk *) ITEM_schunk(d_it);
+        int done = 0;
+        // Fill dch's via a flat data buffer
+        while (len > done && dch) {
+            int todo = (dch->size - dch->used < len - done)
+                ? dch->size - dch->used : len - done;
+            memcpy(dch->data + dch->used, buf + done, todo);
+            done += todo;
+            dch->used += todo;
+            assert(dch->used <= dch->size);
+
+            if (dch->size == dch->used) {
+                item_chunk *tch = do_item_alloc_chunk(dch, len - done);
+                if (tch) {
+                    dch = tch;
+                } else {
+                    return -1;
+                }
+            }
+        }
+        assert(len == done);
+    } else {
+        memcpy(ITEM_data(d_it), buf, len);
+    }
+
+    return 0;
+}
+
+// TODO (v2): out_string() needs to change to just take a *resp, but I don't
+// want to do the huge refactor in this change series. So for now we have a
+// custom out_string().
+static void pout_string(mc_resp *resp, const char *str) {
+    size_t len;
+    bool skip = resp->skip;
+    assert(resp != NULL);
+
+    // if response was original filled with something, but we're now writing
+    // out an error or similar, have to reset the object first.
+    resp_reset(resp);
+
+    // We blank the response "just in case", but if we're not intending on
+    // sending it lets not rewrite it.
+    if (skip) {
+        resp->skip = true;
+        return;
+    }
+
+    // Fill response object with static string.
+
+    len = strlen(str);
+    if ((len + 2) > WRITE_BUFFER_SIZE) {
+        /* ought to be always enough. just fail for simplicity */
+        str = "SERVER_ERROR output line too long";
+        len = strlen(str);
+    }
+
+    memcpy(resp->wbuf, str, len);
+    memcpy(resp->wbuf + len, "\r\n", 2);
+    resp_add_iov(resp, resp->wbuf, len + 2);
+
+    return;
+}
+
+// For meta commands error strings override the quiet flag.
+static void pout_errstring(mc_resp *resp, const char *str) {
+    resp->skip = false;
+    pout_string(resp, str);
+}
+
+#ifdef EXTSTORE
+static void _storage_get_item_cb(void *e, obj_io *eio, int ret) {
+    io_pending_proxy_t *io = (io_pending_proxy_t *)eio->data;
+    assert(io->active == true);
+    mc_resp *resp = io->tresp;
+    item *read_it = (item *)eio->buf;
+    bool miss = false;
+
+    if (ret < 1) {
+        miss = true;
+    } else {
+        uint32_t crc2;
+        uint32_t crc = (uint32_t) read_it->exptime;
+        crc2 = crc32c(0, (char *)read_it+STORE_OFFSET, eio->len-STORE_OFFSET);
+
+        if (crc != crc2) {
+            miss = true;
+            io->badcrc = true;
+        }
+    }
+
+    if (miss && !resp->skip) {
+        resp->iovcnt = 1;
+        if (io->gettype == PROXY_STORAGE_GET) {
+            resp->iov[0].iov_len = 5;
+            resp->iov[0].iov_base = "END\r\n";
+            resp->tosend = 5;
+        } else if (io->gettype == PROXY_STORAGE_MG) {
+            resp->iov[0].iov_len = 4;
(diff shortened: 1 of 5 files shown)


dormando:
Tests pass with the watcher tests temporarily disabled until `LOG_FETCHERS` log generator is added back in, which I intend to do after the rest of the feature is complete so I can do that all at once.

Passes 30 minutes of benchmarking. Needs thorough QA and tests if I can figure out a good structure.

Probably moving on to getting "set" to work. Then the rest of the supportable commands should be done all at once.

The amount of copypasta will be truly unfortunate but since meta doesn't change much right now it's a good time to do it, and we can merge the code back together after this change and cut the total code lines by quite a lot.

dormando:
Taking a break while working on set support...

There are a few sections of the proxy which copy data instead of reading directly into items and this will be another one for now. To be optimized later.

This change is touching a lot of random core internals. Most of the changes are for good: removing `conn *c` from deep sections of code and instead operating on a `mc_resp *resp` so they can be called with fewer side effects. A lot of functions only need the `c->thread` as well. The actual structure of things is changing very little.

Some of this should reduce code bugs as well since there are a few flags that are being twiddled on/off on the `conn *c` carefuly, where if they're on the `*resp`'s instead they get cleaned up naturally.

The full set of commands to port looks like:
```
- process_get_command : get gets gat gats
  process_update_command : set add append cas replace prepend
  process_arithmetic_command : incr decr
  process_delete_command
  process_touch_command
 - lots of shared code:
  process_mget_command
  process_mset_command
  process_mdelete_command
  process_marithmetic_command
  process_meta_command
```

So once get/update are done that's a bulk of the code. incr/decr/delete/touch are all pretty short code.

The meta code reuses the main structure of get_command and update_command so porting those looks like some really painful copypasting but otherwise simple. hopefully just another day or two of hacking at it.

dormando:
Taking a week off but probably going to chip at the remaining items over the next week... I am very tired.

Next big push is getting all of the cache commands working, which should be relatively easy now that the flows all work. Then a TODO/FIXME scrub, performance pass, then QA.

Another full day or two of work to get it to the QA point so far as I can tell. Maybe less.

dormando:
Most of the commands are in now. Just two more meta commands and that's all of them (unsure if I'll do the meta debug command, possibly/probably?)

Then a QA/FIXME/TODO pass. There're a few bugs or missing features that I wanted to tackle after the commands have basic functionality. Also some macros to clean things up...

The code changes for the rest of the commands were thankfully not drastic and we get to simplify the set path considerably.. I'm definitely redoing the main set commands with the preparser prefilling the item data before calling the command.

Tests won't pass for now.

dormando:
everything except the meta debug command works now (which I'm fine with, at least for now..)

lots of typing but relatively painless so far. moving on to the debug/test/cleanup phase now.

dormando:
First audit pass done. Looking better. Just poking at this for an hour a day during the break because that's easier for me :P this change has a lot of small tasks to do which gets exhausting if I try to do all of it at once.

Gonna push on a few more FIXME/TODO's then start QA pass in parallel with the rest of the code cleanups and optimization.

dormando:
Listed out the TODO a bit more, and I have more detailed notes locally for what needs to be done. Think it's another full day or two of dev work and auditing before moving to more strictly QA. A few tasks (like flattening modified requests, chunked items, etc) are more work to do cleanly than anticipated.

dormando:
The two branches this change was based off of are now upstreamed, so this has been rebased on `next` and only contains changes relevant to `mcp.internal`.

It's still huge, but that's largely just the adjusted code from the text commands, and will shrink later.

dormando:
Once #977 is merged request flattening is dealt with, as the request is updated as soon as it's modified. Also removed the temporary mcmc client, but leaving the parser in so we have full response object support.

Next push is probably another TODO/FIXME review and chunked item support.

It might be nice to pull some of the core changes out into a separate PR.

dormando:
Did some git magic: The different functional changes are now in ordered commits.

I'll be peeling off the core commits and updating + upstreaming them independently. This branch will then shrink until it's just the `mcp.internal(r)` change.

dormando:
Note to self: split the resp changes out of the IO API commit, turns out they aren't cross related like I thought. Pull the IO changes first.

dormando:
rebased on the hopefully completed `conn_removal` branch + next.

going to split the IO API commit and then do a tarball cut, merge conn_removal, etc.

dormando:
think I can split this a little more; some small resp changes are still in the big mcp.internal commit.

Then from the mcp.internal commit, I change the coroutine yielding to use a flag. If that part is pulled out and upstreamed the rest of the change should be free of merge conflicts if I need to switch context for a while.

dormando:
not much productive time today... still need to split the last few changes more and review/merge those. two of the pre-reqs are upstreamed now and the change looks a lot cleaner.

dormando:
This change has now been minimized. It's still "large", but only because of the command implementations being copy/pasted from `proto_text.c` and updated to use the new parser interface.

I'm going to hold this un-merged for a little while longer so I can pivot to other changes for a bit. Mainly I want to do a bunch of test-suite work, so I can add a suite of unit tests to this before merging.

dormando:
needs a rebase, and still pending the test suite. probably not going to finish this until I have a good loadgen workload running... or if it suddenly becomes urgent I guess :)

dormando:
rebased... still pending test suite and chunked support.

dormando:
rebased and pushed again.

proxy_internal.c
```diff
@@ -0,0 +1,1660 @@
...
+/*** Lua and internal handler ***/
+
+int mcplib_internal(lua_State *L) {
+    luaL_checkudata(L, 1, "mcp.request");
+    mcp_resp_t *r = lua_newuserdatauv(L, sizeof(mcp_resp_t), 0);
+    memset(r, 0, sizeof(mcp_resp_t));
+    luaL_getmetatable(L, "mcp.response");
+    lua_setmetatable(L, -2);
+
+    // TODO: mcp_queue_io's request flattening.
```
dormando, line 1536: Pretty sure this comment can be removed; requests are immediately flattened on edit now.

dormando:
think all I have to do for this iteration of the feature is make chunked items work. I can merge it even without 100% confidence and minimal test coverage; if you don't use the feature it's completely inert.

dormando:
rebased on top of workerio... maybe I can do the chunked stuff tomorrow...

dormando:
Chunked items work. Did some manual testing with injecting errors to ensure slab chunks got cleaned up during error.

I'm pretty sure it can't hurt to merge it since it doesn't affect anything if unused. Will continue to add and run unit and bench tests to build confidence without having to maintain a branch, or ask users to test a branch.

PR #973, opened 2023-01-01
change signal handlers using write syscall instead of printf.

devnexen:


memcached.c
```diff
@@ -4277,15 +4277,25 @@ static void remove_pidfile(const char *pid_file) {
 
 static void sig_handler(const int sig) {
     stop_main_loop = EXIT_NORMALLY;
-    printf("Signal handled: %s.\n", strsignal(sig));
+    char *strsig = strsignal(sig);
```
en4bz, line 4280: `strsignal` is not reentrant.

dormando:
Thanks!

To be honest I'm not sure I intended to leave printf's in the signal handlers. the side of the code picking up the signal should probably just printf instead of all this workaround..

dormando:
Looking again at the `switch (stop_main_loop)` code I think these two prints can just be removed.

Maybe add an `if (verbose)` for `EXIT_NORMALLY` and move the term print there.

dormando:
better, thanks!

dormando:
released in 1.6.18

PR #974, opened 2023-01-06
proxy: log time now relative to resp lifetime

dormando:
Originally I envisioned taking an inbound request object, tagging it with the time, and at the very end of a function logging is called. This would give you the total time of the "backend" part of a request.

On rethinking, the timing information that's most useful in the proxy's perspective is the time it takes for a response to happen + the status of a response. One request may generate several sub-responses and it is impossible to check the timing of each of those and log outliers.

You now cannot get the total time elapsed in a function anymore, but I believe that is less useful information to the user of a proxy. The best picture of latency will still be from the client, and response latency can educate the proxy on issues with backends.

resp:elapsed() has been added as a compromise; it returns the elapsed microseconds that a response took, so you can add the time together and get an approximation of total time (if running req/resp's sequentially).

This change also means that calling mcp.await() and waiting for multiple responses will give the timing of each sub-response accurately.

Some more work:
- [ ] Adding some new logger versions of await()
- [ ] Raise lua error if "detail" is too long instead of checking length deep in the logger code.
- [ ] Maybe: allow calling log without a request object? Would just have blanks in the fields.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -300,7 +300,13 @@ void proxy_return_cb(io_pending_t *pending) {
     if (p->is_await) {
         mcplib_await_return(p);
     } else {
+        struct timeval end;
         lua_State *Lc = p->coro;
+
+        // stamp the elapsed time into the response object.
+        gettimeofday(&end, NULL);
+        p->client_resp->elapsed = (end.tv_sec - p->client_resp->start.tv_sec) * 1000000 +
+            (end.tv_usec - p->client_resp->start.tv_usec);
 
         // in order to resume we need to remove the objects that were
         // originally returned
@@ -840,6 +846,7 @@ static void mcp_queue_io(conn *c, mc_resp *resp, int coro_ref, lua_State *Lc) {
     memset(r, 0, sizeof(mcp_resp_t));
     r->buf = NULL;
     r->blen = 0;
+    gettimeofday(&r->start, NULL);
     // Set noreply mode.
     // TODO (v2): the response "inherits" the request's noreply mode, which isn't
     // strictly correct; we should inherit based on the request that spawned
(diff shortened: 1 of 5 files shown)


dormando:
Thinking of breaking the await changes into another PR or issue... Been playing with the API's and I hate it:
- Lots of chicken/egg problems internal to the code to safely allocate/reference various things
- All of the re-checking and referencing is bothering me; a slightly different API could net a small perf improvement. I wanted to do similar a while ago but wasn't sure.

Thought:
- In the typical code examples, a route is generated by returning a function with some static upvalues (list of pools and some arguments, mainly)
- At this point an "await descriptor" or "log descriptor" object could be created, given the common arguments and validate them (are all of these things pools? etc)
- If just logging, should speed things up considerably. Could avoid copying the detail string many times (and encourage people to not create strings inline with a route).
- Further, could use lua tricks to remove the result table allocation; `await_obj(r)` would return an "await res" object, that could be used like a table. Then internally we can remove the lua response table.
- Would remove the constant referencing of the argument table (list of pools) too.

Problem is I'm sure this isn't a huge CPU savings...

The upside is if I'm adding more await options for logging/etc this is probably a good time to adjust the API.

dormando:
All the data copying with "detail" and tags is driving me nuts too. I've been trying to keep the API's designed such that when I come in to optimize them, they can be optimized. Long detail strings just cannot be optimized and will result in a lot of chunky data copies. In the grand scheme I doubt this adds up to much but it bothers me and puts an upper limit on how much optimizing we can do later.

I don't have any great ideas on how to mitigate this yet. I'm "mostly okay" with some kind of string being copied when a log line is actually being generated, so some kind of string registration system could work (kind of similar to how stats works?). The problem is not making that too onerous on the developer.

Using the aforementioned modified API might just do this well enough. "await res" objects would need to hold a reference to the "await descriptor" object which I am trying to avoid.

I'm probably off in the weeds at this point, but I could create a "string reference" table that the await/log descriptor objects use. When passed a detail string they register it and get back an integer. That string lives forever. Do we need to clean up descriptor strings? do they change that often? I'm sure someone will rotate them with every request and ruin my day.

dormando:
I think this is mergable already. Requires no user lua changes, barely moves code around internally, doesn't change the log prints at all.

PR #977, opened 2023-01-09
proxy: iterate modified request handling

dormando:
had an API bug where updating a token and then requesting it again returns the previous token. Also have other branches which require use of the flattened request.

I was originally doing this using a lot of lua but just copying the string a few times has some better properties:
1) should actually be faster with less lua + fewer allocations 
2) can be optimized to do minimal copying (avoid keys, append new flags, etc)
3) might be optimized to avoid re-parsing (or do minimal re-parsing) in some cases.

TODO:
- [x] A few more tests before merging.
- [ ] Update docs to reflect that updated tokens are now returned.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -283,8 +283,6 @@ struct mcp_request_s {
     mcp_parser_t pr; // non-lua-specific parser handling.
     mcp_backend_t *be; // backend handling this request.
     bool ascii_multiget; // ascii multiget mode. (hide errors/END)
-    bool was_modified; // need to rewrite the request
-    int tokent_ref; // reference to token table if modified.
     char request[];
 };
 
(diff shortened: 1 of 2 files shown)


dormando:
need this to support updated request objects in the internal code, but it was easy to split out so I'm reviewing and merging it ahead of the main change.

dormando:
Fixed an off-by-one in raw token access.
Made sets work and armored against some buffer issues.

dormando:
Fixed excess space when removing tokens.

Blast radius is limited with this code. Need to do better test suite soon.

dormando:
Think this change is done... I did a bunch of tests and fixed/updated a few things. need to merge it soon since it fixes a confusing API bug.

dormando:
I've tested this a bunch. real unit tests would improve confidence more and I will fill those in soon...

I'm going to merge this to next now. The code is a lot simpler and it fixes a real bug that I keep running into.

PR #978, opened 2023-01-10
core: remove *conn object from cache commands

dormando:
We want to start using cache commands in contexts without a client connection, but the client object has always been passed to all functions.

In most cases we only need the worker thread (LIBEVENT_THREAD *t), so this change adjusts the arguments passed in.

This is a prerequisite change for #969 - I'm splitting into its own PR to focus testing. The main test suite covers these functions very thoroughly, so I feel confident in merging this soon.

TODO:
- [x] Restore LOG_FETCHERS
- [x] Restore LOG_MUTATIONS
- [x] Fix intermittent binary protocol test failure. (testapp.c:1255)

This change is for the most part trivial; in core item fetch/storage functions which were passed `conn *c` typically only used `c->thread`, so we now pass that directly. There were only a handful of places which needed the `conn` object, notably cuddling CAS value during binary protocol parsing. This makes the change look large but very little code was materially changed.

There's still a bug in binprot though, and I have to add LOG_FETCHERS and LOG_MUTATIONS back in higher level callers, as those calls printed `c->sfd` (client file descriptor).


dormando:
Note: meta cas return on set also broken.

dormando:
Not 100% sure why the binary protocol would only _sometimes_ fail, but this commit matches the previous behavior exactly and everything works now. Ran testapp a few hundred times in a loop without fail.

Looping the full test suite now. Just need to add LOG_FETCHERS and LOG_MUTATIONS back and this is mergable, though I may hold it a bit longer.

dormando:
LOG_MUTATIONS: fixing this is relatively easy. Very few callers, and all of the callers are running the log command directly in the case of a failed store.

LOG_FETCHERS: this has a lot of callers through various semi-convoluted calls (`item_get_limited`, `item_get_locked`, etc). The log code mostly needs the item data (key/len/*it info), plus the client file descriptor.
- `do_item_get` logs why an item was missed it if was fetched, which would need to surface upwards.
- `do_item_get` logs the client fd, which requires the conn object.

LOG_FETCHERS has always been more debug info than actual usable info, like "what commands/keys are actually being executed".

For purposes of this change we should probably not try to solve the semantics issue. Options:
- Pass in sfd to item_get(), which sort of sucks.
- Add `LIBEVENT_THREAD->last_sfd` and tag the sfd at the top of process_* commands.

Leaning toward the latter since it's a more minimal change, but it feels more hacky. This lets us use a -1 for sfd in the proxy code for emitting a more proper log line.

dormando:
Fixed up and re-squished. Going to merge this to next a bit earlier than I would normally leave it. It's pretty well tested so the risk is low.

PR #979, opened 2023-01-11
core: simplify background IO API

dormando:
- removes unused "completed" IO callback handler
- moves primary post-IO callback handlers from the queue definition to the actual IO objects.
- allows IO object callbacks to be handled generically instead of based on the queue they were submitted from.

TODO:
- [x] Benchmark extstore with/without this to ensure there's no major regression. (tests roughly similar, no huge measurable change while using heavy multigets)

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -569,7 +569,7 @@ void conn_worker_readd(conn *c) {
     }
 }
 
-void thread_io_queue_add(LIBEVENT_THREAD *t, int type, void *ctx, io_queue_stack_cb cb, io_queue_stack_cb com_cb, io_queue_cb ret_cb, io_queue_cb fin_cb) {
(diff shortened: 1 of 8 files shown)


dormando:
Prerequisite for #969

dormando:
Reviewed this again and I like it still. There were too many different flows for the IO subsystem, and one of them changes the thread ownership of some of the structure. That is no longer true so the whole thing is more straightforward.

It's still not an abstraction I'm happy with, as the mcp.internal code re-implements multiple IO object sub-types again, but maybe we'll get it next time :)

Stilll need to benchmark to check on no regression. For non heavily pipelined workloads there should be no difference.

dormando:
benchmark's fine.

PR #980, opened 2023-01-12
core: remove *c from some response code

dormando:
Allow freeing client response objects without the client object. Clean some confusing logic around clearing memory. Also exposes an interface for allocating unlinked response objects.

Mostly trivial patch that just changes more `*c->thread` to `*t`, but also removes a difficult to read memset and cleans the code a bit. Easy to verify.


dormando:
On reviewing a third time; I am tempted to golf this code shorter but will resist so I don't potentially add bugs right now.

It's certainly more clear than it was and it will be much easier to run a bunch of C unit tests against it if it's not pulling in the connection object.

PR #981, opened 2023-01-12
proxy: clean logic around lua yielding

dormando:
We were duck typing the response code for a coroutine yield before. It would also pile random logic for overriding IO's in certain cases. This now makes everything explicit and more clear.


dormando:
This also exposes an error printer.

Shouldn't be any complaints about merging once I fix an issue from rebasing.

I spent quite a while re-studying this code while doing `mcp.internal(r)` so I wanted to make sure it was more clear. Felt like a source for a difficult to track down bug.

dormando:
Note: was a pre-req for #969 - with this out of the way that change is very minimal outside of the new file.

PR #982, opened 2023-01-13
proxy: allow workers to run IO optionally

dormando:
`mcp.pool(p, { iothread = false }`

By default the IO thread is used. When disabled a backend connection is created for each worker thread. This can be overridden by setting `iothread = false` when creating a pool.

`mcp.pool(p, { beprefix = "etc" }`

If a `beprefix` is added to pool arguments, it will create unique backend connections for this pool. This allows you to create multiple sockets per backend by making multiple pools with the same list of backends, but different prefixes.

There are legitimate use cases for sharing backend connections across different pools, which is why that is the default behavior.

-------------------
This builds on top of https://github.com/memcached/memcached/pull/959 - which was the hard part in making this all work.

TODO:
- [x] I want to add per-pool backend timeout overrides. This might only work when "beprefix" is in use because it doesn't make a lot of sense otherwise.
  Going to do this in a different PR.
- [x] `AWAIT_BACKGROUND` gets broken by this change in the work model. I tried to fix it but not liking the method.
- [x] Find worker hang bug (not sure where I caused this yet and the trace broke down in libevent)
- [x] Decide if avoiding the io thread by default is really the best idea.
- [x] Ensure daemon exits if started with invalid config
- [x] I think bad backend handling is still broken (need to handle after the source loop so `q->count` doesn't reach 0 early)
- [x] If you're just overriding iothread or beprefix settings, pool creation whines about missing `dist`. Pull the default if missing.
- [x] Code audit.
- [x] Specific unit tests using new proxyunits framework.

Notes:
- With `beprefix` you can do a lot by creating pool copies, such as dedicating sockets for gets vs sets. Just have to structure the lua differently.
- This will allow reasonable scaling beyond 4 worker threads. However, the IO thread is very efficient due to batching effects, so CPU usage may be higher. Backend memcached nodes will also be less efficient with more sockets.
- There are a lot of pending optimizations in the proxy code so while you will get more performance it may not match expectations.
- Latency should be lower, but I'm not sure by how much right now.

Details:
- Most of the LOC is from moving some pre-existing code around unchanged.
- The rest of the bulk of the change is from changing the pool init code: how the backend connection cache works, etc. Minor change to the request code paths.
- This is implemented by tacking a "proxy event thread" struct into the main thread workers. This means most of the code for actually handling backend IO's is completely identical.


dormando:
crap I can't seem to trigger the hang anymore :) uhhh oh well. hopefully it'll pop up.

dormando:
well I ran benchmarks on the IO thread and worker threads simultaneously for 30 minutes without a hang. I'll finish the rest of the work and come back to that.

In testing this I've found AWAIT_BACKGROUND may never have worked right, which is awful because I did spend some time testing that. It seems like it still waits on the first request, which I had to patch to make it work with worker threads.

Will go back to the `next` branch and see if that's reproducable, and split the fix out if so.

dormando:
AWAIT_BACKGROUND does seem to work right in `next` so I probably just broke it in a weird way here.

Removed an errant `dump_stack()` call just now

dormando:
thinking of leaving iothread on as the default for first merge, but flipping the default at a no-too-later date, just for testing paranoia.

dormando:
reproduced a crash by killing/starting backends while benchmarking. prooooobably the bad backend issue.

dormando:
bug is worse than I was hoping. it's acting like I'm accessing an event base from multiple threads but I've been over the code a hundred times and I don't see it.

dormando:
Okay I knew it was stupid:
- backend dies
- _reset_bad_backend()
- loops IO's associated with backend
- runs the io return callback inline
- which calls lua
- which does stuff
- which then calls conn_worker_readd() to resume the client
- ... which immediately calls drive_machine()
- which can run more lua
- which then calls submit_cb()
- ... which can mess with the same backend's IO list
- ... and then _rest_bad_backend() resumes and finds the same IO again, now corrupting things.

I forgot about the immediate resume of drive_machine and the crash was causing gdb to not let me step through the reset code somehow.

I've simply removed the special case code for returns; that can be optimized later, but objects will still have to be queued for resume later instead of resumed inline in most cases.

dormando:
Making the await code less convoluted is going to have to wait for more thought on the queuing system. I generally know what to do but it's too much code change for now.

dormando:
Considering this largely done for now. I'll audit it again after letting it sit just in case. It needs some user testing before merge.

Going to do some of the other planned work in subsequent PR's.

dormando:
rebased on top of `next` branch so we can reliably regression test.

still planning on one more code audit before merge, but at this point the blocker is just needing time in the meat grinder.

Which it should be getting more of this week, potentially letting it merge end of week or early next.

dormando:
Rebased again on top of the new testing framework, so now I can add some basic unit tests at least.

Still in the queue for burn-in tests.

dormando:
Taking a little longer to get the burn-in tests done while I try to level-up the whole process.

dormando:
Note: this now has some merge conflicts I need to resolve to pull it up on `next`

dormando:
I've rebased this over `next`. Noting that **unit tests do not pass right now** - it looks like AWAIT_BACKGROUND may be broken. will check tomorrow.

dormando:
`AWAIT_BACKGROUND` works again. I broke it while fixing the race bug and apparently hadn't run the unit tests after rebasing it?

Anyway cool. now when I get a moment I can add more specific tests and burn-in tests.

dormando:
Rebased over `next` changes and pushed.

Really depressed how long this has been queued for merge :(

proxy_lua.c
```diff
@@ -512,43 +513,47 @@ static void _mcplib_pool_dist(lua_State *L, mcp_pool_t *p) {
...
-    p->ctx = lua_touserdata(L, lua_upvalueindex(MCP_CONTEXT_UPVALUE));
-
-    luaL_setmetatable(L, "mcp.pool");
-
-    lua_pushvalue(L, -1); // dupe self for reference.
-    p->self_ref = luaL_ref(L, LUA_REGISTRYINDEX);
-
-    // TODO (v2): move to after function check so we can find the right
-    // backend label to look up.
+// TODO: need to figure out how to index offsets here.
```
dormando, line 516: note: can remove this comment.

proxy_lua.c
```diff
@@ -512,43 +513,47 @@ static void _mcplib_pool_dist(lua_State *L, mcp_pool_t *p) {
...
+        lua_concat(L, toconcat);
+
+        lua_pushvalue(L, -1); // copy the label string for the create method.
         mcp_backend_wrap_t *bew = _mcplib_backend_checkcache(L, bel);
         if (bew == NULL) {
-            bew = _mcplib_make_backendconn(L, bel);
+            bew = _mcplib_make_backendconn(L, bel, t);
         }
         s->be = bew->be; // unwrap the backend connection for direct ref.
+        if (p->use_iothread) {
```
dormando, line 552: logic used to be inverted. should just change line to: `bew->be->use_io_thread = p->use_iothread`

memcached.c
```diff
@@ -3321,7 +3321,8 @@ static void drive_machine(conn *c) {
                 if (q->stack_ctx != NULL) {
                     io_queue_cb_t *qcb = thread_io_queue_get(c->thread, q->type);
                     qcb->submit_cb(q);
-                    c->io_queues_submitted++;
+                    if (q->count != 0)
```
dormando, line 3324: took me a while to remember why this got added, need to add a comment:
- for await_background case, it short-circuits and returns the IO without increasing `q->count`
- so if the only requests submitted here are await_background, we need to not suspend the client as there's nothing to wait for

proto_proxy.c
```diff
@@ -256,33 +227,59 @@ void proxy_submit_cb(io_queue_t *q) {
         } else {
             q->count++;
         }
+        be = p->backend;
+
+        if (be->use_io_thread) {
```
dormando, line 232: Think this is the section I was agonizing over the most. It's honestly good enough for now at least, I just wish I could reuse more of the `proxy_event_handler` code.

This is more efficient than reusing the event handler code since that loops a second time to split queues onto backend objects.

Once I split the backend IO queues into to-write and to-read it's a little tempting to make the to-write queue locked, so they don't have to be iterated again post-submission.

dormando:
did another review... need to rebase again once #999 is merged, then finally write some tests. I think code wise it all makes sense and I did a decent job at reducing the size of it as much as reasonable. If the queueing system were a better abstraction it would be even less code.

dormando:
it's rebased over `next` again, which fixes the event handling a lot. that actually gives me some more confidence in the code, but it still needs a little more work:
- address my review comments from earlier today
- tests for the new specific syntax
- duplicate a few tests for workerio

memcached.c
dormando: actually nevermind, that was extraneous from a prior attempt at allowing zero queues. I've removed the line and run some benchmarks and it seems fine.

dormando:
Ran some benchmarks... it does scale better than IO thread but the perf improvement isn't outsized.

ie; for 4 worker threads the throughput is similar. for 8 it'll be higher but not necessarily double? need to run more tests.
sampled elapsed time in micros is a bit shorter on average for underloaded tests: 60-80 micros localhost vs 80-120 micros.

Once you add in bandwidth delay of networks and things like head of line blocking for pipelined requests worker-io performance should beat the IO thread more often.

started writing unit tests but not super sure what to write beyond basic tests. The actual event handling code is 100% the same between IO thread and worker IO, it's just the submission queues and number of backends that it creates which are different.

Probably worth testing AWAIT_BACKGROUND + worker io in specific since that's the weirdest submission path... but it's still identical between IO and no IO thread.

dormando:
I've spent so much time rebasing this, minimizing the change, benchmarking it, etc. I'm reasonably sure that:

- The changes required for this are common to both code paths for io-thread and non-io thread. (ie; the await changes)
- This means existing test coverage is pretty good
- Benchmarking is probably the most effective tool to find bugs here
- Most of the things I ran into that I thought were bugs in this code turned out to be bugs in the main code (which are now fixed)

I'm too tired right now but I'm going to try a _little _ more tomorrow and then just merge it.

proxy_await.c
```diff
@@ -259,7 +259,11 @@ int mcplib_await_run(conn *c, mc_resp *resp, lua_State *L, int coro_ref) {
     const char *key = MCP_PARSER_KEY(rq->pr);
     size_t len = rq->pr.klen;
     int n = 0;
-    bool await_first = true;
+    // TODO (v3) await_first is used as a marker for upping the "wait for
+    // IO's" queue count, which means we need to force it off if we're in
+    // background mode, else we would accidentally wait for a response anyway.
+    // This note is for finding a less convoluted method for this.
+    bool await_first = (aw->type == AWAIT_BACKGROUND) ? false : true;
```
dormando, line 266: note to self: I think this line can be removed, since we're checking for an AWAIT_BACKGROUND object before await_first, and they're handled the same now.

dormando: Ah nevermind I see it again. it would increase the `q->count` twice.

guessing I should have the IO objects have an "increase the q->count" flag and just blindly add it to q->count, removing all of this test logic on submit_cb. even better might be adding a "next_count" counter to the IO queue and have callers tick that directly. Then copy or add it to q->count on the next submit. You can't kick q->count directly because it causes the system to never resume.

think that's a separate change.

dormando:
The only other test I can think of is to check that the pool mapping aligns in io vs non io:

1) fetch 100 keys through IO thread, map which backend they come out to
2) fetch the same keys through same pool but no io thread
3) make 3 new client connections and repeat (to test each worker offset)

PR #983, opened 2023-01-17
proxy: new integration tests.

dormando:
uses mocked backend servers so we can end-to-end test syntax, basic error conditions, and routing. instead of trying to use a memcached and having to poke/fetch/modify it, we just take a note from `expect` and test that the sockets are seeing what they should.

going to use this pattern of mocking backends and test-specific prefix routing lua to fill out integration tests for a few major scenarios:
- [ ] Basic protocol test, routing (goes to backend B instead of A, etc), error coverage (weird/garbage from server).
- [ ] Config reload tests (use temp files and/or env vars to adjust lua config)
- [ ] Key distribution functions (jump, ketama, ensuring keys land on different backends repeatably, etc)
- [ ] All the different `mcp.await` scenarios (early return, backgrounding, etc)

Each major group above will likely be different sets of files.

Already seeing a weird pause just a few tests in... this will probably take longer than it would look as I stop to understand/fix bugs.

After this set of tests I want to move on to a burn in test suite based off of mc-crusher's bench suites. That will test varying workloads and error conditions (ie; backends being killed / config reloaded while under light/med/heavy traffic)


dormando:
Thought I'd have all the basic commands typed up by now but I already ran into a weird multi-second pause when returning the VALUE line, which I will have to continue investigating tomorrow :)

dormando:
Lol guess I was tired. Did find a real bug which confused me: I was sending a response to `get` without the `END\r\n`, which caused the backend to time out but then send a partial response to the client which it shouldn't do.

I'll add a real test for that later and come back to fix it.

dormando:
got derailed into a debugging session. rebased and continuing on the reload tests today.

dormando:
thinking I'll do super super basic key dist tests so I can move to loadgen tests later today.. once I get those kicked off I'll come back and fill more tests.

ideally I should be able to merge these tests back by EOD today or tomorrow at the latest. There's no downside to merging them since it's just tests.

dormando:
Not been feeling great this week. Taking a while to get through these.

The config reload tests are tricky: need to structure so I can inspect failures easily, which isn't great if I'm writing config from the perl tests.

going to restructure again so the config is always in the lua file and an external file is read to do a "mode switch". Then a distinct test file will validate code change on reload. Most of what we're validating in this file are backend/pool/route changes, `mcp.attach` replacements, etc.

dormando:
happy with the structure of config reload tests now but I'm prety behind and will need to circle back to add more tests.

think I will skip the key dist tests for now. squash and merge so I can use the structure to write tests on PR's.

PR #985, opened 2023-01-21
proxy: fix stats deadlock and mismatched responses

dormando:
- specifically the WSTAT_DECR in proxy_await.c's return code could potentially use the wrong thread's lock

This is why I've been swapping c with thread as lock/function arguments all over the code lately; it's very accident prone.

Am reasonably sure this causes the deadlock but need to attempt to verify more.

- Also fixes a bug I found while trying to reproduce the stats deadlock. See commit message for details there.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -402,7 +402,7 @@ void proxy_cleanup_conn(conn *c) {
     lua_State *L = thr->L;
     luaL_unref(L, LUA_REGISTRYINDEX, c->proxy_coro_ref);
     c->proxy_coro_ref = 0;
-    WSTAT_DECR(c, proxy_req_active, 1);
+    WSTAT_DECR(thr, proxy_req_active, 1);
 }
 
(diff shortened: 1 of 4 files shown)


dormando:
There're two paths for this bug:

1) Wrong thread lock is used, which could cause the number to go weird.
2) Thread changes inbetween the lock and unlock calls, so it would lock one mutex then never unlocked it, or unlock a previously locked mutex.

I got it to reproduce off of `next` once synthetically but it's not happening a second time. Hoping to get it reproducing a bit better so I can validate the patch before merge...

dormando:
Finally reproduced on `next`. fix seems to be holding so far.

dormando:
Fixed another bug I found while trying to reproduce the initial bug... I hate the flush code :(

dormando:
Branch survived an 8ish hour burn-in on a load pattern that would normally trigger either one in under 30 minutes.

PR #987, opened 2023-01-26
proxy: add mcp.await_logerrors()

dormando:
Logs any backgrounded requests that resulted in an error.

TODO
- [ ] code audit
- [ ] a little more testing (another unit test? quick bench to ensure no memory leak/weirdness)


dormando:
Note: double check that the log condition matches `!resp:ok()` and log only if await object has been completed already.

also hell yes for having good unit test structure. took like five minutes to write the whole test.

dormando:
conditions match `resp:ok()`'s check.

dormando:
Updated to only log errors after the await object has completed. forgot to push last night.

dormando:
I'm happy to merge this but will hold it a little longer.

Edited the unit test a little and did some manual testing:
- running bench-warmer against a daemon running the proxyunits.lua config to just exercise the `/awaitlogerr/` path a while.
- two local memcached's + one memcached with some stub lua code:
```
function mcp_config_pools(old)
    return {}
end

function mcp_config_routes(zones)
    mcp.attach(mcp.CMD_SET, function(r) 
        os.execute("sleep 1")
        return "SERVER_ERROR always failing\r\n" 
    end)
end
```

doesn't seem to leak memory, but the read timeouts are working kind of oddly (seeing bursts in timing instead of capping at whatever I've set the read timeout to). That issue isn't caused by this change so I'll have to follow up on that.

It does seem to log properly.

Issue #932, opened 2022-09-06
Autoconf fails in v1.6.17+


lamby:
Calling `autoconf` with version 1.6.17 results in the following:

```
configure.ac:8: warning: 'AM_CONFIG_HEADER': this macro is obsolete.
configure.ac:8: You should use the 'AC_CONFIG_HEADERS' macro instead.
./lib/autoconf/general.m4:2434: AC_DIAGNOSE is expanded from...
aclocal.m4:745: AM_CONFIG_HEADER is expanded from...
configure.ac:8: the top level
configure.ac:181: warning: The macro `AC_TRY_COMPILE' is obsolete.
configure.ac:181: You should run autoupdate.
./lib/autoconf/general.m4:2847: AC_TRY_COMPILE is expanded from...
lib/m4sugar/m4sh.m4:692: _AS_IF_ELSE is expanded from...
lib/m4sugar/m4sh.m4:699: AS_IF is expanded from...
./lib/autoconf/general.m4:2249: AC_CACHE_VAL is expanded from...
./lib/autoconf/general.m4:2270: AC_CACHE_CHECK is expanded from...
configure.ac:138: AC_C_DETECT_SASL_CB_GETCONF is expanded from...
configure.ac:181: the top level
configure.ac:182: warning: The macro `AC_TRY_COMPILE' is obsolete.
configure.ac:182: You should run autoupdate.
./lib/autoconf/general.m4:2847: AC_TRY_COMPILE is expanded from...
lib/m4sugar/m4sh.m4:692: _AS_IF_ELSE is expanded from...
lib/m4sugar/m4sh.m4:699: AS_IF is expanded from...
./lib/autoconf/general.m4:2249: AC_CACHE_VAL is expanded from...
./lib/autoconf/general.m4:2270: AC_CACHE_CHECK is expanded from...
configure.ac:161: AC_C_DETECT_SASL_CB_GETCONFPATH is expanded from...
configure.ac:182: the top level
configure.ac:354: warning: The macro `AC_TRY_LINK' is obsolete.
configure.ac:354: You should run autoupdate.
./lib/autoconf/general.m4:2920: AC_TRY_LINK is expanded from...
lib/m4sugar/m4sh.m4:692: _AS_IF_ELSE is expanded from...
lib/m4sugar/m4sh.m4:699: AS_IF is expanded from...
./lib/autoconf/general.m4:2249: AC_CACHE_VAL is expanded from...
[...]

PR #988, opened 2023-01-29
[devtools] Add new pkg-config dependency to dockerfiles

olof-nord:
Add new pkg-config dependency to dockerfiles

Attempting to build an image from the devtools Dockerfiles currently fails when executing ./autogen.sh.

As is documented in https://github.com/memcached/memcached/issues/932 memcached, and as such also the images, does not build without `pkg-config`.

To verify: run `docker-compose build` on master (build breaks) - execute `docker-compose build` on this branch and the images are built successfully.

### DIFF
--- a/devtools/Dockerfile.alpine
+++ b/devtools/Dockerfile.alpine
@@ -2,7 +2,7 @@ FROM alpine:latest
 
 ARG CONFIGURE_OPTS="--enable-seccomp"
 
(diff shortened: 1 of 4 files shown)


olof-nord:
cc @dormando as you are in the git history :)

olof-nord:
You are welcome! What process is there for this to be merged?

dormando:
Sorry, you don't need to do anything. I'm just really behind on verifying patches.

PR #989, opened 2023-02-01
proxy: add mcp.backend(t) for more overrides

dormando:
ie:
```
local b1 = mcp.backend({ label = "b1", host = "127.0.0.1", port = 11511,
    connecttimeout = 1, retrytimeout = 0.5, readtimeout = 0.1,
    failurelimit = 11 })
```

... to allow for overriding connect/retry/etc tunables on a per-backend basis. If not passed in the global settings are used.

TODO:
- [x] add at least basic tests: I added some test lua and was able to hand verify a bit.
- [x] find the best way to test the overrides are actually being changed. I started by hand and was able to observe the retrytimeout was different across backends.

Note:
- The tradeoff: previously it was possible to change `mcp.backend_read_timeout` and similar without the proxy internally recreating any backends, but now modifying the global settings will re-create all backend objects. This makes reload slower when timeouts are adjusted; realistically this isn't done often and reload is still generally fast.
- If you are _not changing the global settings_, but _only changing individual backend overrides_, only those overridden backends will be recreated on reload. It is still fast.
- This actually fixes a long standing issue where toggling the `tcp_keepalive` function didn't work for reload.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -175,8 +175,6 @@ void *proxy_init(bool use_uring) {
         pthread_mutex_init(&t->mutex, NULL);
         pthread_cond_init(&t->cond, NULL);
 
-        memcpy(&t->tunables, &ctx->tunables, sizeof(t->tunables));
-
 #ifdef HAVE_LIBURING
         if (t->use_uring) {
             pthread_create(&t->thread_id, NULL, proxy_event_thread_ur, t);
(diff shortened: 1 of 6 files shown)


dormando:
Looks like for testing the best bet is just checking if the listener sockets become readable after reload completes (ie; new conn waiting for accept()).

For testing timeouts maybe just long vs very very short so I don't have to sleep the test script much.

dormando:
Tests are added. Had a small hiccup with the test monitor daemon killing the process on SIGHUP.

 I did some fiddling with the tests and I'm reasonably sure this is safe to merge. The tests observe that multiple tunable arguments are different between the backends, that the backend connection cache works, and that the arguments differ from the global settings.

dormando:
Found a little more code to delete on audit. Going to merge now.

PR #991, opened 2023-02-09
proxy: fix write flushing bugs

dormando:
1) the event flags were not being used from the result of the
   flush_pending_write() function, causing it to not listen on WRITE
   events in some cases.
2) be->can_write flag was not being set on short writes; old iterations
   of code would loop-write until EAGAIN but this does not.
   This prevented the code from listening for WRITE events.
3) if many short writes are happening, a successful read event was
   overwriting the WRITE event with a READ event, instead of READ|WRITE.
   fixed via the be->can_write flag fix from 2).

@feihu-stripe

### DIFF
--- a/proxy_network.c
+++ b/proxy_network.c
@@ -856,7 +856,7 @@ static void proxy_event_handler(evutil_socket_t fd, short which, void *arg) {
                 _reset_bad_backend(be, P_BE_FAIL_WRITING);
                 _backend_failed(be);
             } else {
-                flags = be->can_write ? EV_READ|EV_TIMEOUT : EV_READ|EV_WRITE|EV_TIMEOUT;
+                flags |= EV_TIMEOUT;
                 _set_event(be, t->base, flags, be->tunables.read, proxy_backend_handler);
             }
         }
@@ -1276,12 +1276,6 @@ static bool _post_pending_write(mcp_backend_t *be, ssize_t sent) {
         bool flushed = true;
         if (io->flushed)
             continue;
-        if (sent <= 0) {
-            // really shouldn't be negative, though.
-            assert(sent >= 0);
-            break;
-        }
-
         if (sent >= io->iovbytes) {
             // short circuit for common case.
             sent -= io->iovbytes;
@@ -1306,10 +1300,15 @@ static bool _post_pending_write(mcp_backend_t *be, ssize_t sent) {
         if (flushed) {
             did_flush = flushed;
         }
+        if (sent <= 0) {
+            // really shouldn't be negative, though.
+            assert(sent >= 0);
+            break;
+        }
     } // for
 
     // resume the flush from this point.
-    if (io != NULL) {
+    if (io != NULL && !io->flushed) {
         be->io_next = io;
     } else {
         be->io_next = NULL;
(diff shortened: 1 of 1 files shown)


proxy_network.c
```diff
@@ -856,7 +856,7 @@ static void proxy_event_handler(evutil_socket_t fd, short which, void *arg) {
                 _reset_bad_backend(be, P_BE_FAIL_WRITING);
                 _backend_failed(be);
             } else {
-                flags = be->can_write ? EV_READ|EV_TIMEOUT : EV_READ|EV_WRITE|EV_TIMEOUT;
+                flags |= EV_TIMEOUT;
```
feihu-stripe, line 859: For my understand, this corresponds to your first point, so that now we use the `flags` returned by `_flush_pending_write()` and only append `timeout`.

correct me: it appears to me that the previous logic is still valid?

trivial: let `_flush_pending_write()` return flags already contains timeout event.

proxy_network.c
```diff
@@ -1306,10 +1300,15 @@ static bool _post_pending_write(mcp_backend_t *be, ssize_t sent) {
...
+        if (sent <= 0) {
+            // really shouldn't be negative, though.
+            assert(sent >= 0);
+            break;
+        }
     } // for
 
     // resume the flush from this point.
-    if (io != NULL) {
+    if (io != NULL && !io->flushed) {
```
feihu-stripe, line 1311: This makes sense.

Wonder if the benefit is to remove redundant loop by adding this check?

proxy_network.c
```diff
@@ -1334,6 +1333,7 @@ static int _flush_pending_write(mcp_backend_t *be) {
         }
         // still have unflushed pending IO's, check for write and re-loop.
         if (be->io_next) {
+            be->can_write = false;
```
feihu-stripe, line 1336: For my understanding, what do you mean by short writes? Do you mean the writes that still have pending IO left?

proxy_network.c
```diff
@@ -1510,6 +1510,9 @@ static void proxy_backend_handler(const int fd, const short which, void *arg) {
     // Still pending requests to read or write.
     if (!STAILQ_EMPTY(&be->io_head)) {
         flags |= EV_READ; // FIXME (v2): might not be necessary here, but ensures we get a disconnect event.
+        if (!be->can_write) {
+            flags |= EV_WRITE;
+        }
```
feihu-stripe, line 1515: Bravo! 

Is this the cause of write halt we saw in tcpdump?

feihu-stripe (approved):
@dormando thank you for the prompt fix, left some comments to improve my understanding.

proxy_network.c
dormando: `_flush_pending_write` is called in two places; the other place uses the flags directly, this one ignored it and re-sets them, all callers should match logic to reduce potential errors.

Not sure changing the EV_TIMEOUT bit is anything but a style change.

proxy_network.c
dormando: The whole loop needs to be redone; when there are separate lists for to-read and to-write none of that logic will remain. This PR is attempting a minimal churn in the meantime.

proxy_network.c
dormando: Any write syscall can return fewer bytes sent than you attempted to write. I call those short writes.

proxy_network.c
dormando: Among other things yes. This had different symptoms depending on what you were doing. These bugs all worked together to cause problems.

dormando: What I will likely do as soon as I have better test tools:
- split to-write/to-read IO queues
- split to-read and to-write event handlers (maybe). I have them merged because libevent is inefficient but I don't like the bug that this caused.

PR #994, opened 2023-02-15
proxy: fix partial responses on backend timeouts

dormando:
Response object error conditions were not being checked before looking at the response buffer. If a response was partially filled then the backend timed out, a partial response could be sent intead of the proper backend error.

@feihu-stripe - not what we were looking for but a bug I hadn't followed up on from a few weeks ago.


feihu-stripe (approved):
LGTM! Thank you

PR #995, opened 2023-02-15
proxy: add read buffer data to backend errors

dormando:
Errors like "trailing data" or "missingend" or etc are only useful if you're in a debugger and can break and inspect. This adds detail in uriencoding into the log message when applicable.

@feihu-stripe 

TODO
- [x] I wrote a unit test and validated it but it wasn't good enough to add to the commit; need to fix.

Also just very tired right now and don't want to drop this directly into `next` so I can double check it tomorrow.

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -371,6 +371,64 @@ static int _logger_parse_prx_req(logentry *e, char *scratch) {
             );
     return total;
 }
+
+#define MAX_RBUF_READ 100
+static void _logger_log_proxy_errbe(logentry *e, const entry_details *d, const void *entry, va_list ap) {
+    char *errmsg = va_arg(ap, char *);
+    char *be_name = va_arg(ap, char *);
+    char *be_port = va_arg(ap, char *);
+    int be_depth = va_arg(ap, int);
+    char *be_rbuf = va_arg(ap, char *);
+    int be_rbuflen = va_arg(ap, int);
+
+    struct logentry_proxy_errbe *le = (void *)e->data;
+    le->be_depth = be_depth;
+    le->errlen = strlen(errmsg);
+    if (be_name && be_port) {
+        le->be_namelen = strlen(be_name);
+        le->be_portlen = strlen(be_port);
+    }
+
+    le->be_rbuflen = be_rbuflen;
+    if (be_rbuflen > MAX_RBUF_READ) {
+        le->be_rbuflen = MAX_RBUF_READ;
+    }
+
+    char *data = le->data;
+    memcpy(data, errmsg, le->errlen);
+    data += le->errlen;
+    memcpy(data, be_name, le->be_namelen);
+    data += le->be_namelen;
+    memcpy(data, be_port, le->be_portlen);
+    data += le->be_portlen;
+    memcpy(data, be_rbuf, le->be_rbuflen);
+    data += le->be_rbuflen;
+
+    e->size = sizeof(struct logentry_proxy_errbe) + (data - le->data);
+}
+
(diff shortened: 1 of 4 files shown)


logger.c
```diff
@@ -371,6 +371,59 @@ static int _logger_parse_prx_req(logentry *e, char *scratch) {
...
+    struct logentry_proxy_errbe *le = (void *)e->data;
+    le->errlen = strlen(errmsg);
+    if (be_name && be_port) {
+        le->be_namelen = strlen(be_name);
+        le->be_portlen = strlen(be_port);
+    }
+
+    le->be_rbuflen = be_rbuflen;
+    if (be_rbuflen > 100) {
+        le->be_rbuflen = 100;
```
feihu-stripe, line 391: Is this `100` value related to MAX_RBUF_READ?

proxy_network.c
```diff
@@ -1209,6 +1209,8 @@ static int _reset_bad_backend(mcp_backend_t *be, enum proxy_be_failures err) {
     STAILQ_INIT(&be->io_head);
     be->io_next = NULL; // also reset the write offset.
 
+    LOGGER_LOG(NULL, LOG_PROXYEVENTS, LOGGER_PROXY_BE_ERROR, NULL, proxy_be_failure_text[err], be->name, be->port, be->rbuf, be->rbufused);
```
feihu-stripe, line 1212: Is it possible to also log how many requests were failed by this event?

logger.c
```diff
@@ -371,6 +371,59 @@ static int _logger_parse_prx_req(logentry *e, char *scratch) {
...
+    char *errmsg = data;
+    data += le->errlen;
+    char *be_name = data;
+    data += le->be_namelen;
+    char *be_port = data;
+    data += le->be_portlen;
+    char *be_rbuf = data;
+
+    uriencode(be_rbuf, rbuf, le->be_rbuflen, MAX_RBUF_READ);
+    total = snprintf(scratch, LOGGER_PARSE_SCRATCH,
```
feihu-stripe, line 419: Would it be helpful to log some from the end as well?

logger.c
dormando: No I was just putting random limits down. I'll make this more consistent today. 100 is longer than any error message would be, is why I picked that.

dormando: It is semi-related though; it's 300 / 3 (worst case URI encoding)

proxy_network.c
dormando: Maybe..

logger.c
dormando: Not that I can think of, no. It's almost always going to be an error message or some obvious garbage, and in all cases I've seen it's been short data anyway.

proxy_network.c
feihu-stripe: In practice, it would help a lot in correlating client and backend issues. Also provide data on the impact of proxy events.

logger.c
feihu-stripe: Ack

logger.c
feihu-stripe: Thanks, that makes sense. Just from a stand point of getting rid of magic number in the code.

proxy_network.c
dormando: now prints queue depth.

logger.c
dormando: moved the define around

dormando:
fixed a bug, fixed and committed the test. ready to merge.

@feihu-stripe

PR #996, opened 2023-02-16
Update to -f logic

slawomir-pryczek:
- You can now have several -f (factors) that'll be used evenly to calculate slab sizes, there's maximum of 10 factors which is defined as FACTOR_MAX_COUNT.

- Some more logic that'll prevent having incorrect slab sizes instead of throwing an error, eg. if there's 8 byte alignment, having factor of 1 will increase next slab size by 8 bytes

- No longer slabs with different sizes and same perslab will be created


dormando:
Thanks!

I like the fix for different sizes but same perslab. The rest I'm not super sure about right now: I added another feature a long time ago for specifying exact slab sizes directly. So anyone who normally wouldn't care would use the simple factor, but if you really want to hyper-optimize you can do your own factors and directly specify them.

Giving multiple factors seems confusing to users. I'll think about it though and come back when I get a chance.

PR #997, opened 2023-02-17
proxy: fix "missingend" error on reading responses

dormando:
If the backend handler reads an incomplete response from the network, it changes state to wait for more data. The want_read state was considering the data completed if "data read" was bigger than "value length", but it should have been "value + result line".

This means if the response buffer landed in a bullseye where it has read more than the size of the value but less than the total size of the request (typically a span of 200 bytes or less), it would consider the request complete and look for the END\r\n marker.

This bug has been... here forever.

@feihu-stripe

### DIFF
--- a/proxy_network.c
+++ b/proxy_network.c
@@ -1080,7 +1080,7 @@ static int proxy_backend_drive_machine(mcp_backend_t *be) {
             memcpy(r->buf+r->bread, be->rbuf, tocopy);
             r->bread += tocopy;
 
-            if (r->bread >= r->resp.vlen) {
+            if (r->bread >= r->blen) {
                 // all done copying data.
                 if (r->resp.type == MCMC_RESP_GET) {
                     be->state = mcp_backend_read_end;
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -108,6 +108,61 @@ for my $be (@mbe) {
     $mbe[0] = $be;
 }
 
+SKIP: {
+    skip "Remove this skip line to demonstrate pre-patch bug", 1;
+    # Test issue with finding response complete when read lands between value
+    # size and value + response line in size.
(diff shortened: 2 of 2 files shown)


dormando:
I'll attempt to write a test for it tomorrow morning, since I think it should be possible.

I did in fact write unit tests that were very similar scenarios but it didn't trigger the bug, but I think I know why.

dormando:
I feel like this could also cause trailingdata errors but can't prove it yet. I suspect there's still one more bug out here.

The mcshredder test that broke this in two seconds repeatedly has been running for 20+ minutes now.

dormando:
Started straight at this line of code for like two hours across today and yesterday. Man.

feihu-stripe:
Trying to understand different things you said

Example:
```
get key1
VALUE key1 0 4
data
END
```

"value length": that should be 4

"value + result line": what's the result line in case?

"read more than the size of the value but less than the total size of the request": 
size of the value should be 4
what's the total size of the request in this case?

dormando:
`VALUE key1 0 4` is the result line. The total size of the request is `VALUE key1 0 4\r\n` + 'data' + '\r\n' This is properly handled in most of the code (it's calculated into `r->blen`) but that one line wasn't done right.

dormando:
total size of the response, sorry.

dormando:
Test is up. Can't believe how close I was to getting it a few days ago; just had to split the value across an extra sleep call.

There is a bonus test, which will pass if you remove the SKIP parts and run the test against a binary without the patch. Probably nice to leave it in there to demonstrate what the failure mode looks like.

I'm happy to merge this soon, but as customary I'll leave it for another hour or two first so I can revisit it.

dormando:
merged now... on to the next bug, wherever it is.

PR #998, opened 2023-02-19
crawler: don't hold lock while writing to network

dormando:
LRU crawler in per-LRU mode only flushes the write buffer to the client socket while not holding important locks. The hash table iterator version of the crawler was accidentally holding an item lock while flushing to the network. Item locks must NOT be held for long periods of time as they will cause the daemon to lag.

Originally the code used a circular buffer for writing to the network; this allowed it to easily do partial write flushes to the socket and continue filling the other half of the buffer.

Fixing this requires the buffer be resizeable, so we instead use a straight buffer allocation. The write buffer must be large enough to handle all items within a hash table bucket.

Hash table buckets are _supposed_ to max out at an average depth of 1.5 items, so in theory it should never resize. However it's possible to go higher if a user clamps the hash table size. There could also be larger than average buckets naturally due to the hash algorithm and luck.

---
This PR is a two-parter. See the commit message of the second commit, and comments below.

PR TODO's:
- [x] Allow partial buffer flushing (not really sure this is necessary; the buffer is small enough that it shouldn't block much)
- [x] Add minimum buffering during hash crawl mode to cut down on syscalls.
- [x] It passes tests, but anything we can add? Some more manual tests would be smart either way.

### DIFF
--- a/crawler.c
+++ b/crawler.c
@@ -27,8 +27,9 @@
 typedef struct {
     void *c; /* original connection structure. still with source thread attached. */
     int sfd; /* client fd. */
-    bipbuf_t *buf; /* output buffer */
-    char *cbuf; /* current buffer */
+    int buflen;
+    int bufused;
+    char *buf; /* output buffer */
 } crawler_client_t;
 
 typedef struct _crawler_module_t crawler_module_t;
@@ -86,7 +87,7 @@ crawler_module_reg_t *crawler_mod_regs[3] = {
     &crawler_metadump_mod
 };
 
-static int lru_crawler_client_getbuf(crawler_client_t *c);
+static int lru_crawler_write(crawler_client_t *c);
 crawler_module_t active_crawler_mod;
 enum crawler_run_type active_crawler_type;
 
@@ -107,14 +108,13 @@ static void *storage;
 
 /*** LRU CRAWLER THREAD ***/
 
-#define LRU_CRAWLER_WRITEBUF 8192
+#define LRU_CRAWLER_MINBUFSPACE 8192
 
 static void lru_crawler_close_client(crawler_client_t *c) {
     //fprintf(stderr, "CRAWLER: Closing client\n");
     sidethread_conn_close(c->c);
     c->c = NULL;
-    c->cbuf = NULL;
-    bipbuf_free(c->buf);
+    free(c->buf);
     c->buf = NULL;
 }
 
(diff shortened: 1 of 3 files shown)


dormando:
twiddled the `min_items_per_write` define to test the buffer expansion.. works fine.

dormando:
I've added an `lru_crawler mgdump` command, which as the commit message says above:
```
This adds a command "mgdump", which dumps keys in the format:

"mg key\r\nmg key2\r\n"

if a key is binary encoded, it uses the meta binary encoding scheme of
base64-ing keys and appends a "b" flag:

"mg 44OG44K544OI b\r\n"

when the dump is complete it prints an "EN\r\n"
```

A naiive benchmark of dumping 10 million keys:
```
$ time echo "lru_crawler metadump hash" | nc -w 1 localhost 11211 | wc -l
10000001

real    0m5.551s
user    0m0.268s
sys     0m2.508s

$ time echo "lru_crawler mgdump hash" | nc -w 1 localhost 11211 | wc -l
10000001

real    0m3.776s
user    0m0.260s
sys     0m2.234s
```

subtract 1s from the "real" time as that's waiting for the nc timeout.

This is with `lru_crawler_sleep=0` - even with it set to the default of 100 it completes the dump in less time.

This dump format is geared towards getting the keys back out so they can be fetched as fast as possible, without having to decode/re-encode the key, by leveraging the metaget format. Metaget can fetch client flags, time remaining, and generally all information needed to copy the item to another memcached server.

dormando:
Going to let this sit for a few days to make sure I'm happy with it, then add a quick test and the doc/protocol.txt entries. The patch is super trivial since it just reuses the same internal modular interface the rest of the crawl commands use.

feihu-stripe:
Thanks for the fix.

just trying to understand why mgdump took less time than metadump?

dormando:
The eval function uses 40% less CPU. Printf's are very slow and the mgdump doesn't use one. It's only dealing with the key as well.

This is also setting up for cutting the CPU usage of the dump tool by more than half, which you will see once I take some time to write that up.

dormando:
Very high odds I merge this this week. I tried a bit to make an "msdump" stream which just straight gives you the ms command + value but there's a lot of plumbing work necessary. I still like the separate streams since it gives opportunities to dump faster than a single connection would allow. More flexible overall at what I hope will be a small cost.

dormando:
Tested this a bunch with the new `mcdumpload`. Merged it up.

PR #999, opened 2023-02-21
proxy: redo libevent handling code

dormando:
The event handling code was unoptimized and temporary; it was slated for a rewrite for performance and non-critical bugs alone. However the old code may be causing critical bugs so it's being rewritten now.

Fixes:
- backend disconnects are detected immediately instead of on the next time they are used.
- backend reconnects happen _after_ the retry timeout, not before
- use a persistent read handler and a temporary write handler to avoid constantly calling epoll_ctl syscalls for potential performance boost.
- avoids overwriting the read timeout when many requests are being queued for the same non responsive backend

Commit in progress, still hunting some other related timing bugs that seem to still happen.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -323,7 +323,8 @@ struct mcp_backend_wrap_s {
 // FIXME: inline the mcmc client data.
 // TODO: event_thread -> something? union of owner type?
 struct mcp_backend_s {
-    int depth;
+    int depth; // total number of requests in queue
+    int pending_read; // number of requests written to socket, pending read.
     int failed_count; // number of fails (timeouts) in a row
     proxy_event_thread_t *event_thread; // event thread owning this backend.
     void *client; // mcmc client
@@ -333,7 +334,9 @@ struct mcp_backend_s {
     io_pending_proxy_t *io_next; // next request to write.
     char *rbuf; // statically allocated read buffer.
     size_t rbufused; // currently active bytes in the buffer
-    struct event event; // libevent
+    struct event main_event; // libevent: changes role, mostly for main read events
+    struct event write_event; // libevent: only used when socket wbuf full
+    struct event timeout_event; // libevent: alarm for pending reads
     struct proxy_tunables tunables;
 #ifdef HAVE_LIBURING
     proxy_event_t ur_rd_ev; // liburing.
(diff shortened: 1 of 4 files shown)


dormando:
RE tests: the new test suite covers this very well. During development proxyunits.t was failing on some things I had goofed up. I'm not sure what else I would add at the moment.

proxyconfig.t doesn't pass and will need to change a bit, as the proxy is detecting disconnect events fast and is emitting different log lines.

dormando:
Tests should all pass now. I might add one more for proxyconfig.t to test immediate disconnect detection but I think it's optional since I've had to work around it in the tests.

Going to let this sit for a day, review it tomorrow, and probably merge it immediately tomorrow. The old code wasn't really doing anything properly and this is a lot more clear. Also nothing moved around and no new event handlers were added, just the same code split apart.

dormando:
Doing an overload test with mcshredder using a basic test:
- plain gets
- hits primary zone
- overload traffic (three mcshredder threads attempting to run 3 100k rps threads for 300k rps total)
- localhost unimpeded backend nodes
- 10 connections per backend node (soft emulating ten total nodes)
- using the memcached-debug binary, so a prod binary might get more performance.

pre-patch:
- 230k reads/sec average at 200-300 micros sampled latency
- seems to "burst" to 270k sometimes. my pc has turbo enabled...

post-patch:
- had to add an extra 100k rps thread
- 330-340k rps performance, more steady than pre-patch
- 200-300 micros sampled while overloaded

The results are repeatable by switching between binaries.
---

this is a lot more than I was expecting to get from the patch... but a lot closer to the performance I was expecting to get from the proxy in the first place.

---

did a few extra random experiments:
- dropping the # of connections per pool to 1 gives similar performance between `next` and this branch.
- so patch wins a lot as batching decreases and backend connection count increases, which is where the overhead for excess epoll syscalls comes in, as expected.

dormando:
Just did first code review, added a missing stop_timeout in the cleanup routine.

going to do a set of mcshredder tests against network overload scenarios then do one more code audit and probably merge it this afternoon.

dormando:
This is now merged.

@feihu-stripe - feel free to still comment on this PR if you have questions. worst case I add another patch.

dormando:
this code is so much easier to follow and validate than the old original attempt.. wish I did this months ago.

Issue #885, opened 2022-03-27
A possible locking issue


anson-lo:
Hi, should the lock be released after Line 863 to protect `l->dropped++` from races?
https://github.com/memcached/memcached/blob/894e4c16b5162e25ce649cf8a57dfad250965ab0/logger.c#L858-L865

like this way
```
 pthread_mutex_lock(&l->mutex); 
 /* Request a maximum length of data to write to */ 
 e = (logentry *) bipbuf_request(buf, (sizeof(logentry) + reqlen)); 
 if (e == NULL) { 
     l->dropped++;
     pthread_mutex_unlock(&l->mutex); 
     return LOGGER_RET_NOSPACE; 
 } 
```

dormando:
yup, I think that looks wrong. it's accessed under the lock during aggregation.

Thanks!

dormando:
Fixed in 1.6.19

PR #1000, opened 2023-02-23
log: fix race condition while incrementing log entries dropped

rtadepalli:
This fixes the bug reported in #885 

Looks like the reporter didn't make a patch, and the issue has been open for a while, so I took the liberty to make a PR.

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -913,8 +913,8 @@ enum logger_ret_type logger_log(logger *l, const enum log_entry_type event, cons
     /* Request a maximum length of data to write to */
     e = (logentry *) bipbuf_request(buf, (sizeof(logentry) + reqlen));
     if (e == NULL) {
-        pthread_mutex_unlock(&l->mutex);
         l->dropped++;
+        pthread_mutex_unlock(&l->mutex);
         return LOGGER_RET_NOSPACE;
     }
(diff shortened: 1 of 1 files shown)


rtadepalli:
@dormando please let me know if this PR needs to be against `next`. Thanks.

dormando:
Ah heh. sorry. I am terminally behind on issues. Thanks!

Shouldn't need to rebase it against next since it's trivial.

rtadepalli:
Ha, just saw that this is PR number 1000. That's exciting!

dormando:
Congrats :) I'd offer to send you some ancient merch but I have no idea where it all went.

rtadepalli:
Are there some other tasks that need to be run? Looks like the only check that ran was the build test.

dormando:
That's it. I have a bunch of buildbots but they're on my end of things, the github test is just a quick pass to make sure it builds and runs basic tests. I've got a big queue of small PR's to go through in some order..

rtadepalli:
Got it. I was going to ask if this could be merged, but a queue makes sense. Looking forward to getting my first contribution in! Thanks.

dormando:
Out in 1.6.19

PR #1002, opened 2023-03-03
Proxy limiters

dormando:
Adds two limiters:
- mcp.active_req_limit(count) - number of concurrent active requests
- mcp.buffer_memory_limit(kilobytes) - Amount of kilobytes of buffer memory used for request and response values for the proxy.

Not quite as straightforward as I hoped. Needed to make access to mcp.* functions more strict for the config and route routines; you can no longer call config functions from routes and vice versa.

Performance should be good; the counters are per-worker-thread and uses an uncontested mutex to cover. I wanted to go nuts with atomics but that should be a distinct change.

Caveat:
- There is a "hole" for request accounting where the memory can't be accounted for until after it's fully read from the network. More restructuring is required; this is probably fine for the scope of this particular PR as the only way to get into this situation would be for many parallel clients to send a set header, without payload, then hang indefinitely.

TODO:
- [x] unit tests. was going to stop today after writing them but I feel like shit :(
- [x] find a better way to do the read buffer accounting. The IO thread makes this the most difficult.
- [x] audit to see if `AWAIT_BACKGROUND` can escape the limiters (I'm pretty sure it can, but not sure if I will deal with that in this PR)
- [x] Set conn_swallow and sbytes if OOM or active req limit on a set.
- [x] Fix active_req_limit stats output when disabled

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -18,6 +18,18 @@ static void mcp_queue_io(conn *c, mc_resp *resp, int coro_ref, lua_State *Lc);
 /******** EXTERNAL FUNCTIONS ******/
 // functions starting with _ are breakouts for the public functions.
 
+bool proxy_bufmem_checkadd(LIBEVENT_THREAD *t, int len) {
+    bool oom = false;
+    pthread_mutex_lock(&t->proxy_limit_lock);
+    if (t->proxy_buffer_memory_used > t->proxy_buffer_memory_limit) {
+        oom = true;
+    } else {
+        t->proxy_buffer_memory_used += len;
+    }
+    pthread_mutex_unlock(&t->proxy_limit_lock);
+    return oom;
+}
+
 // see also: process_extstore_stats()
 void proxy_stats(void *arg, ADD_STAT add_stats, conn *c) {
     if (arg == NULL) {
@@ -37,12 +49,17 @@ void proxy_stats(void *arg, ADD_STAT add_stats, conn *c) {
 void process_proxy_stats(void *arg, ADD_STAT add_stats, conn *c) {
     char key_str[STAT_KEY_LEN];
     struct proxy_int_stats istats = {0};
+    uint64_t req_limit = 0;
+    uint64_t buffer_memory_limit = 0;
+    uint64_t buffer_memory_used = 0;
 
     if (!arg) {
         return;
     }
     proxy_ctx_t *ctx = arg;
     STAT_L(ctx);
+    req_limit = ctx->active_req_limit;
+    buffer_memory_limit = ctx->buffer_memory_limit;
 
     // prepare aggregated counters.
     struct proxy_user_stats *us = &ctx->user_stats;
@@ -65,6 +82,9 @@ void process_proxy_stats(void *arg, ADD_STAT add_stats, conn *c) {
             }
         }
         WSTAT_UL(t);
+        pthread_mutex_lock(&t->proxy_limit_lock);
+        buffer_memory_used += t->proxy_buffer_memory_used;
+        pthread_mutex_unlock(&t->proxy_limit_lock);
     }
 
     // return all of the user generated stats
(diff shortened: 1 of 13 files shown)


proto_proxy.c
```diff
@@ -325,6 +338,30 @@ void proxy_return_cb(io_pending_t *pending) {
...
+        // isn't set. I'm pretty sure it's impossible as any IO's are created
+        // from a coroutine?
+        assert(p->coro_ref != 0);
+        proxy_ctx_t *ctx = PROXY_GET_CTX(p->coro);
+        // free our buffer here instead of relying on write_and_free since
+        // we're called at the same time anyway and we can do accounting.
+        if (p->client_resp->buf) {
+            free(p->client_resp->buf);
+        }
+        pthread_mutex_lock(&ctx->limit_lock);
```
dormando, line 360: all this needs to move inside the if

dormando:
Still WIP/broken, but revised quite a bit. Had to do some deeper cleanups to make this work properly:

- Allow fast access to the thread/global context in every function
- Start to ensure functions for config_pools cannot be called from config_routes (need to complete this before merge) as the error is now fatal and checks were haphazard before.
- Per-worker-thread accounting for memory buffers, allowing one lock per worker thread; largely uncontested locks aren't very slow. The lock is only necessary when responses have their buffers allocated by the IO thread anyway; with worker IO it'll be totally uncontested.

The memory accounting is still buggy. Might split the commit and upstream the plumbing and active req limit first.

dormando:
Still needs unit tests, but does work now. Edited the commits into two distinct changes.

Did some mcshredder tests and seems okay so far.

@feihu-stripe - courtesy ping.

proxy_lua.c
```diff
@@ -874,11 +876,62 @@ static int mcplib_backend_read_timeout(lua_State *L) {
     return 0;
 }
 
+// TODO (v3): we should be able to push per-thread tunables during the config
```
dormando, line 879: ah I fixed this. need to remove the comment.

dormando:
Another thing bothering me: When the backend OOM is hit it resets the backend, since there's no way yet to swallow the network response.

I'll probably leave that for this PR, since getting the swallowing to work will be better done with some upcoming work on error handling.

proto_proxy.c
```diff
@@ -325,6 +351,27 @@ void proxy_return_cb(io_pending_t *pending) {
...
+            // TODO: lock once, worst case this hashes/locks twice.
+            if (p->miss) {
+                item_unlink(p->hdr_it);
+            }
+            item_remove(p->hdr_it);
+        }
+    } else {
+        // free our buffer here instead of relying on write_and_free since
+        // we can do memory accounting.
+        if (p->client_resp->buf) {
```
dormando, line 365: Tempted to remove this and rely on the GC instead, just to reduce the amount of code a bit.

The tradeoff would be memory sticking around a little longer, but it's pretty much always cleaned up nearly immediately anyway...

dormando:
- need tests around memory buffer limits
- need to properly swallow SET data on error (this isn't done for existing code malloc failures, even)
- might remove some more code

relatively confident this works. I hate the "extra_space" hacks in the read path but fixing that is out of scope here. There are very few places where buffer memory is used/freed so getting the accounting right isn't hard.

proto_proxy.c
dormando: Actually this is a bug I think. the response buffer might still be used later. Will delete

dormando:
want a couple more tests but unsure how offhand...
- testing swallowing bytes (required to keep protocol in sync) requires sending another SET down the same pipe after the first one got eaten... tried but it didn't work so need to try again later.
- testing that commands continue to work down the same connection after hitting a memory limit is tough. might have to reload the server to raise the limit before trying again.

Also:
- Might have to shut client connections once active req limit is hit, since that's being sent back before parsing the request so swallow can't work. As-is it works fine for non-sets.

dormando:
This is pretty much done now. I might add another test or two before merge, and will squash the commits.

Going to hold it for a few days to do more mcshredder based testing and let others take a look.

dormando:
The active request limit isn't too bad, but the buffer memory limit kind of sucks because it's gc-driven. If you set the buffer limit really low it's in theory possible to "settle" with all memory returned but the gc's weren't sufficiently kicked and now you're locked out despite the memory being technically free.

Thinking of setting a minimum memory limit of like.. 256 megabytes. just to hedge against this for now?

Been over it a few times and there just isn't a way around it with the current memory model. If I hack it to kick the garbage collector if it rejects on the memory limit I'll be sent to Programmer Jail

dormando:
Just did another review... looks perfectly fine outside of one nitpick and a concern about low memory limits.

The proxy is disciplined about where it's allocating and freeing memory (just one for req, one for resp) so if this passes tests and a short benchmark it's impossible to screw up. Anywhere it doesn't work should also leak memory.

proto_proxy.c
```diff
@@ -847,13 +890,24 @@ static void proxy_process_command(conn *c, char *command, size_t cmdlen, bool mu
 
     // TODO (v2): lift this to a post-processor?
     if (rq->pr.vlen != 0) {
-        // relying on temporary malloc's not succumbing as poorly to
-        // fragmentation.
-        c->item = malloc(rq->pr.vlen);
+        bool oom = false;
+        pthread_mutex_lock(&c->thread->proxy_limit_lock);
```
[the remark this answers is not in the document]
dormando: I'm a dumb nerd; each of these usages was different, heh. still cleaner to put into a func.

might func-ify it more with later cleanups but just this much helps the code quality.

PR #1004, opened 2023-03-05
proxy: reduce noise for dead backends

dormando:
only log until a backend is marked bad. was previously ticking the "bad" counter on every retry attempt as well.

turned out to be trivial.

TODO:
- [x] Add a specific error log line for "x backend was marked bad, will retry silently"
- [x] Add a counter for connect retries that can keep ticking with a dead backend [edit: maybe some other time]

trying to not overspec this one change since it's trivial.


dormando:
Not sure I care about the counter enough to do it right now. This is ready for merge.

dormando:
merged to `next`.

PR #1005, opened 2023-03-06
proxy: mcp.internal fixes and tests

dormando:
- Refcount leak on sets
- Move the response elapsed timer back closer to when the response was processed as to not clobber the wrong IO object data
- Adds start of unit tests

Requests will look like they run a tiiiiny bit faster than they do, but I need to get the elapsed time there for a later change.

TODO
- [x] Write a few more tests before merging, maybe?
- [x] Write a test specifically for running out of memory.
- [x] Maybe later: internal OOM isn't being surfaced properly, causing a hang.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -290,13 +290,7 @@ void proxy_return_cb(io_pending_t *pending) {
     if (p->is_await) {
         mcplib_await_return(p);
     } else {
-        struct timeval end;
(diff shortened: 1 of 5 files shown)


dormando:
Need to do some deeper fixes on how error messages surface but I'm find labeling this feature experimental for the next cut so long as it's not leaking memory in obvious ways...

PR #1007, opened 2023-03-09
meta: N flag changes append/prepend. ms s flag.

dormando:
Sending 's' flag to metaset now returns the size of the item stored. Useful if you want to know how large an append/prepended item now is.

If the 'N' flag is supplied while in append/prepend mode, allows autovivifying (with exptime supplied from N) for append/prepend style keys that don't need headers created first.

These flags all pre-existed and their meanings are re-used. So fancy.

TODO:
- [x] Code self-review after day delay
- [ ] doc/protocol.txt updates

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -206,9 +206,9 @@ static int _logger_parse_ise(logentry *e, char *scratch) {
     const char * const status_map[] = {
         "not_stored", "stored", "exists", "not_found", "too_large", "no_memory" };
     const char * const cmd_map[] = {
-        "null", "add", "set", "replace", "append", "prepend", "cas" };
+        "null", "add", "set", "replace", "append", "prepend", "cas", "append", "prepend" };
 
-    if (le->cmd <= 6)
+    if (le->cmd <= 8)
         cmd = cmd_map[le->cmd];
 
     uriencode(le->key, keybuf, le->nkey, KEY_MAX_URI_ENCODED_LENGTH);
--- a/proto_text.c
+++ b/proto_text.c
@@ -51,7 +51,7 @@ typedef struct token_s {
     size_t length;
 } token_t;
 
-static void _finalize_mset(conn *c, enum store_item_type ret) {
+static void _finalize_mset(conn *c, int nbytes, enum store_item_type ret) {
     mc_resp *resp = c->resp;
     item *it = c->item;
     conn_set_state(c, conn_new_cmd);
@@ -105,6 +105,16 @@ static void _finalize_mset(conn *c, enum store_item_type ret) {
                 META_CHAR(p, 'c');
                 p = itoa_u64(c->cas, p);
                 break;
+            case 's':
+                // Get final item size, ie from append/prepend
+                META_CHAR(p, 's');
+                // If the size changed during append/prepend
+                if (nbytes != 0) {
+                    p = itoa_u32(nbytes-2, p);
+                } else {
+                    p = itoa_u32(it->nbytes-2, p);
+                }
+                break;
             default:
                 break;
         }
@@ -128,6 +138,7 @@ void complete_nread_ascii(conn *c) {
     int comm = c->cmd;
     enum store_item_type ret;
     bool is_valid = false;
+    int nbytes = 0;
 
     pthread_mutex_lock(&c->thread->stats.mutex);
     c->thread->stats.slab_stats[ITEM_clsid(it)].set_cmds++;
@@ -170,7 +181,7 @@ void complete_nread_ascii(conn *c) {
     } else {
       uint64_t cas = 0;
       c->thread->cur_sfd = c->sfd; // cuddle sfd for logging.
-      ret = store_item(it, comm, c->thread, &cas, c->set_stale);
+      ret = store_item(it, comm, c->thread, &nbytes, &cas, c->set_stale);
 
 #ifdef ENABLE_DTRACE
       switch (c->cmd) {
@@ -203,7 +214,7 @@ void complete_nread_ascii(conn *c) {
 
       if (c->mset_res) {
           c->cas = cas;
-          _finalize_mset(c, ret);
+          _finalize_mset(c, nbytes, ret);
       } else {
           switch (ret) {
           case STORED:
@@ -1379,6 +1390,7 @@ static void process_mset_command(conn *c, token_t *tokens, const size_t ntokens)
     assert(c != NULL);
     mc_resp *resp = c->resp;
     char *p = resp->wbuf;
+    rel_time_t exptime = 0;
 
     WANT_TOKENS_MIN(ntokens, 3);
 
@@ -1427,6 +1439,7 @@ static void process_mset_command(conn *c, token_t *tokens, const size_t ntokens)
     c->noreply = of.no_reply;
     // Clear cas return value
     c->cas = 0;
+    exptime = of.exptime;
 
     bool has_error = false;
     for (i = KEY_TOKEN+1; i < ntokens-1; i++) {
@@ -1448,6 +1461,9 @@ static void process_mset_command(conn *c, token_t *tokens, const size_t ntokens)
             case 'c':
                 // need to set the cas value post-assignment.
                 META_CHAR(p, 'c');
+            case 's':
+                // get the final size post-fill
+                META_CHAR(p, 's');
                 break;
         }
     }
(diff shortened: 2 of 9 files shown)


dormando:
This ended up touching a lot more code than I hoped for. The changes are mild I'm just bummed out :)

proto_text.c
```diff
@@ -105,6 +105,16 @@ static void _finalize_mset(conn *c, enum store_item_type ret) {
...
                 break;
+            case 's':
+                // Get final item size, ie from append/prepend
+                META_CHAR(p, 's');
+                // If the size changed during append/prepend
+                if (nbytes != 0) {
+                    p = itoa_u32(nbytes-2, p);
+                } else {
+                    p = itoa_u32(it->nbytes-2, p);
+                }
```
hemal-shah, line 116: noob question: can we always just rely on this?

proto_text.c
```diff
@@ -1460,10 +1476,20 @@ static void process_mset_command(conn *c, token_t *tokens, const size_t ntokens)
...
+                exptime = of.autoviv_exptime;
+            } else {
+                comm = NREAD_APPEND;
+            }
             break;
         case 'P': // Prepend.
-            comm = NREAD_PREPEND;
+            if (of.vivify) {
+                comm = NREAD_PREPENDVIV;
+                exptime = of.autoviv_exptime;
```
hemal-shah, line 1489: what takes the precedence? is it `T` or `N` taking precedence?

proto_text.c
hemal-shah: at a high level it seems like we are passing a int pointer to `do_item_store` where we update the final length of the `it` based on `it->nbytes`, can we skip passing that pointer and just use the value?

dormando: No, sorry that code is insane. `it` is being overwritten by `new_it` inside do_store_item, and then it's lost forever. I had to add the nbytes bit to get that data out of the function.

I added the conditional in weird places because I'm trying to avoid having the checks for "did nbytes change from do_store_item?" exist outside of append/prepend and `s` flag being used

proto_text.c
dormando: T is ignored when append/prepend are used. That's unfortunately gross but the semantics don't work right otherwise.

hemal-shah: well - to rephrase the question - I just want to know about: 
1. if item exists, memcached will prefer `T`
2. it item gets vivified, memcached will prefer `N`

don't know all the internals of memcached wrt other commands and want to ensure that this new change uses the same approach.

proto_text.c
dormando: only other reasonable option is to make the passed in item ptr be **item and replace it if it's replaced, but that's going to be even less legible + even more fiddling around refcounts than it's already doing.

proto_text.c
hemal-shah: oh - so continuing to use `T` wouldn't make any difference to the item, and have the first exptime used?

proto_text.c
hemal-shah: I get the first part, but don't follow the later. 

>I added the conditional in weird places because I'm trying to avoid having the checks for "did nbytes change from do_store_item?" exist outside of append/prepend and s flag being used

wouldn't the `it` always contain the latest appended/prepended item after execution of `do_item_store`?

dormando: it is passed into do_store_item, not out.

proto_text.c
dormando: That's how it works with the rest of the commands. N initializes and T updates (or sets if used directly). I haven't done a code review in case I goofed it somehow but I think it's right?

proto_text.c
dormando: in append/prepend do_store_item takes an input item, looks up the original item, then merges them together. the item available afterward was still the input item, so without this code it just returns the size of the append command itself.

hemal-shah: ack - that makes sense!

proto_text.c
hemal-shah: okay - I don't know how for sure and just wanted to ensure it's the same approach

proxy_internal.c
```diff
@@ -1204,6 +1216,16 @@ static void process_mset_cmd(LIBEVENT_THREAD *t, mcp_parser_t *pr, mc_resp *resp
                 META_CHAR(p, 'c');
                 p = itoa_u64(cas, p);
                 break;
+            case 's':
```
hemal-shah, line 1219: if the `ms` is supposed to return say `NS` due to cas conflict or something then will `s` flag still return the old item size?

proto_text.c
hemal-shah: I did a small test from your build and not sure if this is expected:
```log
hemal@qa-mydev--0ad632f415f692eb1.northwest.stripe.io:/pay/src$ telnet localhost 11211
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
ms append-test 5 MA N100 T100
12345
HD
ms append-test 5 MA N100 T100
67890
HD
mg append-test v
VA 10
1234567890
mg append-test v t
VA 10 t56
1234567890
ms append-test 5 MA N100 T100
2
22
HD
mg append-test v t
VA 15 t38
12345678902
22
```

It doesn't seem to take the `T100` value in effect

logger.c
```diff
@@ -206,7 +206,7 @@ static int _logger_parse_ise(logentry *e, char *scratch) {
     const char * const status_map[] = {
         "not_stored", "stored", "exists", "not_found", "too_large", "no_memory" };
     const char * const cmd_map[] = {
-        "null", "add", "set", "replace", "append", "prepend", "cas" };
+        "null", "add", "set", "replace", "append", "prepend", "cas", "append", "prepend" };
 
     if (le->cmd <= 6)
         cmd = cmd_map[le->cmd];
```
hemal-shah, line 212: you would need to change this to length of the `cmd_map`, it's not working currently:
```
echo watch fetchers mutations | nc localhost 11211
OK
ts=1678384796.15470 gid=2 type=item_get key=append-test status=not_found clsid=0 cfd=25 size=0
ts=1678384796.15477 gid=3 type=item_store key=append-test status=stored cmd=na ttl=98 clsid=1 cfd=25 size=5
ts=1678384824.945079 gid=4 type=item_get key=append-test status=found clsid=1 cfd=25 size=5
ts=1678384824.945082 gid=5 type=item_store key=append-test status=stored cmd=na ttl=69 clsid=1 cfd=25 size=10
ts=1678384830.459650 gid=6 type=item_get key=append-test status=found clsid=1 cfd=25 size=10
ts=1678384838.767296 gid=7 type=item_get key=append-test status=found clsid=1 cfd=25 size=10
ts=1678384850.559499 gid=8 type=item_get key=append-test status=found clsid=1 cfd=25 size=10
ts=1678384850.559504 gid=9 type=item_store key=append-test status=stored cmd=na ttl=44 clsid=1 cfd=25 size=15
ts=1678384856.493884 gid=10 type=item_get key=append-test status=found clsid=1 cfd=25 size=15
```

proto_text.c
dormando: It's just specifically weird for append. T is completely ignored for append because it normally can't create a new item. If you add the N flag, it _can_ create a new item, but uses the TTL from the token N requires. For mg and ma cases N and T both initialize or update. In a perfect world I would have set use the same semantics, but I didn't and we have to live with it :)

If set used N and T it would be trickier to use anyway; you'd often have to set both flags, or we'd need an exception for set to say T works like N if missing and you're in set mode and etc. Then it's just all complicated.

We might get an extra win if append with T can update the TTL but that's out of scope for this change.

logger.c
hemal-shah: the `cmd=na` should be `cmd=append`

proto_text.c
hemal-shah: okay - that's what you meant by previous comment - its now more clearer. I think in this case we always would have to end up using the `N` flag as the final value instead of keep on extending the TTL

hemal-shah: or we can just do `mg <key> T<extendTTL> time I guess

proxy_internal.c
hemal-shah: okay - then it would be just safe to ignore! thanks!

logger.c
dormando: Whoops, fixed thanks.

proto_text.c
dormando: You've confused me enough that I'm leaving a note here for me to double check the T semantics for append and such. :)

hemal-shah (approved):

hemal-shah:
Thanks for the quick change?

dormando:
you're welcome? lol.

I still need to do a self-review and add the docs later today, but I might just merge it right away after that.

hemal-shah:
> you're welcome? lol.
> 
> I still need to do a self-review and add the docs later today, but I might just merge it right away after that.

😂 
sorry - I meant: thanks for the quick change**!**

proto_text.c
dormando: Double checked: append code explicitly reuses the expiration time of the original item when appending. Making T update TTL on append sounds nice but can be its own change; for now it does nothing.

PR #1008, opened 2023-03-09
check for sys/auxv.h

ffontaine:
Check for `sys/auxv.h` to avoid the following uclibc build failure on aarch64:

```
crc32c.c:277:10: fatal error: sys/auxv.h: No such file or directory
  277 | #include <sys/auxv.h>
      |          ^~~~~~~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/08591fbf9677ff126492c50c15170c641bcab56a

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -556,6 +556,7 @@ AH_BOTTOM([#ifdef HAVE_INTTYPES_H
 #include <inttypes.h>
 #endif
 ])
+AC_CHECK_HEADERS([sys/auxv.h])
 
 dnl **********************************************************************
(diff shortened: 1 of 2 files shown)


dormando:
Thanks! Weird that didn't come up until now

ffontaine:
Actually, this patch was added on buildroot in June 2022: https://git.buildroot.net/buildroot/tree/package/memcached/0002-check-for-sys-auxv.h.patch. But I waited that the first patch was merged to send this one.

dormando:
Released in 1.6.20

PR #1010, opened 2023-03-10
proxy: overhaul backend error handling

dormando:
Cleans up logic around response handling in general. Allows returning
server-sent error messages upstream for handling.

In general SERVER_ERROR means we can keep the connection to the backend.
The rest of the errors are protocol errors, and while some are perfectly
safe to whitelist, clients should not be causing those sorts of errors
and we should cycle the backend regardless.

TODO:
- [x] Could use a few more tests; ensure pipelined commands don't send responses, and that garbage gives garbage errors.
- [x] Tests around the lua API; ensure :ok() fails for an error and :code() works right. It should function the same as an error, it just doesn't blow away the response data further up the chain.
- [x] Self code review. Feels like I'm missing something but it just took a lot of thinking to clean it all up.

At some point I'll either merge code/resp in mcmc or figure out some better way of representing that data, as they do overlap.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -595,7 +595,7 @@ int proxy_run_coroutine(lua_State *Lc, mc_resp *resp, io_pending_proxy_t *p, con
         if (type == LUA_TUSERDATA) {
             mcp_resp_t *r = luaL_checkudata(Lc, 1, "mcp.response");
             _set_noreply_mode(resp, r);
-            if (r->status != MCMC_OK) {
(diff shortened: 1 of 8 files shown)


dormando:
Using the new staging branch to make this easier to review and merge.

dormando:
I'm sure someone's going to want to take the error and shove it into a log line, but I think that's another change.

dormando:
Wowee look at all these tests!

Gonna stage this booger I think. I still kinda hate the resp type vs code thing (And resp type isn't exposed into lua at all), but `resp:ok()` uses the status return value which covers all errors, so that's still perfectly good.

PR #1011, opened 2023-03-14
proxy: fix reversal of pipelined backend queries

dormando:
If a client sends multiple requests in the same packet, the proxy would reverse the requests before sending them to the backend. They would return to client in the correct order because top level responses are sent in the order they were created.

In practice I guess this is rarely noticed. If a client sends a series of commands where the first one generates a syntax error, all prior commands would still succeed.

It would also trip people up if they test pipelining commands as read-your-write would fail as the write gets ordered after the read.

Did run into this before, but I thought it was just the ascii multiget code reversing keys, which would be harmless as the whole command has to complete regardless of key order.

---
I'm creating a PR for this but it will be short lived, so there can be discussion or references here if desired.

So far as I can tell fixing this shouldn't _break_ anything. I discovered it while testing pipelined errors in a different commit.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -224,6 +224,8 @@ void proxy_thread_init(void *ctx, LIBEVENT_THREAD *thr) {
 }
 
 // ctx_stack is a stack of io_pending_proxy_t's.
+// head of q->s_ctx is the "newest" request so we must push into the head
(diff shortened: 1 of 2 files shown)


dormando:
Tagging @feihu-stripe - very small chance this could explain some weird test scenarios, but only if pipelined to the same backend connection.

PR #1012, opened 2023-03-14
proxy: rip out io_uring code

dormando:
with the event handler rewrite the IO thread scales much better (up to 8-12 worker threads), leaving the io_uring code in the dust.

realistically io_uring won't be able to beat the event code if you're using kernels older than 6.2, which is brand new. Instead of carrying all this code around and having people randomly try it to get more performance, I want to rip it out of the way and add it back in later when it makes sense.

I am using mcshredder as a platform to learn and keep up to date with io_uring, and will port over its usage pattern when it's time.


dormando:
Yes I've left a few support bits in, which is fine since I do intend on adding it back later.  Hopefully by the time it gets added back I can use a single alarm timespec and event callback structure, so I don't have to ifdef half the damn codebase.

PR #1014, opened 2023-03-19
[docker] Add debian, upgrade autoconf Ubuntu dependency

olof-nord:
Hello memcached!

Following my previous docker maintenance work (#988), here is another small docker PR.

This one is two-fold:

1) Add a new Debian dockerfile. (perhaps not needed yet, as maintainers are running Debian already?). I do not know which distributions memcached focuses on, but with this file in place without much effort, the Debian people out there might be happier. 

2) Use latest `automake` dependency to build and test ubuntu and debian images. From git history, I understand that v1.11 was verison locked back in 2018 with the initial push for the Ubuntu dockerfile. From what I can tell, the current automake version is 1.16.x.

### DIFF
--- a/devtools/Dockerfile.debian
+++ b/devtools/Dockerfile.debian
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM debian:latest
 
 ARG CONFIGURE_OPTS="--enable-seccomp"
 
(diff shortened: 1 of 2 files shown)


dormando:
Thanks! Fwiw mc is distro-agnostic, but I develop on ubuntu/debian.

dormando:
Released in 1.6.20

PR #1015, opened 2023-03-24
proxy: easier usage of N sockets per backend

dormando:
mcp.backend({ etc, connections = N })

allow creating N tcp sockets for this described backend.

when a request is matched against a pool to retrieve a backend, a container object is pulled:
- the container object holds all tcp sockets
- a tcp socket is selected from the container just before the request is executed
- otherwise all execution is identical to the previous code.

Changing the connection count does not interfere with key distribution within a pool.

TODO:
- [x] specific tests. the test suite does pass as-is, so single connection counts seem to work fine.
- [x] move the connection selection logic to re-enable work batching
- [x] simplify the selection logic
- [x] remove second bad backend checker in backend_run code.
- [x] limit total number of sockets per backend to 8-12? more is unlikely to be helpful to anyone.
- [x] some benchmark testing to ensure no major regressions.

This also deduplicates a little code. Since most related code doesn't change this should be simple to verify.

@feihu-stripe - probably of interest.

### DIFF
--- a/t/proxyconfig.lua
+++ b/t/proxyconfig.lua
@@ -39,6 +39,13 @@ function mcp_config_pools(old)
             test = mcp.pool({b1, b2, b3}, { iothread = false })
         }
         return pools
+    elseif mode == "connections" then
+        local b1 = mcp.backend({ label = "b1c", host = "127.0.0.1", port = 11511,
+                               connections = 3})
+        local pools = {
+            test = mcp.pool({b1})
+        }
+        return pools
     end
 end
 
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -268,7 +268,6 @@ void proxy_submit_cb(io_queue_t *q) {
             be->depth++;
             if (!be->stacked) {
                 be->stacked = true;
-                be->be_next.stqe_next = NULL; // paranoia
                 STAILQ_INSERT_TAIL(&w_head, be, be_next);
             }
         }
--- a/proxy.h
+++ b/proxy.h
@@ -315,6 +315,7 @@ struct mcp_backend_label_s {
     char port[MAX_PORTLEN+1];
     char label[MAX_LABELLEN+1];
     size_t llen; // cache label length for small speedup in pool creation.
+    int conncount; // number of sockets to make.
     struct proxy_tunables tunables;
 };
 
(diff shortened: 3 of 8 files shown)


dormando:
Debating setting a hard limit of 4 or 8 sockets:
- Setting like 100 sockets is guaranteed to be terrible performance, and I know people do that (ie; setting 72 worker threads on a 4 core system)
- Lets me keep the "lowest depth" algorithm. Looping within a 64 byte cacheline is nearly free.
- This still ends up with idle established sockets, which can be problematic for people with dumb firewalls.

tcp keepalive is _supposed_ to be the answer here and I would love to keep the better algorithm. Will reach out to people for feedback.

dormando:
Code audit looks pretty good actually; very little code actually changed. A few loops were added, some code deduplicated, and a passed argument was changed a bunch with no change to its functions.

TODO's still look relevant. Need to decide on a more final chooser function before writing a lot of tests to avoid excess code churn.

dormando:
spent a few days moving my work environment to a new computer. will wrap this up soon.

dormando:
Alright I think most of the work is done now. Needs some benchmarking so I'll come back to that later.

Still debating on hard limiting the connection count. So that's one more TODO item.

t/proxyconfig.lua
```diff
@@ -90,6 +90,14 @@ function mcp_config_pools(old)
             hold = mcp.pool({b2, b3})
         }
         return pools
+    elseif mode == "connections" then
+        local b1 = mcp.backend({ label = "b1c", host = "127.0.0.1", port = 11511,
+                               connections = 3})
+        print("yup")
```
dormando, line 96: dumbass.

dormando:
rebased. this fixed a bug I ended up wasting two hours tracking down the other day :( leaving PR's up for months sucks.

it's been passing mcshredder test suite so far. a few of the tests run better with multiple connections but it's all the heavily pipelined stuff.

going to hold merging for one more day so I can add bandwidth limited tests to the perf suite and re-run everything. Also still need to add the connection limit.. I'll set it to 8 for now.

dormando:
It's passed a bunch of benchmarks.

Think I'll fix up the staging branch and push this in there, so I can start developing on top of it. I intend to include this in the next release but am being extra careful..

dormando:
squashed and merged to staging. I probably won't let it sit in staging for too long, but this lets me write new code on top of it if I need to.

dormando:
rebased over next... going to run the bench suite against it once more and merge.

dormando:
mmm.. can't tell if test suite is hanging due to rare bug or problem with this commit. takes forever to re-run so I guess this may miss the tarball cutoff.

dormando:
Sigh. Paranoia getting the best of me. I'm going to leave this staging for this release and just develop on top of it until I can get more tests built.

I have a lot of benchmark tests now but none that cause timeout/retries mid-test.

dormando:
Leaving a note while I work on this: Just noticed a weird regression that I just pinned to this commit. One specific section of `t/proxy.t` is running extremely slowly (at timeout speeds?). This is only seen in this one test suite and none of the benchmark tests picked it up so far as I can tell.

Will update once fixed.

dormando:
Fixed. Apparently there's only one test in the whole stupid test suite that bothers testing bad backends (or there are but they don't notice the test is taking too long).

Need to get out of the habit of only testing shit that works.

Going to hold this a bit more to write a test that would fail at least somewhere other than the legacy `t/proxy.t` file. Then squash merge and rebase the staging branch.

dormando:
keeping these commits separate for another day while I try to write some tests for the perf suite to see if they'd catch the bug. None of the existing ones do since they don't deal in dead backends.

dormando:
The load generator's "break" tests now reproduce this bug very obviously (traffic goes to zero, 3s lag measurements).

running the whole new break suite now then I can finally merge this damn thing.

dormando:
This is re-merged into staging. it's survived hours and hours of break tests so far.

I'm going to just keep developing on top of staging until I feel the need to move stuff into `next`

dormando:
merged to next. going to close this..

PR #1016, opened 2023-03-24
proxy: fix bug ignoring -R setting for proxy reqs

dormando:
The client connection state machine loops through a few states when handling pipelined requests.

To start:
conn_waiting -> conn_read -> conn_parse_cmd (execution)

After conn_parse_cmd, we can enter:
conn_nread (read a mutation payload from the network) -> conn_new_cmd or directly: conn_new_cmd

conn_new_cmd checks the limit specified in -R, flushing the pipeline if we exceed that limit. Else it wraps back to conn_parse_cmd

The proxy code was _not_ resetting state to conn_new_cmd after any non-mutation command. If a value was set it would properly run through nread -> conn_new_cmd

This means that clients issuing requests against a proxy server have unlimited pipelines, and the proxy will buffer the entire result set before beginning to return data to the client. Especially if requests are for very large items, this can cause a very high Time To First Byte in the response to the client.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -862,6 +862,8 @@ static void proxy_process_command(conn *c, char *command, size_t cmdlen, bool mu
 
         conn_set_state(c, conn_nread);
         return;
+    } else {
+        conn_set_state(c, conn_new_cmd);
     }
 
     proxy_run_coroutine(Lc, c->resp, NULL, c);


dormando:
I wrote a simple utility to check for TTFB latency with large items, when I discovered this problem, it is included below. The lua config is just a bare `return pool(r)` handler, with the single backend set.

Observed the backend data node was flushing data back to the client every 20 requests, but the TTFB observed by the tool kept increasing with the pipeline count. A quick check showed the proxy never leaving the `conn_parse_cmd` state.

Need to spend another hour auditing to ensure this is the proper fix, but it does cap the TTFB observed by the utility and a quick check along the text handler code shows this is correct.

```
#! /usr/bin/perl
#
use warnings;
use strict;

use IO::Socket::INET;
use IO::Select;
use Time::HiRes qw(time);

my $addr = "127.0.0.1:11211";
my $count = 1;

my $sock = IO::Socket::INET->new(PeerAddr => $addr,
                                 Timeout  => 3);
die "$!\n" unless $sock;

$|++;
my $data = '1';
my $size = 900000;
$data = 'x' x $size;
foreach (1 .. $count) {
    print $sock "set /await/foo$_ 0 0 $size\r\n", $data, "\r\n";
    scalar <$sock>;
    print "Going $_\n" if ($_ % 1000 == 0);
}
print "\nDone setting\n";

while (1) {
    my $s = IO::Select->new();
    $s->add($sock);
    my $start = time();
    my $cmd = "mg /await/foo1 s t v\r\n";
    my $buf = '';
    my $count = 200;
    for (1 .. $count) {
        $buf .= $cmd;
    }
    print $sock $buf;
    my @readable = $s->can_read(1);
    my $elapsed = sprintf("%0.8f", time() - $start);
    print "TTFB: ", $elapsed, "\n";
    for (1 .. $count) {
        my $res = <$sock>;
        my $data = <$sock>;
    }
    sleep 1;
}
````

dormando:
Another stupid observation; not all of the text handler code is resetting the command handler (ie touch, delete, etc). It's just always been that way apparently.

I'm starting to think this limit should be "number of requests or amount of queued bytes, whichever comes first" but am hesitant to add even more branches to this hot path. If I can find something else to optimize I may attempt this.

dormando:
Re-reviewed:

- The text handlers do all obey the limit, but most of them are using `out_string()` to write responses, which resets the cmd handler.
- This change is most definitely correct. This spot is either setting nread or `conn_new_cmd` as it should be.

Issue #1017, opened 2023-03-27
Useless verbose if condition


ajccosta:
https://github.com/memcached/memcached/blob/7d6907e659d60512abe7db2154bae039caddee32/items.c#L1008

This if condition is never expanded, what is its purpose?

dormando:
That's a good one, thanks! Probably snuck in during a refactor.

Raimo33:
Any update? I'm happy to tackle this one

ajccosta:
I have no update. It seems to be benign, it is probably optimized out. This path is executed frequently, though. I think the way to go would be to look at git blame and see why it was left like this.

dormando:
Fixed in `next` - gave it a refactor that reduces the branch count anyway.

PR #1019, opened 2023-03-30
fixed verbose FOUND KEY message

ajccosta:
Fixed issue mentioned in #1017 to what seemed to be the original intent.

### DIFF
--- a/proto_text.c
+++ b/proto_text.c
@@ -583,8 +583,7 @@ static inline void process_get_command(conn *c, token_t *tokens, size_t ntokens,
                 {
                   MEMCACHED_COMMAND_GET(c->sfd, ITEM_key(it), it->nkey,
                                         it->nbytes, ITEM_get_cas(it));
-                  int nbytes = it->nbytes;;
-                  nbytes = it->nbytes;
+                  int nbytes = it->nbytes;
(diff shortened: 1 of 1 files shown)


ajccosta:
Removed useless private variable assignment from  proto_text.c:process_get_command

dormando:
Hey,

I merged your second trivial commit. On looking at the first one again though it's incorrect. There are a few prints below that "- nuked by etc" which are supposed to be printed after the "FOUND KEY" etc stuff.

I'll add some in-line comments.

items.c
```diff
@@ -999,19 +999,8 @@ item *do_item_get(const char *key, const size_t nkey, const uint32_t hv, LIBEVEN
             it = NULL;
         }*/
     }
-    int was_found = 0;
 
-    if (settings.verbose > 2) {
-        int ii;
-        if (it == NULL) {
-            fprintf(stderr, "> NOT FOUND ");
-        } else if (was_found) {
```
dormando, line 1008: I think the actual fix is to change this to a bare `} else {` and not change anything else.

It looks like I'd made the same mistake, but left the `else if (was_found` after moving the code back.

dormando:
went with my own refactor for the fix, thanks!

PR #1027, opened 2023-04-25
support: get key range

lidaohang:
support：
get key offset limit

example：
telnet 127.0.0.1 11211
Trying 127.0.0.1...
Connected to 127.0.0.1.
Escape character is '^]'.
set abc 0 0 3
123
STORED
getrange abc 1 2
VALUE abc 0 2
23
END

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1177,6 +1177,10 @@ bool resp_start(conn *c) {
         c->resp->request_addr = c->request_addr;
         c->resp->request_addr_size = c->request_addr_size;
     }
+    resp->getrange = false;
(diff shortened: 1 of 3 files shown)


dormando:
Hey,

Thanks. I won't be adding commands like this; but it would make sense to add as a flag to the metaget command. There are a few ways that could be done and I don't know what I'd prefer offhand.

lidaohang:
Thank you, we are in urgent need of this feature, and hope to contribute.

lidaohang:
@rtomayko

PR #1029, opened 2023-05-03
Add tests scenarios to proxyunits.t

feihu-stripe:
Test scenarios for:
* quiet flag
* request:ntokens()
* request:token()
* request:has_flag()

Want to gather feedback before writing more tests.

Added `announce()` function to group assertion messages with line number.
```
====Test missing END:95====
ok 11 - watcher enabled
ok 12 - get passthrough
ok 13 - backend failure error
ok 14 - got proxy event log line
ok 15 - mock backend created
ok 16 - received version command

====Test trailingdata:122====
ok 17 - watcher enabled
ok 18 - get passthrough
ok 19 - got value back
ok 20 - got data back
ok 21 - got end string
ok 22 - got proxy event log line
ok 23 - mock backend created
ok 24 - received version command

====Test bugfix for missingend:143====
ok 25 # skip Remove this skip line to demonstrate pre-patch bug
ok 26 - get passthrough
ok 27 - got value back
ok 28 - got data back
ok 29 - got end string
```

```
All tests successful.
Files=1, Tests=268,  2 wallclock secs ( 0.03 usr  0.00 sys +  0.07 cusr  0.03 csys =  0.13 CPU)
Result: PASS
```

### DIFF
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -425,8 +425,13 @@ check_version($ps);
 
     print $be "STORED\r\n";
 
-    # Make a version request to ensure STORED is not responded to client.
-    check_version($ps);
+    # To ensure success, make another req and ensure res isn't STORED
+    $cmd = "touch /b/a 50\r\n";
+    print $ps $cmd;
+    is(scalar <$be>, $cmd, "canary touch received");
+    print $be "TOUCHED\r\n";
+
+    is(scalar <$ps>, "TOUCHED\r\n", "got TOUCHED instread of STORED");
 }
 
 check_version($ps);
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -112,7 +112,7 @@ for my $msrv (@mocksrvs) {
 
     # verify a particular proxy event is received
     like(<$w>, qr/ts=(\S+) gid=\d+ type=proxy_backend error=timeout name=127.0.0.1 port=\d+ depth=1 rbuf=EN/, "got proxy event log line");
-    
+
     # backend is disconnected due to the error, so we have to re-establish it.
     $mbe[0] = accept_backend($mocksrvs[0]);
 }
@@ -338,7 +338,7 @@ SKIP: {
     is(scalar <$be>, $cmd, "mg passthrough");
     print $be "HD\r\n";
     is(scalar <$ps>, "HD\r\n", "got mg response");
-    
+
     # ms
     $cmd = "ms /b/a 2";
     print $ps "$cmd\r\nhi\r\n";
@@ -353,14 +353,14 @@ SKIP: {
     is(scalar <$be>, $cmd, "md passthrough");
     print $be "HD\r\n";
     is(scalar <$ps>, "HD\r\n", "got HD from md");
-    
+
     # ma
     $cmd = "ma /b/a\r\n";
     print $ps $cmd;
     is(scalar <$be>, $cmd, "ma passthrough");
     print $be "HD\r\n";
     is(scalar <$ps>, "HD\r\n", "got HD from ma");
-    
+
     # mn and me are not currently supported by proxy.
 }
 
@@ -445,7 +445,7 @@ check_version($ps);
     print $ps $cmd;
     is(scalar <$be>, "ms /b/a 2  \r\n", "set received with q replaced by a space");
     is(scalar <$be>, "hi\r\n", "set payload received");
-    
+
     print $be "HD\r\n";
 
     check_version($ps);
@@ -529,7 +529,7 @@ check_version($ps);
 
     # TODO: command() integer
     #
-    
+
     # check flag and return HD if found.
     print $ps "mg /hasflag/test c\r\n";
     is(scalar <$ps>, "HD C123\r\n", "HD is received");
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -520,10 +520,14 @@ check_version($ps);
     print $ps "ms /token/fetch 2\r\nhi\r\n";
     is(scalar <$ps>, "HD\r\n", "HD is received");
 
-    # command() integer
+    # TODO: command() integer
     #
+    
     # meta:
     # has_flag("F")
+    print $ps "mg /hasflag/test c\r\n";
+    is(scalar <$ps>, "HD C123\r\n", "HD is received");
+
     # test has_flag() against non-meta command
     # flag_token("F") with no token (bool, nil|token)
     # flag_token("F") with token
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -37,8 +37,11 @@ sub accept_backend {
     ok(defined $be, "mock backend created");
     like(<$be>, qr/version/, "received version command");
     print $be "VERSION 1.0.0-mock\r\n";
-
     return $be;
+}
+
+sub announce {
+    print("\n====" . shift . ":" . shift . "====\n");
 }
 
 # Put a version command down the pipe to ensure the socket is clear.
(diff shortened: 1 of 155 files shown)


t/proxyunits.t
```diff
@@ -49,6 +52,7 @@ sub check_version {
     like(<$ps>, qr/VERSION /, "version received");
 }
 
+announce("Initialization", __LINE__);
```
dormando, line 55: please remove `announce` and use the builtin `note()` if you feel like that information is useful. print outputs are selectively eaten while note is explicit.

t/proxyunits.t
```diff
@@ -116,11 +135,12 @@ for my $msrv (@mocksrvs) {
     is(scalar <$ps>, "ok\r\n", "got data back");
     is(scalar <$ps>, "END\r\n", "got end string");
 
-    like(<$w>, qr/ts=(\S+) gid=\d+ type=proxy_backend error=trailingdata name=127.0.0.1 port=\d+ depth=0 rbuf=garbage/, "got backend error log line");
+    like(<$w>, qr/ts=(\S+) gid=\d+ type=proxy_backend error=trailingdata name=127.0.0.1 port=\d+ depth=0 rbuf=garbage/, "got proxy event log line");
```
dormando, line 138: proxy event is a stream; "backend error" is referring to the `type=proxy_backend` here

t/proxyunits.t
```diff
@@ -411,16 +432,55 @@ check_version($ps);
...
+    # make sure q is replaced by a space, and HD is not returned to client
+    my $be = $mbe[0];
+    my $cmd = "ms /b/a 2 q\r\nhi\r\n";
+    print $ps $cmd;
+    is(scalar <$be>, "ms /b/a 2  \r\n", "set received with q replaced by a space");
+    is(scalar <$be>, "hi\r\n", "set payload received");
+
+    print $be "HD\r\n";
+
+    check_version($ps);
```
dormando, line 451: I've typically only been running check_version inbetween each block. though I guess it isn't terrible to do it more often?

t/proxyunits.t
```diff
@@ -444,14 +504,36 @@ check_version($ps);
...
+    # fetch a token and verify in lua then return HD if matched.
+    print $ps "ms /token/fetch 2\r\nhi\r\n";
+    is(scalar <$ps>, "HD\r\n", "HD is received");
+
+    # TODO: command() integer
     #
-    # meta:
-    # has_flag("F")
+
+    # check flag and return HD if found.
```
dormando, line 533: for completeness; run a couple tests with the checked flag in different positions and +/- a token.

dormando (changes requested):
gave it a quick skim. looks like you picked it up pretty quickly :)

address these notes and maybe get the commit history in check? You can add more and I'll check more thoroughly next time.

fwiw for submitting changes to open source projects it's best practice to have sets of independent commits; easier for review and allows quickly accepting things that're less controversial.

ie; new tests and the `announce` thing would've been two commits. not necessary to do here but for future reference.

t/proxyunits.t
```diff
@@ -411,16 +432,55 @@ check_version($ps);
...
+    my $cmd = "ms /b/a 2 q\r\nhi\r\n";
+    print $ps $cmd;
+    is(scalar <$be>, "ms /b/a 2  \r\n", "set received with q replaced by a space");
+    is(scalar <$be>, "hi\r\n", "set payload received");
+
+    print $be "HD\r\n";
+
+    check_version($ps);
+
+    # make sure q is replaced by a space, and none-HD reponse are returned to client.
```
feihu-stripe, line 453: typo: reponse

t/proxyunits.t
feihu-stripe: check_version($ps) is great, one caveat is it does not check backends, I am thinking to evolve it to check both.

t/proxyunits.t
feihu-stripe: ack

t/proxyunits.t
feihu-stripe: ack

t/proxyunits.t
dormando: It doesn't always make sense to check the backends; and the test will still fail if there's junk in there or the backend is dead. I wouldn't want to add too much re-checking between sub tests as it makes the tests too slow.

feihu-stripe: https://github.com/memcached/memcached/pull/1030

It seems to be a good idea to me to clean up and restore the environment between groups of test scenarios? So folks can have a peace of mind to add more tests without worrying about pollutions from previous ones.

dormando: I'd push the opposite direction actually: you should usually clean up a test deliberately if you know it's being left in a bad state. Otherwise we can paper over real bugs, as I've found while writing these scenarios.

ie; I had a regression where the final test in `proxyconfig.t` should have been failing, but it didn't until I changed my environment. Had I been cleaning up the environment after batches of tests it would've never failed.

Basically I try to write groups of tests such that they roll through clean states. Real user environments don't clean up everything as they work.

feihu-stripe: > you should usually clean up a test deliberately if you know it's being left in a bad state. Otherwise we can paper over real bugs, as I've found while writing these scenarios.

I agree with that, and that's exactly why a bug is found thanks to this change, and it serves as a forcing function to make sure everyone cleanups properly otherwise your new tests will fail.

> Basically I try to write groups of tests such that they roll through clean states. Real user environments don't clean up everything as they work.

Can this be an intermediate state before we get there, or you prefer to have this right now?

dormando: What was the bug that got fixed in this example? You said "a bug" but it wasn't obvious what changed or why.

feihu-stripe: https://github.com/memcached/memcached/pull/1030/files#diff-438bf7a9c2561cba34fed3bd2b39404deb1c3fa9891d4f65715d85c64403f6e6R676

The 3rd backend was accidentally killed due to not sending response back.  It took me a while to debug, and have check_sanity() make sure this got caught.

dormando: Ok, I see. thanks!

To summarize my thoughts:

- checking between broader sections is totally fine. I was already doing that part way anyway.
- Checking between every sub-test is not always the right thing to do; you can throw it in more while developing and should use it liberally when debugging. However I've found real code bugs (not test bugs) because feature A used after feature B didn't work as expected. check_sanity() is fine since it's not re-establishing the backend but much more than that isn't helpful.

If we tear everything down all the time we miss a lot of potential bugs, or have to write far more total tests for coverage.

feihu-stripe: Thanks for the summary. We are on the page! `check_sanity` is meant to iterate on `check_versions` that checks between boarder sections.

dormando: Sounds good. You had a section of code where you were running it like 8 times which is why I brought it up.

With proxyconfig.t I struggled with this same issue; to get clean backends I had to issue a proxy reload in a way where the backend labels change so it internally creates new objects and re-connects. It's not abstracted yet.

t/proxyunits.t
```diff
@@ -444,14 +504,36 @@ check_version($ps);
     print $be "END\r\n";
     is(scalar <$ps>, "END\r\n", "ltrimkey END");
 
-    # token(n) fetch
-    # token(n, "replacement")
-    # token(n, "") removal
-    # ntokens()
-    # command() integer
```
feihu-stripe, line 451: @dormando 

What's is this test scenario?

dormando: assert `r:command() == mcp.CMD_GET` etc, probably?

feihu-stripe:
https://github.com/memcached/memcached/pull/1031
https://github.com/memcached/memcached/pull/1030
https://github.com/memcached/memcached/pull/1032

Broken it into 3 PRs, and close this.

PR #1030, opened 2023-05-04
proxy: add check_sanity() to check proxy and backends at boarder sections.

feihu-stripe:
Migrate check_version() to check_sanity() to make sure both proxy and backends buffers are clean between tests.

Found and fixed a bug where a backend was accidentally killed without reconnection.

### DIFF
--- a/t/proxyunits.lua
+++ b/t/proxyunits.lua
@@ -275,6 +275,11 @@ function mcp_config_routes(zones)
         return good_res
     end
 
+    pfx_touch["sanity"] = function(r)
+        local rtable = mcp.await(r, { zones.z1, zones.z2, zones.z3 })
+        return rtable[3]
+    end
+
     mcp.attach(mcp.CMD_GET, toproute_factory(pfx_get, "get"))
     mcp.attach(mcp.CMD_SET, toproute_factory(pfx_set, "set"))
     mcp.attach(mcp.CMD_TOUCH, toproute_factory(pfx_touch, "touch"))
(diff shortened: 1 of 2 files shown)


dormando:
This might make sense for the proxyunits file, but doesn't make sense generically; in other test files the backends won't be 1:1 like this. Which means the test API will now change based on what test file you're looking at.

I'd also at least prefer to keep the functions separate. Ie; there are times where I want to ensure the client socket is clean without touching any backends.

feihu-stripe:
> 

Added check_versions() back, so we can choose which to use.

feihu-stripe:
@dormando added check_versions() back to make it available for proxy only checks.

Please let me know what else can be done.

dormando:
I think the code looks great; the weird merge in the branch is odd though. Can you configure git for rebase instead?

In general the history format is like:

`proxy: some sentence under 70 chars`
[empty newline]
commit message

and squash down commits when you feel they're ready. I do this myself before merging PR's but it's nice if I don't have to.

feihu-stripe:
> I think the code looks great; the weird merge in the branch is odd though. Can you configure git for rebase instead?
> 
> In general the history format is like:
> 
> `proxy: some sentence under 70 chars` [empty newline] commit message
> 
> and squash down commits when you feel they're ready. I do this myself before merging PR's but it's nice if I don't have to.

updated according, will follow in the future. Thanks for the pointers.

dormando:
Looks sweet, thanks!

dormando:
Released in 1.6.20

PR #1032, opened 2023-05-05
proxy: add tests to proxyunits.t

feihu-stripe:
added a few missing tests.

added `proxy_test` that abstracts common test logic .

### DIFF
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -116,7 +116,7 @@ sub proxy_test {
     }
 
     # verify proxy received response
-    if (@{$ps_recv}) { 
+    if (@{$ps_recv}) {
         foreach my $recv (@{$ps_recv}) {
             is(scalar <$ps>, $recv, "ps returned expected response.");
         }
@@ -729,7 +729,7 @@ check_sanity($ps);
 
         my @be_send = ["VALUE $key 0 2\r\nok\r\nEND\r\n"];
         my %be_send = (0 => @be_send, 1 => @be_send, 2 => @be_send);
-        
+
         my @ps_recv = ("VALUE $key 0 11\r\n", "hit hit hit\r\n", "END\r\n");
         proxy_test($ps_send, \%be_recv, \%be_send, \@ps_recv);
     };
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -84,9 +84,50 @@ sub check_sanity {
     is(scalar <$ps>, "TOUCHED\r\n", "sanity check: TOUCHED response received.");
 }
 
-# Basic test with no backends. Write a command to the client, and check the
-# response.
+# $ps_send : request to proxy
+# $be_recv : ref to a hashmap from be index to an array of received requests for validation.
+# $be_send : ref to a hashmap from be index to an array of responses to proxy.
+# $ps_recv : ref to response returned by proxy
+# backends in $be_recv and $be_send are vistied by looping through the @mbe.
+sub proxy_test {
+    my ($ps_send, $be_recv, $be_send, $ps_recv) = @_;
+
+    # sends request to proxy
+    print $ps $ps_send;
+
+    # verify all backends received request
+    foreach my $idx (keys @mbe) {
+        if (exists $be_recv->{$idx}) {
+            my $be = $mbe[$idx];
+            foreach my $recv (@{$be_recv->{$idx}}) {
+                is(scalar <$be>, $recv, "be " . $idx . " received expected response");
+            }
+        }
+    }
+
+    # backends send responses
+    foreach my $idx (keys @mbe) {
+        if (exists $be_send->{$idx}) {
+            my $be = $mbe[$idx];
+            foreach my $send (@{$be_send->{$idx}}) {
+                print $be $send;
+            }
+        }
+    }
+
+    # verify proxy received response
+    if (@{$ps_recv}) { 
+        foreach my $recv (@{$ps_recv}) {
+            is(scalar <$ps>, $recv, "ps returned expected response.");
+        }
+    } else {
+        # makes sure nothing was received when ps_recv is empty.
+        check_version($ps)
+    }
+}
+
 {
+    # Write a request with bad syntax, and check the response.
     print $ps "set with the wrong number of tokens\n";
     is(scalar <$ps>, "CLIENT_ERROR parsing request\r\n", "got CLIENT_ERROR for bad syntax");
 }
@@ -104,14 +145,27 @@ sub check_sanity {
     note("Test missing END:" . __LINE__);
 
     # Test a fix for passing through partial read data if END ends up missing.
-    print $ps "get /b/a\r\n";
     my $be = $mbe[0];
+    my $w = $p_srv->new_sock;
+    print $w "watch proxyevents\n";
+    is(<$w>, "OK\r\n", "watcher enabled");
 
+    # write a request to proxy.
+    print $ps "get /b/a\r\n";
+
+    # verify request is received by backend.
     is(scalar <$be>, "get /b/a\r\n", "get passthrough");
+
+    # write a response with partial data.
     print $be "VALUE /b/a 0 2\r\nhi\r\nEN";
 
+    # verify the error response from proxy
     is(scalar <$ps>, "SERVER_ERROR backend failure\r\n", "backend failure error");
 
+    # verify a particular proxy event logline is received
+    like(<$w>, qr/ts=(\S+) gid=\d+ type=proxy_backend error=timeout name=127.0.0.1 port=\d+ depth=1 rbuf=EN/, "got backend error log line");
+
+    # backend is disconnected due to the error, so we have to re-establish it.
     $mbe[0] = accept_backend($mocksrvs[0]);
 }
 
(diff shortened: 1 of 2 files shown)


t/proxyunits.t
```diff
@@ -69,9 +69,43 @@ for my $msrv (@mocksrvs) {
     push(@mbe, $be);
 }
 
-# Basic test with no backends. Write a command to the client, and check the
-# response.
+# $ps_send : request to proxy
+# $be_recv_ref : ref to a hashmap from be index to an array of received requests for validation.
+# $be_send_ref : ref to a hashmap from be index to an array of responses to proxy.
+# @ps_recv : response returned by proxy
+sub proxy_test {
```
dormando, line 76: The reason why I don't do this more often is the test will fail at what's now an arbitrary line number, making it hard to find the actual failure point, and we can't make the messages as unique. The binary protocol tests are especially bad about this.

So if I want to abstract stuff it should return a pass/fail that can be checked with `ok(proxy_test(etc), "unique message")`

dormando: there's also `subtest` for this type of pattern (see Test::More docs) but I don't use it often.

feihu-stripe: That's a fair point. To improve, I passed the __LINE__ to `proxy_test`, so we won't lose insight.

feihu-stripe: passing unique messages to `proxy_test` should also be easy to implement, but I don't see too much value at the moment.

dormando: In dynamic languages, the line numbers can often be meaningless. That's why the perl test suite centers around the messages instead of line numbers; so please use that instead.

dormando: Messages are also much more informative to end users than line numbers; if a user runs a test that fails and copy/pastes the information to me a line number means I also need to know the exact version/etc. It's much harder to deal with.

feihu-stripe: > In dynamic languages, the line numbers can often be meaningless. 

Could you give me an example?

> so please use that instead.

Absolutely

dormando: Fwiw the above example of calling proxy_test with `ok()` or using subtest are probably the cleanest; that lets proxy_test have unique messages that make sense internally, and a top level message that correlates with the overall test. 

Example: any code generation or function references aren't going to have line numbers that make sense. We don't typically use them in the test suite (look for `eval`), but centering around line numbers cuts us off from being able to use the language properly in these tests.

Also as I added earlier line numbers only make sense for the exact commit that generated them.

feihu-stripe: Thanks for the explain. Subtest does seem to be the best option to me.

Migrated all to subtest, looks good.

```
# Subtest: request:ntokens()
    ok 1 - ps returned expected response.
    ok 2 - ps returned expected response.
    1..2
ok 112 - request:ntokens()
# Subtest: request:token() replacement
    ok 1 - be 0 received expected response
    ok 2 - be 0 received expected response
    ok 3 - ps returned expected response.
    1..3
```

t/proxyunits.t
```diff
@@ -69,9 +69,44 @@ for my $msrv (@mocksrvs) {
...
-# Basic test with no backends. Write a command to the client, and check the
-# response.
+# $ps_send : request to proxy
+# $be_recv_ref : ref to a hashmap from be index to an array of received requests for validation.
+# $be_send_ref : ref to a hashmap from be index to an array of responses to proxy.
+# $ps_recv_ref : ref to response returned by proxy
+sub proxy_test {
+    my ($ps_send, $be_recv_ref, $be_send_ref, $ps_recv_ref) = @_;
+
+    my %be_recv = %{$be_recv_ref};
```
dormando, line 79: skip these. just keep the reference and deref it properly below.

ie `keys %{$be_recv}`
`for my $recv (@{$be_recv->{$idx}})` (arrow deref)
etc.

t/proxyunits.t
```diff
@@ -411,10 +459,44 @@ check_version($ps);
...
+{
+    subtest 'quiet flag 1' => sub {
+        # Test Scenario
+        my $ps_send = "ms /b/a 2 q\r\nhi\r\n";
+        # backend must receive response with quiet flag replaced by a space.
+        my %be_recv = (0 => ["ms /b/a 2  \r\n", "hi\r\n"]);
+        my %be_send = (0 => ["HD\r\n"]);
+        # proxy must not return HD response from the backend.
+        my @ps_recv = ();
+        proxy_test($ps_send, \%be_recv, \%be_send, \@ps_recv);
```
dormando, line 473: Doesn't seem like this asserts that there's no response?

This is where check_version comes in handy (or not trying to wrap everything in one function); assert that you didn't see the HD because you did see a VERSION that was sent later.

dormando: `proxy_test` could check for empty response if @ps_recv is empty easy enough though; `if (@{$ps_recv}) { do the check } else { check_version }`

t/proxyunits.t
```diff
@@ -69,9 +69,44 @@ for my $msrv (@mocksrvs) {
     push(@mbe, $be);
 }
 
-# Basic test with no backends. Write a command to the client, and check the
-# response.
+# $ps_send : request to proxy
+# $be_recv_ref : ref to a hashmap from be index to an array of received requests for validation.
+# $be_send_ref : ref to a hashmap from be index to an array of responses to proxy.
+# $ps_recv_ref : ref to response returned by proxy
+sub proxy_test {
```
dormando, line 76: Having looked more closely at the tests I wonder about this a little more...

Passing in a hash for the expected backends means we lose order, and the function can't test order in which backends receive requests or the order in which it sends them back. Which is fine; this convenience function doesn't need to cover everything.

Except if all of the tests that use it are only ever talking to backend 0 you can cut most of the code out and shorten the text in the top level subtests?

Or did you have something else in mind here?

dormando (changes requested):
Left some comments. Sorry I didn't get to look closely enough last week.

t/proxyunits.t
feihu-stripe: Great, still learning perl.

t/proxyunits.t
feihu-stripe: Options:
1. Rely on external `check_sanity()` at section boarder.
2. Add `check_sanity()` at the end of `proxy_test` to always check both proxy and backends. 
3. Add `check_version()` like you suggested
4. Add `check_version` regardless if ps_recv is empty or not.

I am following your suggestion, but if you see any merit in another option, please let me know.

t/proxyunits.t
dormando: Was attempting to structure things so you wouldn't need to learn much perl to write a bunch of tests :) Oh well. Sorry.

t/proxyunits.t
dormando: I think if the point of the test is to verify a non-response, it should be encompassed within the pass-fail of the subtest. I'd prefer to not add `check_sanity()` at the end of _every_ subtest since that will slow down the suite eventually, otherwise I don't really care.

t/proxyunits.t
```diff
@@ -572,20 +719,23 @@ check_sanity($ps);
...
+
+        my @be_recv = [$ps_send];
+        my %be_recv = (0 => @be_recv, 1 => @be_recv, 2 => @be_recv);
+
+        my @be_send = ["VALUE $key 0 2\r\nok\r\nEND\r\n"];
+        my %be_send = (0 => @be_send, 1 => @be_send, 2 => @be_send);
+        
+        my @ps_recv = ("VALUE $key 0 11\r\n", "hit hit hit\r\n", "END\r\n");
+        proxy_test($ps_send, \%be_recv, \%be_send, \@ps_recv);
+    };
```
feihu-stripe, line 735: This is to show how `proxy_test` can be applies to test with multiple backends.

t/proxyunits.t
feihu-stripe: Did three things:
1. Converted a test scenario to show how `proxy_test` can be used with multiple backends.
2. Added a comment about ordering `# backends in $be_recv and $be_send are vistied by looping through the @mbe.`
3. Loop @mbe to at least have a stable and predictable order of visiting.

t/proxyunits.t
feihu-stripe: `check_version` added.

t/proxyunits.t
dormando: must've missed that one.

guess it's fine, just feels really verbose in the tests.

feihu-stripe: the cognitive burden hopefully can become less when it is applied in more places. :)

feihu-stripe: trade-off between flexibility and verbosity, open for suggestions.

dormando: Most of my point is that it can't be applied in a lot more places. Many tests are checking for order, adding delays or garbage, etc. It's necessary for testing to be flexible, and if the invocation is so large that I'm going to be copy/pasting the whole caller for each test it doesn't seem to actually help much.

For (an incomplete) example:

```
subtest 'foo' => sub {
   my $key = "/basic/a";
    basic_test("get $key\r\n",
      "get $key\r\n",
     "VALUE $key 0 2\r\nhi\r\nEND\r\n",
     ["VALUE $key 0 2\r\n", "hi\r\n", "END\r\n"]);
};
```

for "only check backend zero"; way shorter, way fewer symbols to type, etc.

then add a different function for checking all backends. don't use a function at all (or make another) if order has to be checked, etc.

For named arguments:
```
subtest 'foo' => sub {
  my $key = "/basic/a";
  basic_test(ps_send => "get $key\r\n",
    be_recv => "get $key\r\n",
    etc
  );
};

sub basic_test {
  my %a = @_;
  $a{'ps_send'};
}
```

Still much shorter than the original, a little more typing, but the arguments are named for readability.

dormando: note syntax may not be exact but should be close :P I am groggy and don't write tons of perl anymore.

feihu-stripe: It is not bad with `proxy_test` when testing with backend 0. 

In my mind, `proxy_test` is just for connivence and one can choose when and how to use it.

```
    subtest 'quiet flag 2' => sub {
        proxy_test(
            "ms /b/a 2 q\r\nhi\r\n", 
            {0 => ["ms /b/a 2  \r\n", "hi\r\n"]}, 
            {0 => ["EX\r\n"]}, 
            ["EX\r\n"]
        );
    };
```

feihu-stripe: The same test looks like this in the PR, I personally prefer some verbosity for clarify (but controversial if it really adds clarify or not). 

```
    subtest 'quiet flag 2' => sub {
        my $ps_send = "ms /b/a 2 q\r\nhi\r\n";
        # backend must receive response with quiet flag replaced by a space.
        my %be_recv = (0 => ["ms /b/a 2  \r\n", "hi\r\n"]);
        my %be_send = (0 => ["EX\r\n"]);
        # proxy must return EX response from the backend.
        my @ps_recv = ("EX\r\n");
        proxy_test($ps_send, \%be_recv, \%be_send, \@ps_recv);
    };
```

feihu-stripe: let me looking into `named arguments` and update accordingly.

I need some time to get all the gists of what you said, but I will own up what I submit and deal with future changes.

feihu-stripe: Updated to named argument

feihu-stripe: @dormando can this be merged?

dormando:
Looks great! I gave it a squash and merged into `next`. Looking forward to the next batch :)

dormando:
Released in 1.6.20

PR #1033, opened 2023-05-08
extstore: increase aggressiveness of flush thread

dormando:
Thread wasn't keeping up in high load scenarios with low/default free ratios.

- Increase flush aggressiveness by how far below the "minimum free memory ratio" we are.
- Reset backoff counters when thread goes to sleep.
- Fix where code wasn't setting `do_sleep` to false ever, thus always sleeping the minimum time
- Reduce minimum sleep time from half a millisecond to 200us.

Seems like most of the issue was the backoff counters being reset _after_ the check. Now I more aggressively reset them and that seems to help a lot in local testing.

### DIFF
--- a/storage.c
+++ b/storage.c
@@ -574,8 +574,7 @@ static int storage_write(void *storage, const int clsid, const int item_age) {
 
 static pthread_t storage_write_tid;
 static pthread_mutex_t storage_write_plock;
-#define WRITE_SLEEP_MIN 500
-#define MIN_PAGES_FREE 3
+#define WRITE_SLEEP_MIN 200
 
 static void *storage_write_thread(void *arg) {
     void *storage = arg;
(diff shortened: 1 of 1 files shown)


dormando:
Kind of a dumb thing in retrospect; in testing/benchmarks there's usually a constant load, but production load can wait a long time between sets of flushes. The backoff code meant the thread could wake up a large number of times before even checking any of the classes.

Unless it was asleep for a _very_ long time, where it would then have reduced the backoff over time. With that edit it's much more responsive.

dormando:
I'll hold this for a few days in case of feedback (I think someone might actually test it :P) but if I'm ready for a release I'll merge and include it. It doesn't seem like a regression even if it isn't perfect still.

Just trying to get the thread to survive until the upcoming refactor work.

dormando:
merged since it's had a few days of significant abuse.

dormando:
Released in 1.6.20

PR #1034, opened 2023-05-09
proxy: add test scenarios around request APIs to proxyunits.t

feihu-stripe:


t/proxyunits.lua
```diff
@@ -130,6 +130,85 @@ function mcp_config_routes(zones)
...
+
+    -- Input flags: N10 k c R10
+    -- Output flags: N100 k R100
+    pfx_mg["flagtoken"] = function(r)
+        -- flag_token on non-existing flags: no effect
+        local Ttoken = r:flag_token("T", "T100")
+        local Otoken = r:flag_token("O", nil)
+        local vtoken = r:flag_token("v", "")
+        if vtoken or Otoken or Ttoken then
+            print("flag_token failed on non-existing flags")
```
dormando, line 144: Prints breaks the test suite; the failure code should be returned upstream and handled in the perl side. I believe I do this elsewhere via returning `ERROR [message]`
`FAIL` isn't standard protocol and should be avoided in case of future strictness.

dormando:
just the nitpick about error propagation, I think?

t/proxyunits.lua
feihu-stripe: That's fair, I will update.

I chose to print since FAIL is not standard and causing hang, so print can at least provide some info on where it hangs.

feihu-stripe: I can't just return ERROR from proxy, since <be> must receive the request. Let me think about it.

feihu-stripe: It is the second case `Or is it because the proxy_test function`

I have updated the test to include the error msg part of be response, so they can be print out in perl.

feihu-stripe:
```
# Subtest: request:flag_token()
    not ok 1 - be 0 received expected response

    #   Failed test 'be 0 received expected response'
    #   at t/proxyunits.t line 108.
    #          got: 'mg 'failed at non-existing flags'
    # '
    #     expected: 'mg /flagtoken/a N100 k R100
    # '
    ok 2 - ps returned expected response.
    1..2
    # Looks like you failed 1 test of 2.
```

feihu-stripe: I am thinking to make it consistent that all proxy request should be forwarded to backend, to avoid such a situation.

For example, the following test may verify the following leveraging the P flag.
`be_recv => "ms /token/fetch 2 C123 P/token/fetch\r\nhi\r\n",`

```
    subtest 'request:token() fetch' => sub {
        # ps_recv must received HD for a successful fetch call.
        proxy_test(
            ps_send => "ms /token/fetch 2 C123\r\nhi\r\n",
            ps_recv => ["HD\r\n"],
        );
    };
```

dormando: This is looking like the concern I had about shoehorning every type of test into the one function.

So far as I can tell this test doesn't need to route to the backend at all, and adding all that extra code (the request, to backend, back to client, etc) is just adding a lot of layers that can get in the way of the test. It's also still using non-standard protocol when we have ERROR / SERVER_ERROR / CLIENT_ERROR for exactly this.

I also have done this exact scenario in other tests; just call the command and check for is without proxy_test, since you don't need a backend involved?

feihu-stripe: Added a new commit to return msg to proxy directly and avoid hanging.

dormando: https://github.com/memcached/memcached/blob/master/t/proxyconfig.t#L114

if a test can return directly _or_ write to backend the test you need to test for socket availability. I try to have them be one or the other; either it always returns without talking to a backend or it always talks to a backend. Most of the time it doesn't need to do both, and if it does we can check for socket availability.

dormando: to phrase another way, the test can be strict to "I expect only one of these sockets to wake up" (count the array) "the first socket to wake up should be the backend" (`$ps == etc`)

feihu-stripe: Yes, 💯 `either it always returns without talking to a backend or it always talks to a backend.`

feihu-stripe: Any more change needed for this PR. @dormando ?

Sooner or later I will have to use `select`.

dormando:
did you accidentally close this?

feihu-stripe:
@dormando I was so dumb in git and accidentally deleted everything. Luckily the vscode recovered the files.

dormando:
ah. `git reflog` fwiw. it's pretty hard to actually lose things in git.

dormando:
Released in 1.6.20

PR #1035, opened 2023-05-10
proxy: add response API tests in proxyunits.t

feihu-stripe:


vendor/mcmc/mcmc.c
```diff
@@ -251,6 +252,8 @@ static int _mcmc_parse_response(mcmc_ctx_t *ctx, mcmc_resp_t *r) {
                             } else {
                                 r->vlen_read = buffer_remain;
                             }
+                            // remove vsize and any leading space from response length.
+                            l -= (n - cur) / sizeof(char);
```
dormando, line 256: sizeof(char) is defined as 1 in the C spec. can skip the division.

dormando:
So, mcmc lives here: https://github.com/dormando/mcmc - with its own unit tests. I can port this over and add the tests but if you want to try go ahead.

when I make changes to mcmc I do them in order:
- change upstream
- individual commit updating vendor/mcmc
- individual commit utilizing the updated mcmc

it lets me see historically when vendor stuff changes specifically.

feihu-stripe:
Sure I can give it a try.

dormando:
might be nice in the meantime if you separate out the working tests so I can upstream those. you can leave the broken test in with a skip marker and repair it afterward.

feihu-stripe:
will do

feihu-stripe:
mcmc fix moved to https://github.com/dormando/mcmc/pull/3

PR should now be mergable.

t/proxyunits.t
```diff
@@ -680,15 +680,164 @@ check_sanity($ps);
...
+            be_send => {0 => ["HD\r\n"]},
+            ps_recv => ["HD\r\n"],
+        );
+    };
+
+
+    subtest 'response:hit()' => sub {
+        # ps_recv must not receive an error
+        proxy_test(
+            ps_send => "mg /response/ok\r\n",
```
dormando, line 712: double testing ok() instead of hit here.

should also test cases where ok/hit are false work properly.

dormando:
I think the rest looks good but kinda groggy and need to think more.

for the `hit()` and `ok()` tests there are a few backend conditions which can result in `ok()`. ideally we test more than one of them.

feihu-stripe:
Added more cases for ok() and hit()

dormando:
looks okay I think. I'm going to need to come back after a break for a final review though; if you think you're done feel free to squash it.

feihu-stripe:
> looks okay I think. I'm going to need to come back after a break for a final review though; if you think you're done feel free to squash it.

Did what I can think of, but I guess it is never 100%. 

Squahsed.

dormando:
Thanks. I think it's more important to round this out by ensuring we have `mcp.pool` two-arg tests now. We should be periodically revisiting the tests and adding layers to it instead of trying to waterfall full scenarios anyway.

feihu-stripe:
>  we have mcp.pool two-arg tests now

sure will take a look

dormando:
there're a bunch of examples commented in t/startfile.lua fwiw.

dormando:
merged it. didn't feel like the comment was a blocker.

dormando:
Released in 1.6.21

PR #1036, opened 2023-05-10
proxy: add more tests for await() in proxyunits.t

feihu-stripe:


t/proxyunits.t
```diff
@@ -120,14 +126,18 @@ sub proxy_test {
...
+    if (defined $ps_recv) {
+        if (scalar @{$ps_recv}) {
+            foreach my $recv (@{$ps_recv}) {
+                is(scalar <$ps>, $recv, "ps returned expected response.");
+            }
+            # makes sure nothing remains in $ps.
+            check_version($ps);
+        } else {
+            # when $ps_recv is empty, make sure it is not readable.
+            my @readable = $pss->can_read(0.1);
```
dormando, line 138: not 100% sure I follow what's going on here. can it simply do the check_version either way? Otherwise it's going to add a 0.1s delay every time a test like this is run.

t/proxyunits.t
```diff
@@ -771,46 +784,138 @@ check_sanity($ps);
...
+            be_send => {0 => [$be_hit]},
+            ps_recv => ["VALUE $key 0 1\r\n", "1\r\n", "END\r\n"],
+        );
+        # send response to not timeout
+        $mbe[1]->send($be_miss);
+        $mbe[2]->send($be_miss);
+    };
 
-    # await(r, p(3+), 2)
-    $key = "/awaitone/b";
```
dormando, line 790: It's hard to see because the tests are so verbose now, but I don't think I see this one replaced below?

this was testing 3 pools wait for 2 responses?

feihu-stripe: look for

`subtest 'await(r, p, 2): sent [h, h, h], recv 2 responses' => sub {`

feihu-stripe: I am hoping the title of the subtest make some sense in determining the test behaviors.

t/proxyunits.t
feihu-stripe: I am debating on this too, the behavior I am trying to support if

First `proxy_test`, some backends responded, and expect $ps should receive none.
Second `proxy_test`, other backends responded, $ps received response.

More generically, it allows inserting actions among multiple `proxy_test` calls for complex scenarios.

> Otherwise it's going to add a 0.1s delay every time a test like this is run.

Yes, but unless we explicitly do `ps_recv => [],` which only happens in a single scenario in the quiet flag test.

t/proxyunits.t
dormando: the title is a goal, not an execution plan.

feihu-stripe: Happy to change to good alternatives.

t/proxyunits.t
dormando: I'm not complaining I'm just not (yet?) seeing how it's functionally different.

either you check if the socket is readable (which only happens if something was sent)
or you put something in the after the test runs and the response (a version check). if you get that response, there was nothing else to be read.

they should be functionally equivalent (if the socket became readable that's because there was something in there), but in the second case it can execute immediately instead of after 100ms.

unless I'm still misunderstanding?

t/proxyunits.t
dormando: I'm not really sure how to explain it yet, I'm sorry.

In the old test format I could see a comment (the plan/goal) and then the steps linearly.

When I'm reviewing the tests now, the steps are:
- look at the goal
- look at the argument list, determine if it matches the goal
- look at the function and execute it in my head to see if the execution matches the plan

I'm just stupid/slow. sorry.

feihu-stripe: Your steps makes sense, and I was trying to serve those steps in my way:

Let me explain :) 
`subtest 'await(r, p, 2): sent [h, h, h], recv 2 responses' => sub {` in which:

`recv 2 responses` is the goal
`sent [h, h, h],` is the argument list or input
`await(r, p, 2)` is the function

It can be in any other form/format as long as it serves the purpose, sometimes it is hard for me to get it right without knowing exactly what you are suggesting (not complaint at all, communication is just hard.).

feihu-stripe: If you would like me to convert them into comment, I am happy to do so.

t/proxyunits.t
feihu-stripe:
```
    if (defined $ps_recv) {
        if (scalar @{$ps_recv}) {
            foreach my $recv (@{$ps_recv}) {
                is(scalar <$ps>, $recv, "ps returned expected response.");
            }
            # makes sure nothing remains in $ps.
            check_version($ps);
        } else {
            # when $ps_recv is empty, make sure it is not readable.
            my @readable = $pss->can_read(0.1);
            is(scalar @readable, 0, "ps is expected to be non-readable");
+++     check_version($ps); 
        }
    }
```


This test will fail with added check version
```
    subtest 'await(r, p, 1): send [e, m, h], recv 3 responses and 1 reconn' => sub {
        my $key = "/awaitone/a";
        my $ps_send = "get $key\r\n";
        my $be_hit = "VALUE $key 0 2\r\nok\r\nEND\r\n";
        my $be_miss = "END\r\n";
        my $be_error = "ERROR backend failure\r\n";
        proxy_test(
            ps_send => $ps_send,
            be_recv => {0 => [$ps_send], 1 => [$ps_send], 2 => [$ps_send]},
            be_send => {0 => [$be_error], 1 => [$be_miss]},
            ps_recv => [],
        );
        proxy_test(
            be_send => {2 => [$be_hit]},
            ps_recv => ["VALUE $key 0 1\r\n", "3\r\n", "END\r\n"],
        );
        # reconnect due to ERROR
        $mbe[0] = accept_backend($mocksrvs[0]);
    };
```

failure is 
```
    #   Failed test 'version received'
    #   at t/proxyunits.t line 74.
    #                   'VALUE /awaitone/a 0 1
    # '
    #     doesn't match '(?^:VERSION )'

    #   Failed test 'ps returned expected response.'
    #   at t/proxyunits.t line 132.
    #          got: '3
    # '
    #     expected: 'VALUE /awaitone/a 0 1
    # '
```

I assume this is because `VERSION` unblocked await(), but please correct me.

t/proxyunits.t
dormando: I think my point is that as a reviewer I need to make sure the test code matches the comment. I can read the note just fine, moving it to a comment doesn't change the functionality of the sentence :)

I also need to make sure the test is exercising the memcached code in a way that is useful, so there's possibly more going on in my head than an outsider.

t/proxyunits.t
feihu-stripe: works without added check_version()

t/proxyunits.t
feihu-stripe: > moving it to a comment doesn't change the functionality of the sentence :)

That's what I thought :)

feihu-stripe: Understood, I am doing what I can to make that easy for you as the reviewer.

t/proxyunits.t
dormando: okay so I guess when you send the version it's going to block until there's a timeout, then you'll get some response from the await as it will complete. which is weird since it looks like you're not getting an error or a miss, you're getting a VALUE. So maybe something's actually broken somewhere? or elsewhere in the test structure?

t/proxyunits.t
dormando: Actually I think you misread my original breakdown:

- look at the goal: `await(r, p, 2): sent [h, h, h], recv 2 responses`
- look at the argument list: `my $blah; proxy_test(etc, etc, etc, etc)`
- look at the function and execute it in my head: `sub proxy_test { arg arg loop test loop test }`

previously this was all linear in place, and if I wanted to adjust or add debugging to any step it was also in-place and independent of any other test being executed.

feihu-stripe: `sub proxy_test { arg arg loop test loop test }` is generic, and soon you wouldn't need to look into `proxy_test`

t/proxyunits.t
feihu-stripe: await() is unblocked as confirmed by print from Lua after await()

dormando: yeah; but your example is showing a `VALUE` response during the `check_version` call, but you're not sending that from the backend earlier. So where is that coming from? or is the example copy/pasted wrong?

dormando: To be clear I'm saying "I smell a bug", unless the example you copy/pasted was wrong.

feihu-stripe: yeah, I am seeing an unexpected timeout event, which marked the third response as error, but I am not yet clear why timeout can happen.

dormando: I'm not talking about the timeout; I'm talking about the string you're getting back that you didn't send. The timeout makes sense.

dormando: The timeout will happen when you send a version command because the test will block until it gets _any_ response, so the backend waiting for a response to the await will eventually time out and error, letting the await complete, then send what it had upstream.

but in your example you send an error and a miss, but get a `VALUE` back during the version test. that should be a bug somewhere?

feihu-stripe: > I'm not talking about the timeout; I'm talking about the string you're getting back that you didn't send. 

That is hard coded in Lua, when await() is unblocked. So timeout unblocked the await() and return that string

feihu-stripe: Sorry you might be right, I need some more time to sort this out.

dormando: Derp. I'm an idiot. But now that I'm looking at the right place it's more confusing. but I think I get it:

- awaitone/a is waiting for "1" response. The default mode is GOOD (looking for hits, ideally)
- in your test you send an error and a miss, then it dies and times out because of the weird modification I asked for because I'm dumb.
- await returns all of its responses so far, which ends up being 1 (the miss, I'm guessing).

So either A) That error should've also been turned into a response? I think not but can't remember why it doesn't offhand. IIRC backend failures are harder errors than an error message type.
or B) awaitone should be counting the number of `ok()` or good or hits instead of just the total number. Since it looks like we can accidentally write buggy tests with it without being more clear.

feihu-stripe: OK, I think I got what you meant.

So if we sent await() and version to proxy, the return sequence is await() and version, meaning version response is always after await().

then timeout makes sense, like you explained.

And to your question, again the VALUE string is hard-coded in Lua, that always returns after await(), make sense?

dormando: Circling back to the original problem; I think once we're doing magical state across multiple calls to `proxy_test` via special arguments that change behavior it no longer meets my threshold for simplicity. If that's okay?

The readable wait you added can stay though. now that I actually understand why it's there.

feihu-stripe: OK I will refrain from using multiple `proxy_test` for complex tests.

Sure we can verify the count of ok and none oks

t/proxyunits.lua
```diff
@@ -298,13 +307,22 @@ function mcp_config_routes(zones)
...
+        local ok_cnt = 0
+        local hit_cnt = 0
         local count = 0
         for i, res in pairs(rtable) do
+            if res:ok() then
+                ok_cnt = ok_cnt + 1
+            end
+            if res:hit() then
+                hit_cnt = hit_cnt + 1
+            end
```
feihu-stripe, line 319: can refactor boilerplates later when more tests are added for await()

dormando:
Looks way better now, thanks! Latest changes just levelled up the test accuracy I think.

dormando:
Do you feel like this is final from your end now? I need a break before I can properly review it but a skim looked okay.

feihu-stripe:
This is the first batch for proof of concept. Will follow up with more await() tests.

dormando:
Sorry for what ended up being a monthlong delay. Had to ruthlessly prioritize.

This is now merged into `next`.

dormando:
Released in 1.6.21

PR #1037, opened 2023-05-11
proxy: add tests for ustats

feihu-stripe:


t/proxyustats.t
```diff
@@ -0,0 +1,227 @@
...
+# the server so we can pass that in?
+my $configfile = "/tmp/proxyustats.lua";
+
+if (!supports_proxy()) {
+    plan skip_all => 'proxy not enabled';
+    exit 0;
+}
+
+# Set up some server sockets.
+sub mock_server {
```
dormando, line 23: can omit this stuff since it's unused?

dormando:
I think this is good enough? I might've tested changing individual counters but I think if any of them are changeable it's going to be fine.

t/proxyustats.t
feihu-stripe: removed, thanks for the insight.

dormando: for what it's worth I was trying to add some of these functions to the test library, but the test library can't run tests easily. I need to revisit that so we don't have to carry them in every file. It's not like that because I'm a mess it's just low priority.

feihu-stripe: I don't see that as an urgent or blocking issue TBH. As we have more files, the pattern will be clear and then we can take action on refactoring.

dormando:
merged to `next`

dormando:
Released in 1.6.20

PR #1038, opened 2023-05-22
Handle incorrect ext_path units like MB and GB gracefully

dannykopping:
Previously, supplying `ext_path=5GB` (note the `B`) would result in extstore being limited to 0 bytes due to the `multiplier` being initialized to 0 and never set.


dormando:
merged for the next release finally. I also added another commit to fail if 0 bytes assigned and added some unit tests.

sorry for the long delay, I had to aggressively prioritize for a few weeks.

dormando:
Released in 1.6.21

PR #1039, opened 2023-05-23
Bucketstore proof of concept

dormando:
Takes keys in the format of: `/first/part/of/keyNNNN`, takes everything after the _final_ `/` character and turns that into a hashed bucket. Appends data together.

Example configuration:

```
function mcp_config_pools()
    return {}
end

function mcp_config_routes(o)
    mcp.attach(mcp.CMD_MS, function(r)
        local new = r:bucket(20)
        local res = mcp.force_recache(new)
        if res == nil then
            --print("miss, fine to run command")
            return mcp.internal(new)
        elseif res == true then
            --print("successfuly recached or already in memory, fine to run command")
            return mcp.internal(new)
        else
            print("failed to fetch or recache")
            return "NS\r\n"
        end
    end)
    mcp.attach(mcp.CMD_MG, function(r)
        local new = r:bucket(20)
        local resp = mcp.internal(new)
        local value = resp:bucketscan(r)
        if value ~= nil then
            -- value has \r\n already.
            return "VA " .. string.len(value)-2 .. "\r\n" .. value
        else
            return "EN\r\n"
        end
    end)
end
```

TODO:
- [ ] Update example configuration for set retries. (it should be fine to retry 1-3 times if you hit the "failed to fetch" condition)
- [ ] Edge handling and cleanups.
- [ ] Configuration on how to parse the key
- [ ] Copy meta flags around internally.

DO NOT attempt to clean up this code! It is written deliberately to make it easier to rebase in the future.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -9,11 +9,13 @@
  */
 
 #include "proxy.h"
+#include "storage.h"
 
 #define PROCESS_MULTIGET true
 #define PROCESS_NORMAL false
 static void proxy_process_command(conn *c, char *command, size_t cmdlen, bool multiget);
 static void mcp_queue_io(conn *c, mc_resp *resp, int coro_ref, lua_State *Lc);
+static int mcp_storage_recache(conn *c, mc_resp *resp, int coro_ref, lua_State *Lc);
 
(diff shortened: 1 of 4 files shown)


dormando:
gonna throw this away.

PR #1041, opened 2023-05-26
Fix confusing cflags comment in struct item, and replace suffix with cflags.

liangxs:
The comment `flags length\r\n` in struct `item` is outdated and incorrect, which introduces confusing.
There should be comment on client flags.

And the old `suffix` related code should be updated.

### DIFF
--- a/items.c
+++ b/items.c
@@ -153,22 +153,16 @@ unsigned int do_get_lru_size(uint32_t id) {
 /**
  * Generates the variable-sized part of the header for an object.
  *
- * nkey    - The length of the key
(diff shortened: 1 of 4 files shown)


dormando:
Thanks and sorry for the delay (as usual, sigh).

I like the cleanup of various comments and centralizing the size logic. I was purposefully leaving the suffix listed as a suffix in case I wanted to push more data in there based on header bits, but on reading your change this is probably not necessary.

I do need to very carefully review this since it's a core change but I've read it over a few times and it looks right.

PR #1047, opened 2023-06-14
extstore: fix data bugs on high overwrite key

dormando:
Scenario required:

- likely item_age=2 (or some low value), so cache memory is otherwise not used to hold data very long.
- one, or a few keys, are frequently updated and re-written to disk.
- at some point extstore compaction runs

There are two faults to this bug:

1) We may accidentally "rescue" an older version of the frequently
   overwritten key. More likely the data will go missing.
2) More likely, extstore will log too many "deletes" against an active
   disk page. It will then think the page is empty when it isn't and
   invalidate it.
   This results in other random items disappearing when they shouldn't.

The fix is simple in retrospect: when validating an in-memory header against data on disk, we check that the page and page version are the same, but it was not checking that the *offset on disk* is correct. Now it checks all three identifiers which fixes the bug.

Holding this for a bit until I can run more tests against it, but I can't think of any reason why it would be incorrect.

### DIFF
--- a/storage.c
+++ b/storage.c
@@ -788,7 +788,7 @@ struct storage_compact_wrap {
 
 static void storage_compact_readback(void *storage, logger *l,
         bool drop_unread, char *readback_buf,
-        uint32_t page_id, uint64_t page_version, uint64_t read_size) {
+        uint32_t page_id, uint64_t page_version, uint32_t page_offset, uint64_t read_size) {
     uint64_t offset = 0;
     unsigned int rescues = 0;
     unsigned int lost = 0;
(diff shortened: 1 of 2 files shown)


dormando:
uh oh. blew up here but tests are passing locally... race condition?

ah got chunked-extstore to fail. I wonder if the tests were accounting for this breakage already :(

dormando:
still unsure if the test or code is wrong. it's only failing for large items.

Can see it finding headers for keys but the offsets are all wrong. will resume tomorrow..

dormando:
oooooooooooooooooooooooo right this is readback chunk offset, not total page offset. need to add the read buffer's page offset.

dormando:
All clean. Gonna let it sit and merge after I check a few other tasks in case I think of something.

dormando:
Released in 1.6.21

PR #1049, opened 2023-06-15
proxy: fixes for memory tracking

dormando:
- fix ability to disable buffer memory limiter.
- fix memory accounting leak for ascii multigets

In the ascii multiget case it was over-accounting for the "END\r\n" marker, leaking 5 bytes per key read.

---

My test suite is still showing another leak or accounting bug, so keeping this open for additional fixes.


dormando:
Not actually a leak, bench suite is now showing memory freeing up properly.

Should be good to merge.

PR #1050, opened 2023-06-29
Xhash update

devnexen:


dormando:
Thanks!

Changelog? What does this fix specifically? Looks like maybe build fixes under new gcc?

"If it ain't broke" I don't see the need to update.

devnexen:
> Thanks!
> 
> Changelog? What does this fix specifically? Looks like maybe build fixes under new gcc?

There is a bit of these but also [those](https://github.com/Cyan4973/xxHash/releases/tag/v0.8.1) (e.g. XXH3 improvements)

> 
> "If it ain't broke" I don't see the need to update.

sergey-dryabzhinsky:
Is there any improvements gain by updating it to 0.8.3?

PR #1052, opened 2023-06-30
Proxy optimization pass

dormando:
This series is some low hanging fruit on pending performance updates.

Been developing a stability test suite and having something to benchmark/test is useful, plus having performance improvements is always good.

The total improvement is between 3% and 20% depending on the workload. Latencies are typically improved.

TODO:
- [x] In return IO optimization, see if we can remove a potentially redundant next pointer from the IO object. (skipping for now to reduce code churn)
- [x] Check potential tail latency impact of the worker IO batch commit.
- [x] Step through the epoll_ctl changes to double check read pending states are retried
- [x] Just run it through the suite a few more times during development as a bug check.

I picked these as relatively low risk high return optimizations.


dormando:
After a mcshredder bugfix I got a clean test pass and it's a pretty good consistent performance uplift and no crashes or obvious errors so far.

3% to 20% is still about right. Surprised the mcp.internal benchmark didn't really improve so I guess that one's bottlenecked elsewhere.

dormando:
Improvements to test suite made it easier to run subsets of tests. Just ran a comparison with the worker IO patch completely removed and the performance was the same. I've backed up to the commit to test again some other time.

So that reduces code churn even more.

Once I hand verify the epoll_ctl optimization this should be safe to drop into `staging`

dormando:
No longer WIP. Has survived multiple runs of the stress suite including the new fault tests.

Still need to hand-verify the epoll_ctl change before merging into staging.

dormando:
fixed missing state machine text for verbose output.

manually stepped through the state machine and verified that it does what's expected.

ready to merge into staging and move on.

Issue #1051, opened 2023-06-30
Inconsistent item counts with track_sizes


56quarters:
**Describe the bug**

Hello!

When running with the option `--extended=track_sizes`, item counts returned by the command `stats sizes` are sometimes incorrect after letting items in the cache expire via TTL.

This seems like a race condition [updating the sizes histograms](https://github.com/memcached/memcached/blob/efee763c93249358ea5b3b42c7fd4e57e2599c30/items.c#L929-L948) based on the output of `stats sizes` after all items have been expired from the cache:

```
$ printf 'stats sizes\r\n' | nc -N localhost 11211
STAT 3168 4294967295
STAT 3456 4294967295
END
```

Note that both these counts are `2^32-1`, which makes sense if they were `unsigned int`s set to `0` and then decremented again.

According to `stats items` and `lru_crawler`, there aren't actually any items in the cache after letting items expire via TTL.

```
$ printf 'stats items\r\n' | nc -N localhost 11211
END
```

```
$ printf 'lru_crawler metadump all\r\n' | nc -N localhost 11211
END
```

For what it's worth, I'm able to fix the incorrect counts by [adding locking](https://github.com/memcached/memcached/commit/24e9edd0373f24d02a3a35546cbb7eb4b06bbb7e) around updates to the sizes histogram.

**To Reproduce**

_Note that the steps below use a Go tool I've been using for testing, but anything that does lots of concurrent reads and writes should work._

1. After building from commit `efee763c93249358ea5b3b42c7fd4e57e2599c30` (I don't think this specific commit is required, it's just what I've been using), start Memcached with the `track_sizes` option: `./memcached -v -m 1024 -I 64m --extended=track_sizes`.
1. In the directory that contains the Memcached repo, check out [jankcache](https://github.com/56quarters/jankcache).
1. Run the [`reads-and-writes.go`](https://github.com/56quarters/jankcache/blob/9393d02d9b410416fbad7f7cf564e88a104281a0/tools/reads-and-writes.go) tool to `get` and `set` Memcached source code in a local Memcached instance: `go run tools/reads-and-writes.go ../memcached`
1. Let the tool run for about 30 seconds and then stop it via `^C`
1. Wait for the TTL of items in the cache to expire (5 minutes when using `reads-and-writes.go`).
1. Run `stats sizes` and confirm that counts of some sizes are not `0`: `printf 'stats sizes\r\n' | nc -N localhost 11211`

**Expected output**

```
[...]

PR #1053, opened 2023-06-30
Note that counts for `track_sizes` are best effort

56quarters:
Add note to help and protocol docs that counters for the sizes histogram are approximate since increments and decrements are not synchronized.

Related #1051


dormando:
Thanks!

dormando:
released in 1.6.22

PR #1057, opened 2023-07-12
proxy: mcp.backend({ down = true }) option

dormando:
Starts backend in a "bad" state, so any attempt to access it will result in "SERVER_ERROR backend failure".

The only times a backend's bad flag can be removed is on a successful connection attempt. So all we need to do is avoid fully setting up or tearing down the backend and starting it marked bad.

TODO:
- [x] one more code audit pass.
- [x] walk through code in debugger during the cleanup phase to verify.
- [x] maybe one more test with down = false to ensure it does still connect? use the same label to ensure it updates.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -188,6 +188,7 @@ struct proxy_tunables {
 #endif // HAVE_LIBURING
     int backend_failure_limit;
     bool tcp_keepalive;
+    bool down; // backend is forced into a down/bad state.
 };
 
 typedef STAILQ_HEAD(pool_head_s, mcp_pool_s) pool_head_t;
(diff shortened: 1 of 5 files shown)


dormando:
might be trivial enough to merge to staging immediately after the audit and additional test.

dormando:
added the extra test. works fine.

need to do the debugger stepping and closer code audit. then should be clear to merge to staging later today.

dormando:
Code looks sensible on an audit after waiting a day.

Just stepped through the cleanup process in a debugger for both down = true and down = false after toggling and the objects look the way I expect them to.

Could hypermile this by avoiding allocating an mcmc client or rbuf for down backends, but that's just more code churn and the efficiency gain is nil. This is probably good to merge to staging.

dormando:
Think the event check conditional for cleanup isn't even necessary... but I want to leave it there out of paranoia.

PR #1058, opened 2023-07-12
proxy: backend antiflap detection.

dormando:
This change allows detecting if backend servers are "flapping" when
given traffic. This means they connect, validate, and can answer
responses, but under normal traffic workloads they break. This could be
due to noisy neighbor issues, network congestion, bugs, etc.

This change also updates an existing setting:
mcp.backend_retry_timeout(seconds), is now:
mcp.backend_retry_waittime(whole seconds)

... this setting is the amount of time to hold a backend in a "bad"
state, before attempting to reconnect again, once it has exceeded its
consecutive failure limit. It must be at least one second and is
specified as whole seconds.

The new tunables are:
mcp.backend_flap_time(seconds) (default: 0)
- Once a backend is successfully connected, if this much time does not
  pass before it throws a connection error (timeout/read/write/etc), we
increase a "flap" counter. This flap counter is checked against
"backend_failure_limit"; once exceeded we consider the backend to be
flapping.

When it is flapping we mark the backend as bad even if it has not
exceeded the _consecutive_ backend_failure_limit count.

Two additional tunables govern how long we wait before trying to connect
again:

mcp.backend_flap_backoff_ramp(seconds) (default: 1.5)
This is a factorial multiplied against backend_retry_waittime() by the
number of times we have consecutively flapped. This increases the
waittime between attempts.

mcp.backend_flap_backeoff_max(whole seconds) (default: 3600)
The maximum amount of time to wait before retrying a flapping server.
With the default we would try at least once per hour.

TODO:
- [x] Make a decision on adding a new log type or overriding the backend_error log.

### DIFF
--- a/proxy_network.c
+++ b/proxy_network.c
@@ -48,8 +48,7 @@ static int _prep_pending_write(struct mcp_backendconn_s *be);
 static void _post_pending_write(struct mcp_backendconn_s *be, ssize_t sent);
 static int _flush_pending_write(struct mcp_backendconn_s *be);
 static void _cleanup_backend(mcp_backend_t *be);
-static int _reset_bad_backend(struct mcp_backendconn_s *be, enum proxy_be_failures err);
-static void _backend_failed(struct mcp_backendconn_s *be);
+static void _reset_bad_backend(struct mcp_backendconn_s *be, enum proxy_be_failures err);
 static void _set_main_event(struct mcp_backendconn_s *be, struct event_base *base, int flags, struct timeval *t, event_callback_fn callback);
 static void _stop_main_event(struct mcp_backendconn_s *be);
 static void _start_write_event(struct mcp_backendconn_s *be);
@@ -90,13 +89,14 @@ static int _proxy_beconn_checkconnect(struct mcp_backendconn_s *be) {
         // FIXME (v2): if a connect fails, anything currently in the queue
         // should be safe to hold up until their timeout.
         _reset_bad_backend(be, P_BE_FAIL_CONNECTING);
-        _backend_failed(be);
         return -1;
     }
     P_DEBUG("%s: backend connected (%s:%s)\n", __func__, be->be_parent->name, be->be_parent->port);
     be->connecting = false;
     be->state = mcp_backend_read;
     be->bad = false;
+    // seed the failure time for the flap check.
+    gettimeofday(&be->last_failed, NULL);
     be->depth = 0; // was set to INT_MAX if bad, need to reset.
     be->failed_count = 0;
 
@@ -105,7 +105,6 @@ static int _proxy_beconn_checkconnect(struct mcp_backendconn_s *be) {
 
     if (_beconn_send_validate(be) == -1) {
         _reset_bad_backend(be, P_BE_FAIL_BADVALIDATE);
-        _backend_failed(be);
         return -1;
     } else {
         // buffer should be empty during validation stage.
@@ -236,7 +235,6 @@ static void _setup_backend(mcp_backend_t *be) {
             event_add(&bec->main_event, &bec->tunables.connect);
         } else {
             _reset_bad_backend(bec, P_BE_FAIL_CONNECTING);
-            _backend_failed(bec);
         }
     }
 }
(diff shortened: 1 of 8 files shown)


dormando:
some by-hand testing shows it does work.

Feels like the backoff should be more aggressive. IE: if the max backoff is an hour and retry time is set to 3 seconds, right now it will retry in 15 seconds, 18 seconds, etc until it hits an hour. This could be cut to maybe 10 tries until it hits the limit.

Might want to remove err: connecting from the flap detector. It ends up marking completely down servers as also immediately flapping. We want to focus on ones that connect/validate but eat dirt when traffic arrives.

Think I just need to let this one sit overnight to see how I feel about the algo tomorrow.

dormando:
Made the above mentioned change re: connection errors not counting against flaps.

It probably makes the most sense to keep few tunables and have a conservative backoff algorithm for now. If anything a "backoff ramp" tunable (or internal) might work, and just eat having to do a for loop on a failure.

```
retry_wait = 3
backoff_ramp = 1.5
for each flap_count above failure_limit do
  retry_wait = retry_wait * backoff_ramp
  break if retry_wait > backoff_max
done
```

should give a sharper ramp and then tunables for backoff_ramp and maybe backoff_max?

default of 3s with ramp of 1.5 gives ~19 tries to hit a 3600s cap, and would take ~5.5 hours to reach the limit. I think any ephemeral problem would pass reasonably within that timeframe.

proxy_network.c
```diff
@@ -748,13 +745,32 @@ static void proxy_backend_retry_handler(const int fd, const short which, void *a
...
-static void _backend_failed(struct mcp_backendconn_s *be) {
-    if (++be->failed_count > be->tunables.backend_failure_limit) {
-        struct timeval tmp_time = be->tunables.retry;
+static void _backend_reschedule(struct mcp_backendconn_s *be) {
+    bool failed = false;
+    struct timeval tmp_time = be->tunables.retry;
+    char *badtext = "markedbad";
+    if (be->failed_count > be->tunables.backend_failure_limit) {
+        failed = true;
+    } else if (be->flap_count > be->tunables.backend_failure_limit) {
```
dormando, line 754: the order of checks here need to be flipped.

proxy_network.c
```diff
@@ -767,6 +783,44 @@ static void _backend_failed(struct mcp_backendconn_s *be) {
...
+            if (flap->tv_sec != 0 || flap->tv_usec != 0) {
+                struct timeval delta = {0};
+                int64_t subsec = 0;
+                gettimeofday(&now, NULL);
+                delta.tv_sec = now.tv_sec - be->last_failed.tv_sec;
+                subsec = now.tv_usec - be->last_failed.tv_usec;
+                if (subsec < 0) {
+                    delta.tv_sec--;
+                    subsec += 1000000;
+                    delta.tv_usec = subsec;
```
dormando, line 804: should probably leave a comment that this code is weird because "tv_usec must have room for at least [-1, 1000000]", meaning technically we can't guarantee going to -2...

dormando:
Having a tough time coming up with more tests that actually exercise code differently. Could iterate through a bunch of crap but it doesn't improve coverage.

dormando:
Decision to make: I assume/hope log parsers are flexible, but maybe we want to leave depth/rbuf in and just always add retry=

dormando:
Looks good in the packet loss stress test:

``
watcher:	OK
watcher:	ts=1689289910.789714 gid=2 type=proxy_backend error=timeout name=10.191.24.56 port=11211 depth=40 rbuf=
watcher:	ts=1689289914.749621 gid=3 type=proxy_backend error=timeout name=10.191.24.56 port=11211 depth=36 rbuf=
watcher:	ts=1689289918.119921 gid=4 type=proxy_backend error=timeout name=10.191.24.56 port=11211 depth=40 rbuf=
watcher:	ts=1689289918.622162 gid=5 type=proxy_backend error=conntimeout name=10.191.24.56 port=11211 depth=40 rbuf=
watcher:	ts=1689289924.723835 gid=6 type=proxy_backend error=timeout name=10.191.24.56 port=11211 depth=40 rbuf=
watcher:	ts=1689289924.723898 gid=7 type=proxy_backend error=markedbadflap name=10.191.24.56 port=11211 retry=15
watcher:	ts=1689289946.781733 gid=8 type=proxy_backend error=timeout name=10.191.24.56 port=11211 depth=40 rbuf=
watcher:	ts=1689289946.781764 gid=9 type=proxy_backend error=markedbadflap name=10.191.24.56 port=11211 retry=22
```

need to decide on the logging situation though. it _should_ use a new type but I _should've_ used a new type for the markedbad stuff. not sure how painful this would be.

dormando:
logging's fine as is.

Issue #972, opened 2022-12-30
Extstore update project


dormando:
Issue tracking a number of extstore changes I intend to make close together. In roughly sorted order.

- [ ] Complete the tiered storage / page management overhaul PR: https://github.com/memcached/memcached/pull/432

This change simplifies a lot of code, adds the ability to split storage across multiple files organized by purpose, and potentially reduces write amplification. It's a good starting point for broader changes.

- [ ] Add warm restart support for extstore: https://github.com/memcached/memcached/issues/871

This change is not complicated but requires a lot of typing. The page data needs to serialize into the restart file, then be parsed and reloaded, plus structure for cleanly stopping and closing all of the extstore files. May delay this to the end in case I want to adjust the page structure.

- [ ] Flush items to extstore's write buffer directly instead of using bg thread: https://github.com/memcached/memcached/issues/541

This work fixes the slab mover's "out of memory while moving object" errors, allowing it to pull memory from the LRU tail instead. Further, during a "set" we will directly flush the tail item to extstore's write buffer and reuse the memory directly. This should end up being an average tradeoff in CPU usage with a small increase in set latency. This may perform better during high load as it removes pressure on the global slab lock. It also removes user pain of having to tune or worry about overwhelming the background flusher thread.

- [ ] Fix dropping of items during compaction if the item is actively referenced.

There's an ancient `TODO` marked in `storage.c:storage_compact_readback()` where if an item header is actively being referenced it cannot "rescue" the valid item into a new page. This item is to just add the missing code to re-allocate and replace the header in this case. Need to also audit for caveats for the now orphaned header item.

- [ ] Fix issues where the same item is written to the same page multiple times, and compaction gets confused.

If you aggressively flush items to extstore (ie; `ext_item_age=0` or have very little memory free, a key that gets frequently updated can end up in the same page multiple times. The compaction code does not try hard enough to validate that the item header it pulled for a key found in storage matches the original object that was written into the page. This can lead to weird bugs where an item is deleted multiple times, etc.

Fixing this might be as simple as ensuring the hdr_it's offset matches the current position in the readback buffer? I got stuck on this before and not sure why I didn't just do that.

- [ ] Change IO thread count setting to be "maximum" number of IO threads, start with 1 thread, and use a heuristic for spawning threads up to the limit.

[...]

PR #1059, opened 2023-07-14
extstore: fix item loss during page defrag

dormando:
If an item that is in the process of being moved to a new disk page during defragmentation, is also being accessed by a client connection, we end up losing the item. This is because we cannot update the item header unless we have exclusive ownership over the item.

This is an extremely rare scenario. You must:
- Be running with item_age=0 or a very low number
- Constantly causing defragmentation
- Get very unlucky during defragmentation.

With this patch we will attempt to re-allocate the item header if we cannot overwrite it implicitly: this is the same process as happens with arithmetic commands in do_add_delta().

TESTING:
- I changed the refcount == 2 requirement so items are always re-allocated, and stepped through it in a debugger while running the overdel tests. Fixed a few bugs and data looks correct.
- Re-ran all of the tests (which complete rescues) with this modified refcount requirement. Everything passes on multiple runs.

I'd like to add some explicit tests but it will be difficult and it's unclear if it's worth the effort here. Will need some level of override or environment variable check if NDEBUG is missing.

Either way I'll let it sit for a day or two before merging and self-review.

### DIFF
--- a/storage.c
+++ b/storage.c
@@ -856,8 +856,32 @@ static void storage_compact_readback(void *storage, logger *l,
                         hdr->offset = io.offset;
                         rescues++;
                     } else {
-                        lost++;
-                        // TODO: re-alloc and replace header.
+                        // re-alloc and replace header.
+                        uint32_t flags;
+                        FLAGS_CONV(hdr_it, flags);
+                        item *new_it = do_item_alloc(ITEM_key(hdr_it), hdr_it->nkey, flags, hdr_it->exptime, sizeof(item_hdr));
+                        if (new_it) {
+                            // need to preserve the original item flags, but we
+                            // start unlinked, with linked being added during
+                            // item_replace below.
+                            new_it->it_flags = hdr_it->it_flags & (~ITEM_LINKED);
+                            new_it->time = hdr_it->time;
+                            new_it->nbytes = hdr_it->nbytes;
+
+                            // copy the hdr data.
+                            item_hdr *new_hdr = (item_hdr *) ITEM_data(new_it);
+                            new_hdr->page_version = io.page_version;
+                            new_hdr->page_id = io.page_id;
+                            new_hdr->offset = io.offset;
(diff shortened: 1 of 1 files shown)


storage.c
```diff
@@ -856,8 +856,30 @@ static void storage_compact_readback(void *storage, logger *l,
                         hdr->offset = io.offset;
                         rescues++;
                     } else {
-                        lost++;
-                        // TODO: re-alloc and replace header.
+                        // re-alloc and replace header.
+                        uint32_t flags;
+                        FLAGS_CONV(hdr_it, flags);
+                        item *new_it = do_item_alloc(ITEM_key(hdr_it), hdr_it->nkey, flags, hdr_it->exptime, sizeof(item_hdr));
+                        if (new_it == 0) {
```
dormando, line 863: eh flip this if. looks ugly.

dormando:
super want to merge this, but paranoid I'm somehow missing something. it's surviving testing so far...

dormando:
fixes an item in #972

storage.c
```diff
@@ -856,8 +856,30 @@ static void storage_compact_readback(void *storage, logger *l,
...
+                            lost++;
+                        } else {
+                            // copy some item data.
+                            new_it->it_flags |= ITEM_HDR;
```
dormando, line 867: pretty sure I need to copy over a few more flags:

token_sent/reserved/stale/key_binary and probably fetched.

probably just need to copy over it_flags wholesale from the old header but mask off or reset ITEM_LINKED.

dormando: remove ITEM_LINKED and ITEM_SLABBED

dormando:
spent a while in a debugger and with extra print output and test fiddling and this now seems right.

have to preserve most of the item internal flags, or else we can lose some important state (stale/win/key binary/etc)

dormando:
released in 1.6.22

PR #1063, opened 2023-08-02
proxy: fix off-by-one if \r is missing

dormando:
A bunch of the parser assumed we only had \r\n, but I didn't actually have that strictness set. Some commands worked and some broke in subtle ways when just "\n" was being submitted.

I'm not 100% confident in this change yet so I'm opening a PR to stage it while I run some more thorough tests.

### DIFF
--- a/proxy_request.c
+++ b/proxy_request.c
@@ -9,7 +9,7 @@
 // where we later scan or directly feed data into API's.
 static int _process_tokenize(mcp_parser_t *pr, const size_t max) {
     const char *s = pr->request;
-    int len = pr->reqlen - 2;
+    int len = pr->endlen;
 
     // since multigets can be huge, we can't purely judge reqlen against this
     // limit, but we also can't index past it since the tokens are shorts.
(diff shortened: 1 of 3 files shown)


proxy_request.c
```diff
@@ -294,15 +294,25 @@ int process_request(mcp_parser_t *pr, const char *command, size_t cmdlen) {
...
+    const char *s = memchr(command, ' ', pr->endlen);
     if (s != NULL) {
         cl = s - command;
     } else {
-        cl = cmdlen - 2;
+        cl = pr->endlen;
     }
     pr->keytoken = 0;
     pr->has_space = false;
-    pr->parsed = cl + 1;
```
dormando, line 305: I suspect this change is wrong but it's passing tests so far, so I need to revisit its usage in a debugger.

dormando:
running the stability suite on it now too.

proxy_request.c
dormando: verified pr->parsed is _only_ used for the multiget garbage.

that reduces the search space... I'll walk through these few functions with a debugger and confirm it's correct.

dormando: looks fine in debugger. tried adding extra spaces and things and it's fine.

dormando:
okay I'm good with this now. it's passed half the stability suite while I stepped through the code in a debugger.. so it's run through the multiget tests several times now.

PR #1065, opened 2023-08-08
Fix proxy response code for backend error

FinnitoProductions:
## Background 
The error code returned by the proxy to the Lua routing layer in the event that one or more backend nodes fail was not being set. As a result, after a backend error, invocations to `res:code()` in Lua returned `0` rather than the intended `MCMC_SERVER_ERROR`, even though the _error status_ was correctly being set to `SERVER_ERROR`. This fix correctly sets the response code in the event of a backend error.

### Manual Testing
I performed manual testing to ensure that the response code was set correctly.

I first added the following print statement to our Lua routing code for proxy writes:
```
print("RES CODE AFTER OPERATION: " .. res:code())
```

Then, I executed the following command sequence on a proxy instance (CAS IDs obscured for sake of example). Between executing the meta-get and meta-set commands, I terminated a backend memcached instance to induce a server error.

```
mg /testnamespace/key v c N100
VA 0 c1234 W

ms /testnamespace/key 4 C1234
data
SERVER_ERROR backend failure
``` 

The emitted log lines in both the original code and the modified code are shown below.

Original Code
```
RES CODE AFTER OPERATION: 0
```

Modified Code
```
RES CODE AFTER OPERATION: 23
```

### Automated Testing
I had intended to create unit tests to test this new error functionality. However, it appeared that a comparable test of the Lua response code already existed [here](https://github.com/memcached/memcached/blob/next/t/proxyunits.t#L1409-L1412). We confirmed that this test is not being skipped, and we are thus unclear on why this test passes for the existing implementation. Any idea @dormando? I am happy to write additional tests, but I wanted to clear up my confusion around the existing tests first -- thanks.

### DIFF
--- a/t/proxyunits.lua
+++ b/t/proxyunits.lua
@@ -552,6 +552,15 @@ function mcp_config_routes(zones)
         return zones.dead(r)
     end
 
+   pfx_get["deadrespcode"] = function(r)
+       local res = zones.dead(r)
+
+       if res:code() == mcp.MCMC_CODE_SERVER_ERROR then
+         return "ERROR code_correct\r\n"
+       end
+       return "ERROR code_incorrect: " .. res:code() .. "\r\n"
+   end
+
     mcp.attach(mcp.CMD_GET, toproute_factory(pfx_get, "get"))
     mcp.attach(mcp.CMD_SET, toproute_factory(pfx_set, "set"))
     mcp.attach(mcp.CMD_TOUCH, toproute_factory(pfx_touch, "touch"))
--- a/proxy_network.c
+++ b/proxy_network.c
@@ -303,6 +303,7 @@ void proxy_run_backend_queue(be_head_t *head) {
                 io = STAILQ_FIRST(&bec->io_head);
                 STAILQ_REMOVE_HEAD(&bec->io_head, io_next);
                 io->client_resp->status = MCMC_ERR;
+                io->client_resp->resp.code = MCMC_CODE_SERVER_ERROR;
                 bec->depth--;
                 return_io_pending((io_pending_t *)io);
             }
@@ -845,6 +846,7 @@ static void _reset_bad_backend(struct mcp_backendconn_s *be, enum proxy_be_failu
         // TODO (v2): Unsure if this is the best way of surfacing errors to lua,
         // but will do for V1.
         io->client_resp->status = MCMC_ERR;
+        io->client_resp->resp.code = MCMC_CODE_SERVER_ERROR;
         be->depth--;
         return_io_pending((io_pending_t *)io);
     }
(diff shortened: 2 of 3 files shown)


dormando:
Can you explain why you think that test covers this scenario?

FinnitoProductions:
I had thought it did, but upon closer inspection it looks like it's testing the scenario in which the backend responds with an error, not when it fails to respond entirely. Is this distinction correct?

dormando:
That is correct.

FinnitoProductions:
Got it, thanks. I just created an automated test which I believe should be testing these changes. It passes with my changes, but it also passes on `next`. I've tried to investigate why that might be, but I can't quite figure it out. Is there something I'm missing here @dormando? Thanks!

FinnitoProductions:
Never mind, feel free to disregard the above -- I  was running `prove` to execute tests without running `make` in-between switching branches. The unit test now passes consistently with my changes and fails consistently on `next` -- looking forward to hearing your feedback! Thanks.

dormando:
did you drop one of the lines during a merge or something?

change looks okay from just a quick read.

PR's for this project are rebased and squashed. so your next task will be to at least rebase it cleanly over `next`. Note that `git reflog` exists in case you get into trouble.

FinnitoProductions:
Should be rebased against `next` now. 

And yes, I dropped one of the lines because the corresponding error check was deleted [here](https://github.com/memcached/memcached/commit/fd4e933e9f2e2ec52cef755c66c9e4d58891e75a#diff-99e8935ed7f4c65f12659540564eaf224e075249bc77cd7ee61b71a7f4a57721L141-L147) (the selected lines are only visible when you `Load diff` on `proxy_network.c`) from `next`, but it still exists on `master`.

FinnitoProductions:
Hi @dormando, just wanted to follow up -- was wondering when you think you'll be able to take a closer look here? No worries if you're still feeling under the weather (get well soon!), I was just hoping to get a sense of the timeline on this

t/proxyunits.lua
```diff
@@ -552,6 +552,15 @@ function mcp_config_routes(zones)
         return zones.dead(r)
     end
 
+   pfx_get["deadrespcode"] = function(r)
+       local res = zones.dead(r)
+
+       if res:code() == mcp.MCMC_CODE_SERVER_ERROR then
+         return "CODE_CORRECT\r\n"
```
dormando, line 559: Small nitpick: In most of these tests I try to use valid protocol in responses. This currently works, but if we ever add strict return code parsing in the test client or server it would fail and the tests would need to be updated.

You can add `ERROR ` in front of both of these `CODE_` lines: `ERROR code_correct\r\n` etc.

dormando:
just one nitpick. If you feel up to it, please squash the branch. Or else I'll take care of that for you.

FinnitoProductions:
Alright, I addressed your change and squashed the branch into one commit. Let me know if there's anything else I need to do!

dormando:
PR is merged to `next` pending the next release, congrats!

dormando:
released in 1.6.22

PR #1074, opened 2023-08-29
log: Fix delete log for 2038

bmwiedemann:
code was added in commit 9b3c946f485309cdf50f7fb1622707b5373e33f3


dormando:
Thanks! I wonder how difficult it would be for me to add a buildbot for this.

I'm going to merge this before the next release but leave it for a little bit so I can see about a buildbot with +16 years clock.

bmwiedemann:
Hi. I think this has been forgotten and should be safe to merge.
I rebased it onto current master.

PR #1078, opened 2023-10-04
proxy: add lua operation label in backend proxy event logs

FinnitoProductions:
## Background
The `label` field passed at the Lua layer, typically used to denote specifics about the given type of operation, is currently only stored in Lua memory and not stored in the internal backend struct. Therefore, proxy events concerning backend errors are currently unable to emit the label provided in Lua.

This change extracts the `label` provided by Lua and stores it in the backend struct. It then logs it as a proxy event whenever a backend failure takes place.  

### Automated Testing
Modified existing backend error proxy event tests to include `label` in their expected output for proxy events.

Created a new unit tests which tests that proxy events for backends without a label provided are correctly emitted.
### Manual Testing
Manually tested by bringing down a single backend node and attempting a meta-get operation on a proxy node. Observed the following events emitted, with hostnames and ports omitted:
```
watch proxyevents
OK
ts=1696457089.21527 gid=6 type=proxy_backend error=disconnected name=[hostname] port=[port] label=[hostname]-read depth=0 rbuf=
ts=1696457089.21784 gid=7 type=proxy_backend error=connecting name=[hostname] port=[port] label=[hostname]-read depth=0 rbuf=
ts=1696457089.21980 gid=8 type=proxy_backend error=connecting name=[hostname] port=[port] label=[hostname]-read depth=0 rbuf=
ts=1696457089.22147 gid=9 type=proxy_backend error=connecting name=[hostname] port=[port] label=[hostname]-read depth=0 rbuf=
ts=1696457089.22150 gid=10 type=proxy_backend error=markedbad name=[hostname] port=[port] label=[hostname]-read retry=3
```

### DIFF
--- a/proxy_network.c
+++ b/proxy_network.c
@@ -773,7 +773,7 @@ static void _backend_reschedule(struct mcp_backendconn_s *be) {
         if (!be->bad) {
             P_DEBUG("%s: marking backend as bad\n", __func__);
             STAT_INCR(be->event_thread->ctx, backend_marked_bad, 1);
-            LOGGER_LOG(NULL, LOG_PROXYEVENTS, LOGGER_PROXY_BE_ERROR, NULL, badtext, be->be_parent->name, be->be_parent->port, 0, NULL, 0, retry_time);
+            LOGGER_LOG(NULL, LOG_PROXYEVENTS, LOGGER_PROXY_BE_ERROR, NULL, badtext, be->be_parent->name, be->be_parent->port, be->be_parent->label, 0, NULL, 0, retry_time);
         }
         be->bad = true;
         be->depth = INT_MAX/2; // fast-path cache for "bad" marker
@@ -856,7 +856,7 @@ static void _reset_bad_backend(struct mcp_backendconn_s *be, enum proxy_be_failu
 
     // Only log if we don't already know it's messed up.
     if (!be->bad) {
-        LOGGER_LOG(NULL, LOG_PROXYEVENTS, LOGGER_PROXY_BE_ERROR, NULL, proxy_be_failure_text[err], be->be_parent->name, be->be_parent->port, depth, be->rbuf, be->rbufused, 0);
+        LOGGER_LOG(NULL, LOG_PROXYEVENTS, LOGGER_PROXY_BE_ERROR, NULL, proxy_be_failure_text[err], be->be_parent->name, be->be_parent->port, be->be_parent->label, depth, be->rbuf, be->rbufused, 0);
     }
 
     // reset buffer to blank state.
--- a/t/proxyunits.t
+++ b/t/proxyunits.t
@@ -202,7 +202,7 @@ sub proxy_test {
     is(scalar <$ps>, "SERVER_ERROR backend failure\r\n", "backend failure error");
 
     # verify a particular proxy event logline is received
-    like(<$w>, qr/ts=(\S+) gid=\d+ type=proxy_backend error=timeout name=127.0.0.1 port=\d+ depth=1 rbuf=EN/, "got backend error log line");
+    like(<$w>, qr/ts=(\S+) gid=\d+ type=proxy_backend error=timeout name=127.0.0.1 port=\d+ label=b\d+ depth=1 rbuf=EN/, "got backend error log line");
 
     # backend is disconnected due to the error, so we have to re-establish it.
     $mbe[0] = accept_backend($mocksrvs[0]);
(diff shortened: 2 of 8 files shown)


proxy_network.c
```diff
@@ -773,7 +773,7 @@ static void _backend_reschedule(struct mcp_backendconn_s *be) {
         if (!be->bad) {
             P_DEBUG("%s: marking backend as bad\n", __func__);
             STAT_INCR(be->event_thread->ctx, backend_marked_bad, 1);
-            LOGGER_LOG(NULL, LOG_PROXYEVENTS, LOGGER_PROXY_BE_ERROR, NULL, badtext, be->be_parent->name, be->be_parent->port, 0, NULL, 0, retry_time);
+            LOGGER_LOG(NULL, LOG_PROXYEVENTS, LOGGER_PROXY_BE_ERROR, NULL, badtext, be->be_parent->name, be->be_parent->port, be->be_parent->label, 0, NULL, 0, retry_time);
```
FinnitoProductions, line 776: I manually confirmed that these are the only two places where we log an event with tag `LOGGER_PROXY_BE_ERROR`.

FinnitoProductions:
@dormando Sorry about that, I have fixed all the unit tests in `proxyunits.t` and `proxyantiflap.t`. I'll be more careful in the future. Everything passes in the unit test suite except for a series of tests which are failing for me on both `next` and this branch, specifically the following tests in `proxy.t`.

```
t/proxy.t                 (Wstat: 5376 Tests: 113 Failed: 21)
  Failed tests:  33, 50, 52, 54, 56, 58, 60, 62, 64, 66
                68, 70, 72, 74, 76, 78, 80, 82, 84, 86
                88
```

This is the failure message for most of them.
```
#   Failed test 'stored test value'
#   at t/proxy.t line 210.
#          got: 'SERVER_ERROR backend failure
# '
#     expected: 'STORED
# '
```
I can't quite figure out why they're failing, but given that these tests are failing for me on `next`, too, I wanted to double check if this was a known issue, or possibly an issue with my setup. Thanks!

dormando:
You can get more info from a particular test by running it directly with a `-v`, ie: `prove -v t/proxy.t` - the proxy.t files are older tests that need to be ported over to proxyunits. They work by starting up actual memcached instances for backends instead of the mocked ones the rest of the tests use.

So if I had to guess `proxy.t` is failing because your dev machine already has something listening on one of the tcp ports the test is using, so it ends up talking to that instead of what it intended to.

FinnitoProductions:
Got it, thanks @dormando — I’ll take a closer look tomorrow to try and fix up my local setup. I can confirm that the rest of the test suite is passing (given that all the failing ones are failing locally on master, too), so should be ready to review — thanks!

t/proxyunits.t
```diff
@@ -169,7 +169,7 @@ sub proxy_test {
 
     print $ps "get /deadrespcode/foo\r\n";
     is(scalar <$ps>, "ERROR code_correct\r\n", "Backend had correct response code on failure");
-}
+ }
```
dormando, line 172: errant space

t/proxyunits.t
```diff
@@ -228,11 +228,40 @@ sub proxy_test {
...
+    is(scalar <$mbe_nolabel>, "get /nolabel/a\r\n", "get passthrough");
+
+    # write a response with partial data.
+    print $mbe_nolabel "VALUE /nolabel/a 0 2\r\nhi\r\nEN";
+
+    # verify the error response from proxy
+    is(scalar <$ps>, "SERVER_ERROR backend failure\r\n", "backend failure error");
+
+    # verify a proxy event logline is received with empty label
+    like(<$w>, qr/ts=(\S+) gid=\d+ type=proxy_backend error=timeout name=127.0.0.1 port=11414 label= depth=1 rbuf=EN/, "got backend error log line");
```
dormando, line 262: small bit: when possible we should avoid tripping timeouts when testing error paths, since it makes the tests take longer to run if you do it often.

Maybe some junk return data from the backend instead?

Unless I misread, sorry.

dormando (changes requested):
I think this is all fine. Just a couple tiny nitpicks, and then please squash the commits down.

I have to do one more pass once it cools off enough to turn my test machine back on :)

Fwiw, best practice: always start with a clean test run before, and after making any changes. Memcached test suite is intended to run correctly 100% of the time. We don't have noisy/ignorable tests, they get deleted or fixed. The first thing I'm going to do with any PR is run the test suite against it and bounce it back without reading if it fails.

FinnitoProductions:
Hi @dormando, quick update -- I made your requested changes and confirmed that all tests passed locally (except for those in `proxy.t`, which I am still investigating). I also squashed the commits down. Let me know if there's anything else you'd like me to change!

dormando:
merged to next, thanks!

dormando:
released in 1.6.22

PR #1085, opened 2023-10-17
proxy: request context refactor and new backend request API

dormando:
See #1098 for merge details. I've decided to keep this PR around for now as development history instead of just squashing this particular branch.

See t/proxyfuncgen.lua for documentation on API changes.

Depresses me greatly to post this still early. I am in a bug squashing and
testing phase.

This change is fully backwards compatible with the old API. If you are using
the old API, some of the C changes around execution of the lua code. All
parsing code (client or backend handling) is unchanged.

Motivation
---
This change revolves around holding context during request execution.
Previously the lua context simply "floated", being carried with a backend IO
request when necessary and resumed during a later callback. It was not
possible to know things for the full duration of a request. It was also not
possible to hold references for the full duration of a request.

For example, previously when lua executes a request against a single pool
(`pool(req)` syntax), the pool and request objects are held in the coroutine's
execution stack, then the coroutine is suspended. This prevents the pool from
being garbage collected while the request executes. Nothing else explicitly
prevents this from happening.

Nor can we pre-allocate memory and request objects. In a typical request, at
minimum:
- a coroutine thread is allocated
- a request object is allocated
- response objects are allocated
- any strings are allocated
- all of the above must later be garbage collected.

This can add up to a lot of CPU. Object creation is especially slow (setting
metatables) and argument checking can get expensive. This puts a high floor on
the CPU usage of a request. With pre-allocation it should be possible to run a
request from start to finish with zero allocations (and thus zero garbage
collection).

Further, the "async API" available via `mcp.await` was supposed to be a
temporary workaround before making the API change available in this pull
request. `mcp.await` is implemented by bolting itself over the existing
system, with hacks to know when to resume the coroutine. Further, it creates
even more allocations:

- the await tracking object
- a table to hold the response objects to be returned to lua

... and itself is limited. There is no way to execute callbacks, to execute
different requests against different pools, to wait on specific pools, and so
on.

Further, there is no way to compose a configuration as a directed graph, as
was done with mcrouter and similar systems. Without this configurations are
less composible, requiring significant code changes to do things like shadow
traffic, pre-warm pools, and so on.

The change
---

We now use a request context to track an incoming request until the response 
has been sent to the client. We use a pre-generation stage to allow
pre-allocating objects, caching lookups, creating callback contexts, and so
on. These pre-allocted request contexts and their associated functions are
re-used until the configuration is reloaded and overwritten.

See `t/proxyfuncgen.lua` for a description of the API and examples for how to
use it. Easier than describing it more here.

TODO
--

The remaining TODO items are split into "potential testing/merge blockers" and
"blocking usage of the new API"

Full blockers:
- [x] handle "TODO: return rctx" in `proto_proxy.c`
- [x] handle "FIXME: allow to fail" in `proto_proxy.c`
- [x] trace usage of `finalize_rctx_cb` and ensure nothing is missed
- [x] counters for how many funcgens exist and slots are in use (cannot trace
  on a per-funtion basis, just global)
- [x] The basic test work noted below.

Completion work:
- [x] Replace fgen free list with STAILQ linked list to remove error handling
  around allocation [not doing for now, low/no risk]
- [x] See if `max_queues` argument can be removed by executing a "learning run"
  with the function generator.
~~Argument for setting the size limit of a request object (might be global)~~ (change too big for this pr)
- [x] Cleanup code for funcgens being replaced (will expand this into sub-items)
- [x] Various cleanups around the funcgen code (TODO/FIXME's in the code)
~~Disallow further queuing from callbacks~~ (delaying; need to advise users to not do this for now)
- [x] Disallow calling wait funcs while waiting.
- [x] Don't allocate a request if a request context is a subcontext
- [x] Pre-allocate IO objects into queue slots to avoid runtime memory error handling
- [x] Pre-allocate result objects into queue slots where possible
- [x] Cleanups around non-result-object result handling (strings, nil, etc)
- [x] `rctx:wait_handle()` - accept just the handle if a queue slot already
  activated with a request

Test work:
- [ ] See `t/luafuncgen.t` - fill out tests until I run out of ideas.
- [ ] Separate test file for funcgen reload tests
- [x] Separate test file for testing the builtin router
- [ ] Specific funcgen tests added to the shredders test suite.
- [x] Executing the shredders stabilization and performance suites for the
  original API.

Maybe work:
- [x] A builtin prefix router, which avoids pulling the key into lua and
  executing a slow regular expression check. This adds up to a lot of CPU
usage...
- [x] extend router to accept command maps. ie: prefix -> mg -> func.
- [ ] "Max slots" argument to cap how many parallel requests of a particular function are allowed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -62,6 +62,7 @@ memcached_SOURCES += proto_proxy.c proto_proxy.h vendor/mcmc/mcmc.h \
 					 proxy_ratelim.c \
 					 proxy_jump_hash.c proxy_request.c \
 					 proxy_network.c proxy_lua.c \
+					 proxy_luafgen.c \
 					 proxy_config.c proxy_ring_hash.c \
 					 proxy_internal.c \
 					 md5.c md5.h
--- a/memcached.c
+++ b/memcached.c
@@ -1022,6 +1022,14 @@ void resp_reset(mc_resp *resp) {
         resp->item = NULL;
     }
     if (resp->write_and_free) {
+#ifdef PROXY
+        if (resp->proxy_res) {
+            LIBEVENT_THREAD *t = resp->bundle->thread;
+            pthread_mutex_lock(&t->proxy_limit_lock);
+            t->proxy_buffer_memory_used -= resp->wbytes;
+            pthread_mutex_unlock(&t->proxy_limit_lock);
+        }
+#endif
         free(resp->write_and_free);
         resp->write_and_free = NULL;
     }
@@ -1104,6 +1112,7 @@ static mc_resp* resp_allocate(conn *c) {
             }
             b->next = 0;
             b->prev = 0;
+            b->thread = th;
             th->open_bundle = b;
             resp = &b->r[0];
             memset(resp, 0, sizeof(*resp));
@@ -1230,6 +1239,14 @@ mc_resp* resp_finish(conn *c, mc_resp *resp) {
         resp->item = NULL;
     }
     if (resp->write_and_free) {
+#ifdef PROXY
+        if (resp->proxy_res) {
+            LIBEVENT_THREAD *t = resp->bundle->thread;
+            pthread_mutex_lock(&t->proxy_limit_lock);
+            t->proxy_buffer_memory_used -= resp->wbytes;
+            pthread_mutex_unlock(&t->proxy_limit_lock);
+        }
+#endif
         free(resp->write_and_free);
     }
     if (resp->io_pending) {
--- a/memcached.c
+++ b/memcached.c
@@ -876,7 +876,7 @@ static void conn_cleanup(conn *c) {
 
     conn_release_items(c);
 #ifdef PROXY
-    if (c->proxy_coro_ref) {
+    if (c->proxy_rctx) {
         proxy_cleanup_conn(c);
     }
 #endif
--- a/memcached.c
+++ b/memcached.c
@@ -3778,8 +3778,13 @@ static int server_sockets(int port, enum network_transport transport,
             const char *tagstr = "tag";
             if (strncmp(p, tagstr, strlen(tagstr)) == 0) {
                 p += strlen(tagstr);
-                if (*p == '[') {
+                // NOTE: should probably retire the [] dumbassery. those're
+                // shell characters.
+                if (*p == '[' || *p == '_') {
                     char *e = strchr(p, ']');
+                    if (e == NULL) {
+                        e = strchr(p+1, '_');
+                    }
                     if (e == NULL) {
                         fprintf(stderr, "Invalid tag in socket config: \"%s\"\n", p);
                         free(list);
--- a/memcached.h
+++ b/memcached.h
@@ -780,6 +780,9 @@ typedef struct _mc_resp {
      */
     bool skip;
     bool free; // double free detection.
+#ifdef PROXY
+    bool proxy_res; // we're handling a proxied response buffer.
+#endif
     // UDP bits. Copied in from the client.
     uint16_t    request_id; /* Incoming UDP request ID, if this is a UDP "connection" */
     uint16_t    udp_sequence; /* packet counter when transmitting result */
@@ -794,6 +797,7 @@ typedef struct _mc_resp {
 struct _mc_resp_bundle {
     uint8_t refcount;
     uint8_t next_check; // next object to check on assignment.
+    LIBEVENT_THREAD *thread;
     struct _mc_resp_bundle *next;
     struct _mc_resp_bundle *prev;
     mc_resp r[];
--- a/memcached.h
+++ b/memcached.h
@@ -861,7 +861,7 @@ struct conn {
     int io_queues_submitted; /* see notes on io_queue_t */
     io_queue_t io_queues[IO_QUEUE_COUNT]; /* set of deferred IO queues. */
 #ifdef PROXY
-    unsigned int proxy_coro_ref; /* lua reference for active coroutine */
+    void *proxy_rctx; /* pointer to active request context */
 #endif
 #ifdef EXTSTORE
     unsigned int recache_counter;
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -408,6 +408,13 @@ void proxy_submit_cb(io_queue_t *q) {
 // pool calls and mcp.await()
 void proxy_return_rctx_cb(io_pending_t *pending) {
     io_pending_proxy_t *p = (io_pending_proxy_t *)pending;
+    if (p->client_resp->blen) {
+        // FIXME: workaround for buffer memory being external to objects.
+        // can't run 0 since that means something special (run the GC)
+        unsigned int kb = p->client_resp->blen / 1000;
+        lua_gc(p->rctx->Lc, LUA_GCSTEP, kb > 0 ? kb : 1);
+    }
+
     if (p->is_await) {
         p->rctx->async_pending--;
         mcplib_await_return(p);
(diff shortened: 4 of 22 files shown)


dormando:
looks like I get some time this week for this PR.

Added a few more tests as a warm-up. Next, going to fix some assert's in result handling from sub-rctx functions. Already a TODO for it.

dormando:
and I immediately hurt my back again :( not much time at the desk this week.

dormando:
and now we can track the number of functions (so long as they are given a name) and number of slots. should be able to reuse this same system to track backends by label too, hooray. at least with infrequently updating data like the hostname/port and current status (up/down/dead/etc).

dormando:
think the next thing is to hit the "full blockers" TODO list then start tests for more error handler coverage.

could probably run the shredders suite on it now in parallel. first run with the existing suite with just the old API's, then again with some new API's in use.

dormando:
pretty low on critical work now. most of the remaining are optimizations or code audits. Some of the missing error handling is impossible/rare to trigger.

next to do:
- stability tests with new API in use.
- more unit tests.
- cry. I dunno. nobody reads this.

dormando:
Ended up needing to write the internal router function. Using a funcgen for a ton of slots has scaling issues for a few reasons; mainly that it caches the subrctx's for every slot. So every instance of a prefix router would hold 20+ sub-slots from each sub-fgen. Every new parallel request had to generate a new one of everything.

Rather than disable the subrctx caching (either conditionally or unconditionally) I decided it needed doing anyway. It's a big performance win overall.

It's working but some cleanups remain before committing.

dormando:
Done for the day. Builtin router works... works pretty well, and has a good abstraction to allow me to optimize and extend it later.

Should be a huge performance boost especially for longer keys, since we were copying the whole string around a few times and likely kicking the GC way more often. Won't be noticeable if your configuration is slow for other reasons.

Next is a specific test file for the rest of the router configuration options.

After that just going to chip on small things for at most an hour a day. pretty sick of working on this.

dormando:
Ah, I forgot to add the "Autocreate a funcgen" from mcp.attach() to mcp.router_new(). that's probably necessary.

dormando:
Updated the TODO's list a bit. commenting here because it won't be obvious or notify otherwise.

dormando:
The plan on making request objects dynamically sized will have to wait until after this PR is merged. The request render routine will need to be rewritten to allow in-place editing since it won't be reasonable to use a 64k stack array and copy the string to/from like I currently do.

Since that's a larger change I have to eject it from this plan, but it can happen soon after so not a big worry...

dormando:
Want to tackle removing the `max_queues` argument next but I feel like that's going to be a ton of code.

I want to keep the queue memory in-line with the request context to cut a layer of pointer chasing (queues as external memory) plus avoid memory expansion code. (queues expanding as they get assigned).

The idea:
- create a fake rctx that just knows `queue_assign` and count the handles.
- run the generator once to "learn" how many queues are needed
- assign that to the fgen and etc from there.

However, I think limiting the queues explicitly gives some extra safety (accidental or unintentional loops).

So maybe I'll add a global for `funcgen_queues_default` and default that to like 3. If you have 1-2 unused slots it won't affect much, so you'd only need to realistically override it when making larger function generators.

dormando:
Also today I'm going to make one API change:
orig: `res = rctx:wait_handle(r, h)`
new:
```
res = rctx:queue_and_wait(r, h)
res = rctx:wait_handle(h)
```

I keep running into a pattern of queueing everything at once and then wanting to wait on just a handle, but the only way to do that is to pass a request in.

dormando:
Big problem with running changes this long is that I want to keep changing them. The `max_queues` and `rctx:queue_assign()` stuff is _fine_ but now it's bothering me enough that I might change the syntax.

dormando:
Think the rest of the work is super minor at this point. The TODO items around error handling and object creation are still blockers but they are tiny amounts of work so I'll do them very slowly.

Still plenty of tests to write, and adding command maps to the builtin router is pretty important. It should be ready to write some shredders api2 stability tests as well.

From this point I'm just going to try to write a library with it for a couple weeks, in case I run into any other issues that I'd prefer to revise the API for.

dormando:
Took a break; still need to try writing a library with it. Will fill out missing unit tests and add shredders stability/perf tests in parallel.

dormando:
having a hard time deciding how to finalize this. My main worry is that I might want to change it further once I've been writing library code with it, but I also don't want to hold the change from merging. I've been working on unit tests/stabilization in parallel.

I'm having a hard time coming up with a library design that I like. `simple.lua` is a good structure but sucks.

In the short term I think running the stability suite again and adding API2 specific tests for stability/perf are probably important. Then I'll try the library work again.

dormando:
Hate distractions. This project has been absolutely cursed from the beginning.

Just kicked off the pre-existing stability suite to validate compat mode. Will try _again_ to write new stability/perf shredders tests tomorrow.

dormando:
Stability test suite was showing high memory usage. On comparing test runs between `next` and `api2` I saw api2 using much higher buffer memory usage, with some variability.

Finally remembered seeing this earlier in the development: currently only `res` objects are being created (and then thrown away), and buffer memory from reading an object off the network is being reclaimed by the lua garbage collector directly.

Response memory can/should have its ownership handed off to the `mc_resp` object after completion of an `rctx` call, then freed/decremented after the resp is written to the network successfuly.

then rctx can reset any result memory that wasn't handed off as soon as the slot is returned.

however this is all crap I was supposed to add then forgot about. I'm going to look at it a bit more and then consider adding it post-release and just having us stay slightly shy of being perfectly "allocation free" for the sake of this PR.

This means I need to at least tune the GC to account for buffer memory being allocated, as result objects are all tiny and the buffer memory is external to it.

dormando:
found a second "memory leak" via the shredders stability suite. GC got behind with mcp.internal in certain situations.

shredders performance suite completed overnight (13 hours). haven't looked at the results yet; kicked off a fresh run with `next` to compare to.

debating if I have the energy to do the pre-made res object part before merge...

dormando:
bunch of minor behavior changes and function renames today.

got a performance run completed on `next` finally, so the machine's free to do the API2 stability test run I'd intended to do today :)

It's looking like I can finish the burn down by sunday. I'd prefer tomorrow (Friday) but there's probably too much left.

dormando:
passed stability test last night on the first try, yay.

proceeding with the burn down today... got a few more renames, tests, a couple error handlers, and router code to finish.

dormando:
pain in the ass burn-down taking forever.

I'm relatively confident in the backcompat modes now, as both stability and perf have been doing just fine. might leverage this to merge API2 in an "experimental" state for one release cycle. it's going to be treated that way anyway. during that time I can chip away at adding more shredders tests without being afraid of making other changes and causing a rebase nightmare here.

just a handful of important gotchas left in my list. mostly avoiding accidental recursion or re-queueing requests.

`mcp.internal` + extstore needs dedicated tests because it has none so far.

internal router needs some command map features.

everything else can happen in extra PR's. more tests, features, code cleanups. if I merge it I can convert `simple.lua` and get some testing going at least.

dormando:
writing tests for `mcp.internal` + extstore and it's pretty badly broken. working on it...

dormando:
Alright... got just a couple small things left and then I'm going to call a merge. I'll mark API2 as experimental briefly.

The compat path has been validated quite a bit so anyone not using API2 shouldn't have issues.

This should unblock development of the main tree as many things depend on this being merged, and I've been avoiding adding a lot into `next` or cutting releases to prevent rebase time churn.

I'll open an issue with a list of important follow-ups once this is merged.

dormando:
headache all day. try again tomorrow..

dormando:
tired. not much done. try again tomorrow...

pretty much down to just footgun-safety checks and extending the router for command maps.

dormando:
Alright... just going to take a look at the router object then call it for merge.

still plenty of tests I can write but can backfill those without blocking the whole project further.

PR #1088, opened 2023-10-20
Use malloc_trim to return memory to OS

liangxs:
One `page` is freed by `slabs.c:memory_release` when `cache_memlimit` command adjust memory limitation. But sometimes `glibc` do not return the freed memory to operating system, which make users confusing with memcached's memory `Resident Size`. The following example shows that about 25GB memory is not return to OS.

image

[malloc_trim](https://man7.org/linux/man-pages/man3/malloc_trim.3.html) can release memory back to the system.


dormando:
Think there're some cases where this doesn't work, and this isn't checking the result of the function call.

Might be already if it's just an optimistic attempt.. I'll see about testing this, sorry for the delay.

PR #1090, opened 2023-11-03
docs: use spaces instead of tab character

namsic:
Currently doc directory subfiles have inconsistent indentation.
Change the tab characters in the doc/*.xml files to spaces and include them in the whitespace test.

I think the same change could be applied to some other files(`*.xsl`, `*.full`, ...).

Are these files excluded from testing because they are immutable?
(Does a license that does not allow modifications apply?)

### DIFF
--- a/devtools/clean-whitespace.pl
+++ b/devtools/clean-whitespace.pl
@@ -4,7 +4,6 @@ use FindBin qw($Bin);
 chdir "$Bin/.." or die;
 
 my @exempted = qw(Makefile.am ChangeLog doc/Makefile.am README README.md md5.c md5.h);
-push(@exempted, glob("doc/*.xml"));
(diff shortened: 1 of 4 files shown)


dormando:
Those xml files used to be generated from the xsl's... so we had no control over the tabs there. That tool bitrotted so I ended up just committing the generated files via an older OS. I'm not expecting them to change anymore so I left the exemption and am just letting them stay stale.

Did something break or is this just something you noticed?

namsic:
> Did something break or is this just something you noticed?

This still works fine now. I just noticed.

Unlike Makefiles, `doc` don't necessarily require tabs. So I didn't think there was a need to exclude them from testing.

Thanks for the explanation. I'm closing this PR.

PR #1093, opened 2023-11-16
slabs: add stats for exact bytes accounting

dormando:
add quick summary at bottom of stats slabs output which says exactly how many bytes are available for use and how many are used. when bytes_available reaches zero the system must evict.

however the system _may evict_ much sooner than when bytes_availble reaches zero. you must also account for the system moving memory around slab classes.

combine with the output from totalling 'stats items' -> mem_requested to find the slab overhead.

TODO:
This doesn't tell us if we're about to evict, as that calculation is:
- if any slab class has zero chunks available
- the page allocs are at the memory limit
- and `slab_global_page_pool` is 0
- the next allocation will evict.

### DIFF
--- a/slabs.c
+++ b/slabs.c
@@ -555,6 +555,7 @@ unsigned int global_page_pool_size(bool *mem_flag) {
 /*@null@*/
 static void do_slabs_stats(ADD_STAT add_stats, void *c) {
     int i, total;
+    uint64_t bytes_used = 0;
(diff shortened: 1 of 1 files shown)


dormando:
don't think I'm going to merge this but I'll leave it here as food for thought:
- some kind of total allocation counters (include various buffers, hash table, etc)
- algorithmic summaries that are commonly used

dormando:
not doing this right now.

PR #1101, opened 2024-01-09
Proxy tweaks

dormando:
PR with a couple small pre-release commits for the proxy.

- Do not use the iothread by default anymore. Adds `mcp.backend_use_iothread(bool)` to restore original behavior.
- Finishes support for sub-command maps in routers

Might add one or two other small things here after reviewing the TODO lists, then merge this and cut 1.6.23.


dormando:
decided to delay more work until after a release, so I can do docs and shit.

PR #1103, opened 2024-01-18
proxy: add global rate limiter

dormando:
If created during mcp_config_pools() and then passed into mcp_config_routes(), utilizes a centralized rate limiter.

TODO:
- [x] step through with debugger to validate the objects/proxies are correct
- [x] add self_ref code (during proxy setup?)
- [x] generalize the "manager" code used by pools for de-allocating pools
- [x] port pool and global ratelim to the general code.
- [x] add __gc routine for global_tbf so it can destroy its mutex.

waiting a few days then more code review/testing and merge.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -214,7 +214,7 @@ struct proxy_tunables {
     bool down; // backend is forced into a down/bad state.
 };
 
-typedef STAILQ_HEAD(pool_head_s, mcp_pool_s) pool_head_t;
+typedef STAILQ_HEAD(globalobj_head_s, mcp_globalobj_s) globalobj_head_t;
 typedef struct {
     lua_State *proxy_state; // main configuration vm
     lua_State *proxy_sharedvm; // sub VM for short-lock global events/data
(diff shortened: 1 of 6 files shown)


dormando:
Think the next time I need to add one of these (an object copyable between config vm and worker vm's) I should be able to convert it all into a more generic system.

For now I just want to convert the GC side to be generic and save my brainpower for other things. The proxy object copy code is trivially small.

dormando:
man I really want to make this More Magic, IE:
- not require a different type of ratelim object
- instead of you create it in pools() and pass it into routes() it does magic proxy stuff

instead of having a different new() and such... however I'd have to pad the struct with a bunch of garbage since I can't re-allocate a pre-existing userdata.

dormando:
gg e-z.

since the GC routine is just using the exactly pool GC code I don't have to do a ton of validation work. pool's got good test coverage already.

dormando:
took a few extra days to finally circle back to this. added the missing GC routines. needs some more specific tests or a gbd walk to confirm.

really glad a decent pattern has come out for this kind of thing. the next time I do it I'll generalize even more code and it'll get easier.

2421Sd (commented):
No está bien

PR #1104, opened 2024-01-25
fix: prevent giving negative number with -R, -m options

ing-eoking:
I think giving negative number when using -R, -m options rarely happens.
But, It makes the program behave abnormally.

Other options were written consistent with the log message like below.
```c
case 't':
    settings.num_threads = atoi(optarg);
    if (settings.num_threads <= 0) {
        fprintf(stderr, "Number of threads must be greater than 0\n");
        return 1;
    }
```
Only those two are different. Are those intentional?

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -5145,7 +5145,7 @@ int main (int argc, char **argv) {
             break;
         case 'R':
             settings.reqs_per_event = atoi(optarg);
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! Those options are probably just super old and never revisited. This is probably fine to add.

dormando:
released in 1.6.24

PR #1108, opened 2024-02-13
Proxy lua GC tuning

dormando:
Tries to move the GC run from inline requests to outside of requests.

Passes tests but not benchmarked yet. Might not be interacting properly with the "GC Pause" tunable on when to actually kick off a GC.

Also avoids running GC during code reload and ignores memory allocated during code reload.

TODO:
 - [x] Change the old GCSTEP calls to instead increment `t->proxy_vm_extra` and add during poke
 - [x] Doesn't currently emulate the internal "GC pause" algorithm well, might end up using too much CPU. edit: seems fine.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -19,18 +19,92 @@ static void *mcp_profile_alloc(void *ud, void *ptr, size_t osize, size_t nsize);
 /******** EXTERNAL FUNCTIONS ******/
 // functions starting with _ are breakouts for the public functions.
 
-// We want to ensure we slowly iterate through VM memory in two cases:
-// - If using the newer zero-allocation API, old backends will never close
-// after reload.
-// - If the server is completely idle and user is issuing reloads, they will
-// be confused when old backends do not close.
-// So every few seconds we push one minimal GC step.
-static void proxy_gc_poke(evutil_socket_t fd, short event, void *arg) {
-    LIBEVENT_THREAD *t = arg;
+static inline void _proxy_advance_lastkb(lua_State *L, LIBEVENT_THREAD *t) {
+    int new_kb = lua_gc(L, LUA_GCCOUNT);
+    if (new_kb > t->proxy_vm_last_kb) {
+        // clamp increases of the base memory per run to help
+        // prevent runaway scenarios.
+        t->proxy_vm_last_kb += (new_kb - t->proxy_vm_last_kb) * 0.5;
+    } else {
+        t->proxy_vm_last_kb = new_kb;
+    }
+}
+
+// The lua GC is paused while running requests. Run it manually inbetween
+// processing network events.
+void proxy_gc_poke(LIBEVENT_THREAD *t) {
     lua_State *L = t->L;
+    int vm_kb = lua_gc(L, LUA_GCCOUNT) + t->proxy_vm_extra_kb;
+    if (t->proxy_vm_last_kb == 0) {
+        t->proxy_vm_last_kb = vm_kb;
+    }
+
+    // equivalent of luagc "pause" value
+    int last = t->proxy_vm_last_kb;
+    if (t->proxy_vm_gcrunning <= 0 && vm_kb > last * 2) {
+        t->proxy_vm_gcrunning = 1;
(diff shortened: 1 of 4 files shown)


dormando:
did some quick shredders performance tests with pipelined metaget requests on API1 (ie; trying to make the GC stand out vs just pure syscalls)

... latency is actually considerably improved. CPU usage is very slightly higher but requests are bucketing much more in lower microsecond ranges.

kind of a waste because moving to API2 removes this latency/CPU penalty entirely, but here we are.

dormando:
in "backend IO, low client total, low GC pressure vs syscall pressure" test this is very slightly worse (probably owing to slightly extra CPU usage). The more pipelining (requests per syscall) or GC pressure the better it seems to do.

At 8 get batches the latency improvement is something like 30% at any request rate.

dormando:
Should probably test "banking" the memory debt until memory usage doubles to see if that helps with the CPU load, but I'm confident that'll make latency worse so I won't spend much time on it.

dormando:
stability test ran without any memory bloat... need some reload tests with more GC ballast I think.

Will skip the extra memory alloc thing and memory banking thing.

dormando:
not working super great. turns out it's hard to decide when and how hard to run a GC.

proto_proxy.c
```diff
@@ -19,17 +19,28 @@ static void *mcp_profile_alloc(void *ud, void *ptr, size_t osize, size_t nsize);
...
+            t->proxy_vm_extra_kb = 0;
+        }
+    }
+}
+
+// every couple seconds we force-run one GC step.
+// this is needed until after API1 is retired and pool objects are no longer
+// managed by the GC.
+static void proxy_gc_timerpoke(evutil_socket_t fd, short event, void *arg) {
+    LIBEVENT_THREAD *t = arg;
```
dormando, line 41: Note: 

- set a flag so timerpoke does nothing if the GC did something recently
- check the result of lua_gc in timerpoke and reset proxy_vm_last_kb if it finished a routine

proto_proxy.c
```diff
@@ -19,17 +19,28 @@ static void *mcp_profile_alloc(void *ud, void *ptr, size_t osize, size_t nsize);
...
-    LIBEVENT_THREAD *t = arg;
+// Attempting to push the lua GC algorithm to run outside of when requests are
+// being processed.
+void proxy_gc_poke(LIBEVENT_THREAD *t) {
     lua_State *L = t->L;
+    int vm_kb = lua_gc(L, LUA_GCCOUNT) + t->proxy_vm_extra_kb;
+    // only add GC debt with what we grew past the last reload size.
+    if (vm_kb > t->proxy_vm_last_kb * 1.5) {
+        // reset our counter once a full GC has completed
+        if (lua_gc(L, LUA_GCSTEP, vm_kb - t->proxy_vm_last_kb)) {
```
dormando, line 30: this should step the amount of change since the last time this visited. currently seems to just chuck the full delta of debt in each time.

could also try:
- start when VM size doubles
- run minimal steps
- if VM size is triple, start increasing amount passed to LUA_GCSTEP

dormando: also instrument this with GCISRUNNING to see if I'm accidentally unpausing it.

dormando:
Made a big lua proxy config that let me manually stress the GC in various ways (loading coroutines, tables, strings, etc). Also instrumented the lua GC with a bunch of prints so I can see when and where it does work (and how much). Then spent a while jabbing at the standard GC and then this modified GC runner.

Need to run A/B against the performance suite again to validate it. Last time even with the algo being wrong it had a pretty strong inprovement in latency.

dormando:
Bench test looks nice. Moves a big batch of requests from 100us bucket into the 10us bucket. fewer total outliers into the ms ranges. overload behavior looks alright. seems to wobble a bit.

might run a full perf suite in A/B before merge or might not. haven't decided.

going to sleep on it and check in the morning: paranoid of a condition where memory ends up growing forever because the GC gets behind. probably worth a run through the stability suite.

dormando:
stability suite finished. memory looked stable. it jumped when tests changed.

PR #1112, opened 2024-02-28
fix build on uclibc-ng

ffontaine:
Fix the following build failure with uclibc-ng raised since version 1.6.18 and https://github.com/memcached/memcached/commit/875371a75cbf1f92350de2d1fa0fae4a35ed572b:

```
/home/buildroot/autobuild/instance-2/output-1/host/lib/gcc/arc-buildroot-linux-uclibc/10.2.0/../../../../arc-buildroot-linux-uclibc/bin/ld: memcached-thread.o: in function `thread_setname':
thread.c:(.text+0xea2): undefined reference to `pthread_setname_np'
```

Fixes:
 - http://autobuild.buildroot.org/results/e856d381f5ec7d2727f21c8bd46dacb456984416

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -694,6 +694,7 @@ AC_CHECK_FUNCS(clock_gettime)
 AC_CHECK_FUNCS(preadv)
 AC_CHECK_FUNCS(pread)
 AC_CHECK_FUNCS(eventfd)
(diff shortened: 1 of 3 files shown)


dormando:
thanks!

does uclibc-ng have any method of changing the thread name?

ffontaine:
`pthread_setname_np` is enabled on uclibc if the toolchain has been compiled with NPTL (Native Posix Thread Library) support. Otherwise, there is no function to change the thread name.

dormando:
merged for the next release. thanks!

PR #1114, opened 2024-03-06
proxy: add new request meta flag APIs

dormando:
`req:flag_add("F", token)`

If token is nil, just adds "F" to request.
If token is "example", adds "Fexample" to request. returns boolean false if failed (flag exists, request too long)

`req:flag_set("F", token)`

If token is nil, adds "F" to request.
If token is "example", adds "Fexample" to request. returns boolean false if failed (request too long)

`req:flag_replace("F", "N", token)`

If token is nil, adds "N" to request.
If token is "example", adds "Nexample" to request. returns boolean false if failed (request too long)

These functions are able to modify the underlying request string without creating lua string garbage. IE:

`req:flag_add("F" .. token)` would create a new string then pass that into the function, needing to GC it later.

TODO:
- [ ] Unit tests.


dormando:
seems good but I want test coverage for the two other functions before merging.

dormando:
Believe this is done. Going to sit on it for a few hours then review the tests one more time.

PR #1115, opened 2024-03-07
Meta CAS Override

dormando:
Allows overriding the CAS value if the `E` flag is supplied.

Pretty deep internal change, but it made some bugs more obvious.

This shouldn't break any existing feature of CAS, and users who elect to override should use CAS values that typically increase. There's no reason that two different items cannot share the same CAS value either, so data can be independently versioned or CRC'ed, or both.

TODO:

### Audit the CAS override code

I went over it with a comb today and can't find anything else wrong and can't think of anything else to test. I looked at every `ITEM_get_cas` and every `ITEM_set_cas` in the code. I validated that both of these macros require `ITEM_CAS` flag to be set on an item already, and that gets set during initial allocation. So times when we call the macros but system CAS might be disabled they will do the right thing.

Still should run through the commands with CAS disabled to ensure they don't throw errors.

###  `flush_all` relies on item CAS to cope with a timing glitch.

[DONE]

Ancient `flush_all` had a bug where no items would be accepted for the duration of the second in which `flush_all` was called. In 2015 I improved this by using an "oldest CAS" value to be a tie breaker, so new values would be immediately settable. However if you disable CAS you still get the brokenish behavior.

Looking at it now, it seems like we can just use an offset for the `current_time` variable:
- on flush, set `oldest_live` to `current_time`
- increment `current_time` via an offset added to it or by straight up incrementing it, allowing items to be set again (worst case items live in the same second for a bit longer than usual).
- Since `flush_all` isn't something most people run constantly this will be completely fine.
- Thus remove reliance on CAS

### Online enable/disable of `stats sizes` relies on strict CAS ordering

[DONE]

There's an open issue dealing with a bug/race condition around turning `stats sizes` on and off at runtime. Given we now have `metadump` it might be best to just let the online on/off part of that feature die.

### DIFF
--- a/items.c
+++ b/items.c
@@ -59,7 +59,7 @@ static unsigned int sizes[LARGEST_ID];
 static uint64_t sizes_bytes[LARGEST_ID];
 static unsigned int *stats_sizes_hist = NULL;
 static int stats_sizes_buckets = 0;
-static uint64_t cas_id = 0;
+static uint64_t cas_id = 1;
 
 static volatile int do_run_lru_maintainer_thread = 0;
 static pthread_mutex_t lru_maintainer_lock = PTHREAD_MUTEX_INITIALIZER;
@@ -488,7 +488,7 @@ static void item_unlink_q(item *it) {
     pthread_mutex_unlock(&lru_locks[it->slabs_clsid]);
 }
 
-int do_item_link(item *it, const uint32_t hv) {
+int do_item_link(item *it, const uint32_t hv, const uint64_t cas) {
     MEMCACHED_ITEM_LINK(ITEM_key(it), it->nkey, it->nbytes);
     assert((it->it_flags & (ITEM_LINKED|ITEM_SLABBED)) == 0);
     it->it_flags |= ITEM_LINKED;
@@ -501,7 +501,7 @@ int do_item_link(item *it, const uint32_t hv) {
     STATS_UNLOCK();
 
     /* Allocate a new CAS ID on link. */
-    ITEM_set_cas(it, (settings.use_cas) ? get_cas_id() : 0);
+    ITEM_set_cas(it, cas);
     assoc_insert(it, hv);
     item_link_q(it);
     refcount_incr(it);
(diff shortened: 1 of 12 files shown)


dormando:
Need to think on the `flush_all` stuff now. If we rate limit to run once per second it'll be bug free, but someone might get mad. who would do that on purpose outside of a test suite?

Maybe a second settings flag for "time was adjusted" so we don't double adjust current_time? kill the flag each time time updates...

dormando:
`flush_all` is a pita. without CAS we need to restore the old behavior, but with proper locking...

old behavior was:
- set oldest_live to `current_time - 1`
- walk each LRU and boot everything newer than this time

with CAS behavior:
- set oldest_live to `current_time`
- set oldest_cas to the next cas-id
- skip everything with older cas or time

`lru_pull_tail` is full featured but only pokes tails. think need to bite the bullet and just walk every possible LRU and check the heads, but with the proper locks this time.

dormando:
Tests all pass. time to take the `stats_sizes` command out back and shoot it.

dormando:
Well doing this feature has helped fix a bunch of old issues/bugs:

- CAS changing in certain conditions.
- `item_flush` sucks without CAS enabled
- `stats sizes_disable` can crash the daemon
- handful of good improvements to the test suite.

Now I need to lay some paranoia on it and see if I can come up with more unit tests. I don't have a shredders stability suite for the core storage system yet... not sure if I want to start writing that for_this_ particular change. Shredders needs more cleanup first.

dormando:
tempted to upstream the other two commits first. they're independent of the CAS change.

items.c
```diff
@@ -123,14 +123,9 @@ void set_cas_id(uint64_t new_cas) {
 
 int item_is_flushed(item *it) {
     rel_time_t oldest_live = settings.oldest_live;
-    uint64_t cas = ITEM_get_cas(it);
-    uint64_t oldest_cas = settings.oldest_cas;
-    if (oldest_live == 0 || oldest_live > current_time)
-        return 0;
-    if ((it->time <= oldest_live)
```
dormando, line 130: note to self: audit all callers to `item_is_flushed` to make sure they skip the item crawler and exptime == 0 stuff. they should, and the new line below is exactly what was moved to this function originally... but paranoia.

dormando: all good

dormando:
upstreamed the dependent fixes. now the commit looks nice and lean and easier to validate.

memcached.c
```diff
@@ -2357,7 +2357,7 @@ enum delta_result_type do_add_delta(LIBEVENT_THREAD *t, const char *key, const s
         }
         memcpy(ITEM_data(new_it), buf, res);
         memcpy(ITEM_data(new_it) + res, "\r\n", 2);
-        item_replace(it, new_it, hv);
+        item_replace(it, new_it, hv, (settings.use_cas) ? get_cas_id() : 0);
```
dormando, line 2360: This has to be wrong.

dormando: nevermind. it's being swapped out by the caller... which is fine since it's not actually used anywhere else.

dormando:
Need to update `doc/protocol.txt`. just re-re-re-re-re-re-read the code. can't think of anything else that's not already tested.

the add_delta stuff sucks a little bit, but I'm planning a refactor of how that works this year which would remove the COW on refcount issue.. so that's fine.

dormando:
All merged up.

PR #1116, opened 2024-03-08
proxy: allow early freeing rcontexts

dormando:
Any allocated request context would stay in memory attached to its function generator until the next reload, which would replace the function generators. A replaced fgen would eventually run out of attach references and cleanup itself, removing all rcontexts.

A one-time burst of concurrent request contexts would thus take up memory forever. With this change we use an easing function to slowly free unneeded request contexts.

This can help for both memory usage scenarios and instances where the GC is still used in the request path and tail latency can become poor.

TODO:

- [x] Need specific tests.

There aren't tests monitoring `stats proxyfuncs` so something specific to this is probably necessary. We need to validate it across reloads and during various easing scenarios after temporary concurrent blasts.

Should be easier to test concurrency via pipelined requests than doing lots of client sockets. Numbers don't have to be huge they just have to be at least a few and then drop back down after many non-concurrent requests.

- [x] add slot bloating to the reload test

Should make a memory leak related to said cleanup more obvious.

### DIFF
--- a/proxy_luafgen.c
+++ b/proxy_luafgen.c
@@ -4,6 +4,7 @@
 
 static mcp_funcgen_t *mcp_funcgen_route(lua_State *L, mcp_funcgen_t *fgen, mcp_parser_t *pr);
 static int mcp_funcgen_router_cleanup(lua_State *L, mcp_funcgen_t *fgen);
+static void _mcplib_funcgen_cache(mcp_funcgen_t *fgen, mcp_rcontext_t *rctx);
 static void mcp_funcgen_cleanup(lua_State *L, mcp_funcgen_t *fgen);
 static void proxy_return_rqu_cb(io_pending_t *pending);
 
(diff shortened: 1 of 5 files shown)


proxy_luafgen.c
```diff
@@ -323,52 +394,16 @@ static void mcp_funcgen_cleanup(lua_State *L, mcp_funcgen_t *fgen) {
...
-        name = "mcp_router";
-    } else {
-        name = "anonymous";
-    }
     LIBEVENT_THREAD *t = PROXY_GET_THR(L);
-    mcp_sharedvm_delta(t->proxy_ctx, SHAREDVM_FGEN_IDX, name, -1);
+    mcp_sharedvm_delta(t->proxy_ctx, SHAREDVM_FGEN_IDX, fgen->name, -1);
 
     // Walk every request context and issue cleanup.
     for (int x = 0; x < fgen->total; x++) {
```
dormando, line 401: The actual list length is `free_max`. Prior code would only free when `free == total`.

Need to change this to one of:
- walk the local registry (for each non-integer value in table)
- walk until `free_max`, but during alloc/realloc of the free list we need to set all new entries to NULL

dormando:
I'm out of steam and will resume monday. Tests prove it seems to mostly work. It sees total memory grow slightly when I increase the number of slots to test, but that could be lua stacks/registry expanding and the change is a few kilobytes so it's hard to tell if it's a leak or not.

Had to rearrange a lot of code. need to give one more thorough check that the lua stack is balanced properly when freeing data.

maybe extend tests for sub-rctx's and routers as well.

dormando:
derp. one more fix (that didn't show up in lua memory..)

dormando:
Edited tests locally to:

- Load a few hundred slots
- Reload
- Repeat once per second forever

config VM memory and worker VM memory were 100% stable but total RSS would very slowly grow.

ran under valgrind and found one chunk of leaked memory per reload. Fixed and ran 500+ reload cycles with RSS being perfectly still.

Probably should complicate the configuration more and re-run the tests:
- Prefix router
- Multiple funcgens
- Sub-funcs

+ run the change through shredders suite. then merge.

dormando:
latest commit only affects usage of v2 with sub-routes.

proxy_luafgen.c
```diff
@@ -26,16 +27,103 @@ int mcplib_funcgen_gc(lua_State *L) {
 // These "generator functions" handle pre-allocating and creating a memory
 // heirarchy, allowing dynamic runtimes at high speed.
 
+// must be called with fgen on top of stack in fgen->thread->L
+static void mcp_rcontext_cleanup(mcp_funcgen_t *fgen, mcp_rcontext_t *rctx, int fgen_idx) {
+    lua_State *L = fgen->thread->L;
+    luaL_unref(L, LUA_REGISTRYINDEX, rctx->coroutine_ref);
```
dormando, line 33: I think it's better for general performance to get all of these references into the fgen local reference table.

Then there are numerous smaller tables instead of one gigantic lua registry, and those smaller tables get released on reload.

probably not worth doing in this commit, but I'll look in case it's easy and relatively safe.

dormando:
Got a not wonderful but worktable shredders test going which demonstrates the slots bursting and then bleeding off quickly. The algorithm is definitely conservative with allowing a lot of alloc/frees to avoid leaving slot counts high for a long time.

I'd also like counters for slotcache hit/miss rate but that doesn't look trivial either.

I'll add to the TODO backlog for remaining optimizations and some extra counters when possible. Think this is done.

PR #1118, opened 2024-03-14
Add a couple of sanity checks

QuChen88:
- Handle failure in `malloc()` for temp port number file initialization to avoid NULL pointer de-referencing
 - Ensure value of `c->rbytes` (which is `int` type) is positive before using it in memmove, which implicitly cast it into `size_t`. This avoids undesired behavior if the value of `c->rbytes` is negative.
 - Avoid implicit conversion of unsigned int to signed int for `keylen` in `proto_bin.c`


dormando:
this got released a while ago, oops. closing.

PR #1120, opened 2024-03-28
proxy: `stats proxybe` to show bad backends

dormando:
stats command intended for backend state information.

so far just outputs "bad_label" [count] for each backend currently marked bad. The count is the number of backend objects marked bad.

If in "per worker thread mode" (default) if all workers recognize the backend as down, the count will be the same as the number of worker threads. If the IO thread is in use it would be 1.

If a backend has normal status, it is not included in the output.

TODO:
- [x] was hand testing.. need a specific unit test.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -251,8 +251,39 @@ void process_proxy_funcstats(void *arg, ADD_STAT add_stats, conn *c) {
             snprintf(key_str, STAT_KEY_LEN-1, "slots_%s", name);
             APPEND_STAT(key_str, "%d", slots);
         } else {
-            // TODO: Is it safe to delete keys in the middle here?
-            // not worried at all about just leaking memory here.
+            // TODO: It is safe to delete keys here. Slightly complex so low
+            // priority.
+        }
+    }
+
+    pthread_mutex_unlock(&ctx->sharedvm_lock);
+}
+
+void process_proxy_bestats(void *arg, ADD_STAT add_stats, conn *c) {
+    char key_str[STAT_KEY_LEN];
+    if (!arg) {
+        return;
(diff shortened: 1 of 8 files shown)


dormando:
sharedvm thing is pretty convenient for doing this, but isn't the _fastest_ and is globally locked either way. Thus trying to only poke it when things go into and out of bad status. Counting "alive" backends would hit the shared state a lot harder.

Other things could be added in the future, like a failure count if it is marked bad, retries since marked bad, etc.

I know there's been a request to *see* all of the configured backends but I'm not completely convinced this is a good idea - they wouldn't be organized into pools or anything, but a randomly ordered list of backends. Printing all of the pools and their backends would require querying the configure thread instead and having it iterate things... so a different problem.

proto_proxy.c
```diff
@@ -1305,6 +1336,16 @@ void mcp_sharedvm_delta(proxy_ctx_t *ctx, int tidx, const char *name, int delta)
     pthread_mutex_unlock(&ctx->sharedvm_lock);
 }
 
+void mcp_sharedvm_remove(proxy_ctx_t *ctx, int tidx, const char *name) {
```
dormando, line 1339: ended up not using this but might leave the code in anyway..

dormando:
In writing the unit test I ran into a bug I'd seen before: if a backend connection fails after the TCP connection is established it will never be marked bad. The bad status and failure count were being reset after the TCP connection was established, but it should have been after the full validation stage.

So this fixes that too, or else tests couldn't run. I was going to try to upstream the fix first but this whole change is trivial so it can go together.

dormando:
merged up, whoop.

PR #1122, opened 2024-04-02
proxy: wait timeout API

dormando:
give request contexts the ability to time out on wait operations. this allows moving along the logic without cancelling in flight requests or resetting backends.

TODO:
- [x] Might go with an extra argument instead of `_timeout()` versions of each function.

This is due to the API still missing a few variants of `wait` (ie: enqueue and wait cond, wait and return "best", etc). Internally the `_timeout()` version is identical except for the extra argument parsing. So with `_timeout()` variants we'd have like ten wait functions...

- [x] port to `wait_handle` and etc once above is decided.
- [x] add `rctx:sleep()` for testing disaster scenarios
- [x] remove an armed alarm during rctx cleanup phase.
- [x] complete the testing. basic tests pass as of this writing.

not feeling great today so finishing this tomorrow.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -847,7 +847,9 @@ static void _proxy_run_tresp_to_resp(mc_resp *tresp, mc_resp *resp) {
 int proxy_run_rcontext(mcp_rcontext_t *rctx) {
     int nresults = 0;
     lua_State *Lc = rctx->Lc;
-    int cores = lua_resume(Lc, NULL, 1, &nresults);
+    assert(rctx->lua_narg != 0);
+    int cores = lua_resume(Lc, NULL, rctx->lua_narg, &nresults);
+    rctx->lua_narg = 1; // reset to default since not-default is uncommon.
     size_t rlen = 0;
     conn *c = rctx->c;
     mc_resp *resp = rctx->resp;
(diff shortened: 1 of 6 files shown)


dormando:
upside of no `_timeout()` variants: remove a bunch of API boilerplate. allow users to optionally timeout or not by just passing an extra 0 or removing the argument.

downside of no `_timeout()` variants: fixed overhead of checking for the extra argument.

I'm currently using `lua_isnumber()` (which does a little fiddling) for the 4th arg check... but I could reduce the overhead by using `argc = lua_gettop()` at the top of the functions and just checking for the number, then use tointegerx if the arg count check passes. This could apply to the argument checking that's already in `wait_cond` for a very mild speedup.

Maybe that's good enough.

dormando:
alright, made a decision and removed the dedicated `_timeout()` API's. need to make one more decision on whether to continue using a float argument or instead spec milliseconds. would be a pretty mild reduction in overhead but mostly for reasons of user confusion.

dormando:
fixed a bug, wrote more tests. want to a few more tests but need to think. interface is pretty simple.

dormando:
disappointed I can't make `rctx:sleep()` work. API1 needs to die very soon.

dormando:
This looks complete to me. Going to let it sit for a few more hours or another day then do another self-review and merge.

dormando:
lol dammit. I completely forgot to merge this.

dormando:
re-reviewed and merged.

PR #1123, opened 2024-04-06
fix typos in 2 files

RoboSchmied:
fix same typo 5 times


dormando:
Thanks!

Sadly we won't be taking this PR: these two files are vendored files pulled from an outside repository, so we avoid making changes to them directly so updating the files is simpler. You can report those issues to the owners of the original files.

PR #1125, opened 2024-04-10
proxy: config thread cron functions

dormando:
Adds a system for periodically running functions in the configuration thread. These crons may also signal to reload the configuration after they run. IE: you may shell out or run a module to fetch and download new json data, then reload the configuration.

mcp.register_cron("name",
  { every = seconds, func = function()
    print("cron running")
  end })

Also accepts { rerun = false } to run the cron once after a reload. Can be used to check something after a reload has run, or simply issue a reload after an exact amount of time since the previous reload finished.

Crons that are not seen after a config reload are unloaded. IE: the only crons that may run must have been registered during the last configuration reload.

If a cron is overwriting itself, and the 'every' period has not changed, it will "inherit" the next scheduled run time. Thus config reloads will not interrupt the scheduling of crons that are not changing their time schedule.

Also adds `mcp.config_reload()` which will schedule the system to reload its configuration after the cron finishes running.

TODO:
- [x] some lua tests.
- [x] maybe rename `mcp.config_reload` to `mcp.schedule_config_reload()` to better indicate it doesn't immediately run the reload.

It's already been manually tested with the config:
```lua
if bcount == nil then
    bcount = 0
end

function set_crons()
    mcp.register_cron("foo",
        { every = 1, func = function()
            print("FOO function called")
        end })

    mcp.register_cron("bar",
        { every = 6, func = function()
            print("BAR function called: ", bcount)
            bcount = bcount + 1
        end })
    mcp.register_cron("reload",
        { every = 10, func = function()
            print("CONFIG RELOAD")
            -- signals to reload, does not immediately reload
            mcp.config_reload()
        end })

    -- will run once per reload.
    mcp.register_cron("once",
        { every = 2, rerun = false, func = function()
            print("NO RERUN!")
        end })
end

function mcp_config_pools()
    set_crons()
end

function mcp_config_routes()
    -- do nothing.
end
```

reloaded while commenting out `set_crons()` to confirm they unload/reload/etc. Will do some basic automated tests but I don't think it needs a lot.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -125,10 +125,12 @@ void proxy_stats(void *arg, ADD_STAT add_stats, conn *c) {
        return;
     }
     proxy_ctx_t *ctx = arg;
-    STAT_L(ctx);
 
+    STAT_L(ctx);
     APPEND_STAT("proxy_config_reloads", "%llu", (unsigned long long)ctx->global_stats.config_reloads);
     APPEND_STAT("proxy_config_reload_fails", "%llu", (unsigned long long)ctx->global_stats.config_reload_fails);
+    APPEND_STAT("proxy_config_cron_runs", "%llu", (unsigned long long)ctx->global_stats.config_cron_runs);
+    APPEND_STAT("proxy_config_cron_fails", "%llu", (unsigned long long)ctx->global_stats.config_cron_fails);
     APPEND_STAT("proxy_backend_total", "%llu", (unsigned long long)ctx->global_stats.backend_total);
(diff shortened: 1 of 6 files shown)


dormando:
I wrote tests and found an actual bug! The bug is fixed!

Might still rename `mcp.config_reload()`. going to wait until tomorrow morning and give it a final review then merge-up.

dormando:
Made some adjustments:

- removed the sysevents log entries on every cron run. too noisy.
- added a couple counters for cron runs + fails (failures were already logged too)
- renamed config_reload to schedule_config_reload

should be good to go now.

PR #1126, opened 2024-04-11
proxy: `mcp.server_stats(subcmd)`

dormando:
Allow accessing server stats from the configuration thread (useful for crons, checking start args, etc). Does not support cachedump or detail.

Returns a table of the results. The results are thus unordered.

---

Much smaller code change than I expected, though I have to rely on an unfortunate global variable until some further refactoring of the stats commands.

TODO:

- [x] Decide on if we want to order or add further structure to some of these.

It's not hard to parse the ordered outputs (items, conns, slabs) since you can just loop up until the numbers miss, but it might be more convenient to be able to loop the table directly sometimes. It would add a lot of code, I think?

How would that work? Detect :'s and make a sub-table until the # stops matching or the : goes away?

- [x] Add an `mcp.log` that works from config thread so the unit tests can check results. Among other uses...
- [x] Add some more explicit tests. Currently it's not bad since the stats will either output 100% of the time or crash, more or less...
- [x] Ensure ustats are tested or at least audited... have a bad feeling about the lock order there.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1796,7 +1796,7 @@ void append_stat(const char *name, ADD_STAT add_stats, conn *c,
 }
 
 /* return server specific stats only */
-void server_stats(ADD_STAT add_stats, conn *c) {
+void server_stats(ADD_STAT add_stats, void *c) {
     pid_t pid = getpid();
     rel_time_t now = current_time;
 
@@ -1861,7 +1861,7 @@ void server_stats(ADD_STAT add_stats, conn *c) {
     APPEND_STAT("get_expired", "%llu", (unsigned long long)thread_stats.get_expired);
     APPEND_STAT("get_flushed", "%llu", (unsigned long long)thread_stats.get_flushed);
 #ifdef EXTSTORE
-    if (c->thread->storage) {
+    if (ext_storage) {
         APPEND_STAT("get_extstore", "%llu", (unsigned long long)thread_stats.get_extstore);
         APPEND_STAT("get_aborted_extstore", "%llu", (unsigned long long)thread_stats.get_aborted_extstore);
         APPEND_STAT("get_oom_extstore", "%llu", (unsigned long long)thread_stats.get_oom_extstore);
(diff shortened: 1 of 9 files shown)


dormando:
Another question: do I enable this from the protocol handlers? I don't think so.

dormando:
having trouble deciding on the best path here, so I'll hold from merging it for a little bit and do the stats stuff.

at very least I'll add the `mcp.log` bit upstream and test ustats and stuff.

dormando:
Added `mcp.log` and updated the tests a bit. Seems pretty good.

Lock order for proxy stats with ustats would be: `config_lock -> stats_lock -> worker_stats_lock`. It looks fine from an audit.

Still punting on deciding how the results are organized. Going to come back to it.

dormando:
Aight... need to make a decision here today and get this merged. Question is if I should make sub-tables for slabs, items, conns, or present the data as an array with key/value tuples inside.

Probably no on the latter, since the user would have to loop to find individual values.

On the former, pro's:
- Easier to just iterate the table since all of the data related to a specific "slot" will be within a sub-table. IE: walking the table directly now would show slab 5 chunk size, then slab 2 chunk count, etc.
- Otherwise you would have to know all of the keys for a category ahead of time in order to blind walk and ship them

Cons:
- Mostly performance related. More tables means more garbage to iterate.

Maybe:
- For now, just make more garbage.
- In future, allow taking an extra argument which is the stats result from a previous run
- This would replace the values (mostly numbers) while keeping the keys and sub-tables, generating almost no garbage.
- Downside is that subcategories that "Disappear" underneath (like from stats conns or items) would have stale data since we wouldn't know to delete them from the table without doing even more work.

Maybe:
- Pass in the previous stats output as an argument (should be for the same command!)
- Make a new base table
- Copy over sub-tables that are seen and update those
- Still makes a new main table but reduces garbage + string allocs significantly.

Maybe that's too much and I'll just accept the garbage rate...

dormando:
think this is ready for self review...

dormando:
should be ready for merge.

PR #1130, opened 2024-05-02
proxy: fix race condition leading to hang

dormando:
If a lua function schedules extra requests after an initial set of requests, we need to use a workaround to ensure the second set of requests is actually submitted to the proxy backends.

This workaround was in the wrong place. If connections are being cut and reconnected between batches of requests it's possible for the workaround to trigger on a connection object that has since moved to another thread.

This patch tightens up the workaround to run before the client connection has a chance to resume.

---

TODO:
- [x] Passes unit test suite
- [x] Passes basic stability bench tests
- [x] Passes partial suite stability bench tests
- [x] Passes full suite full time stability bench tests

### DIFF
--- a/memcached.h
+++ b/memcached.h
@@ -753,6 +753,7 @@ typedef struct {
     pthread_mutex_t proxy_limit_lock;
     int proxy_vm_extra_kb;
     int proxy_vm_last_kb;
+    unsigned int proxy_vm_negative_delta;
     int proxy_vm_gcrunning;
     bool proxy_vm_needspoke;
     uint64_t proxy_active_req_limit;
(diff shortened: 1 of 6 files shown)


dormando:
note: new set of tests seem to hit an unrelated (hopefully)  bug causing a slot leak, but requires mixing worker io and non worker io in the same request. going to run the full suite without it then focus on that as a separate bug.

but I won't merge this until I understand the other bug.

looks like 980,000 funcgens is about 7G of RAM...

dormando:
Full suite passed at 10s per subtest. re-running with 60s per subtest against the same running daemon.

dormando:
Added some extra GC tuning to cope with the new stability tests being too hard on the debug binary, mostly.

Re-running the test suite at 60s per test now.

I generally understand the nature of the slot leak I ran into and it does not appear to be related to this. I'll merge this change once the main suite runs through then get back to work on that other bug.

dormando:
Longer form test suite finally passed at nearly 6 hours.

dormando:
Looking at another bug that came up before deciding to merge this.

dormando:
re-running stability suite one more time with the final fix. 30s per test (default) should be fine.

dormando:
passed all tests. merged up.

PR #1132, opened 2024-05-10
proxy: improve gc handling of mcp.request

dormando:
Since userdata have to pass through the GC twice before being collected, we usually make their allocations look oversized to the GC. This wasn't being done with mcp.request() because I don't personally see it causing problems.


dormando:
wasn't the bug, closing.

PR #1136, opened 2024-05-15
Reorder reads and writes for itemstats aggregation.

aractnido:
Reorder reads and writes for itemstats aggregation to leverage SIMD instructions, improving code generation by ensuring that the variables are accessed in the order they are declared in memory. This change optimizes data locality and enhances the compiler's ability to generate efficient vectorized code.

See https://godbolt.org/z/7GGEM1e7j for a comparison of the generated assembly.

No functional changes.


dormando:
This is cool, thanks!

dormando:
released in 1.6.28

PR #1140, opened 2024-05-21
proxy: backend TLS support

dormando:
WIP. Connects, handshakes, and routes requests as of this commit. Needs further work before mergeable.

TODO:
- [x] Flesh out error handling. 
- [x] ~parameter for write buffer size~ de-scoping
- [x] Reduce the size of the code change in `proxy_network.c` by refactoring
- [x] Configure option for enabling support instead of hard coding
- [x] `proxy_tls.h` to blank out functions if TLS not enabled at compile time
- [x] enum for common return codes, vs the 0/-1/-2 currently in use
- [x] ~add some specific counters for handshake/tls errors~ de-scoping
- [x] SSL_shutdown and SSL_free() calls from gc are missing.
- [x] Move some initialization code to config thread.
- [x] Per-backend TLS option
- [x] Ensure all new TODO/FIXME notes are dealt with.

TODO MAYBE:
- [ ] Ship error detail to proxyevents log stream (not useful right now, may punt to next PR)

ETA: maybe a week? It does currently work, so it might not take long to finish.

GOALS:
- Enable TLS on a global or per-backend basis. Propagate errors properly.
- Minimize `#ifdef` soup and testing burden by using a shim for accessing OpenSSL and leaving some TLS related code always compiled in. The shim nulls itself out if TLS support is not compiled in.

FOR FUTURE PR's:
- Load and reload client certificates/keys
- Peer certificate validation
- sessions / session tickets
- maybe: sidethread for handshake calcs (if not reusing a session)
- io_uring preparedness.

I'm separating certificate work out of this PR so I can keep the change focused for initial validation. Tired of trying to waterfall every individual feature.

NOTES:
There is a companion PR to this work over in mcshredder: https://github.com/memcached/mcshredder/pull/5 - this is a much more complicated approach that would be necessary to add `io_uring` back to the proxy. Since we're not currently using io_uring I used a simpler approach that lets OpenSSL handle the socket fd and run syscalls itself.

Since `proxy_network.c` was already structured to abstract out the read and write socket calls into just two small functions, this allowed the PR to be much less invasive than it would be otherwise.

However, if you are naive you might think: "why aren't you using a function ptr for read/write and modify even less?" - well, OpenSSL is way more complicated than that and we need to change how we interact with the event loop based on what the _library_ returns vs what any underlying socket thinks. It sucks.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,6 +65,7 @@ memcached_SOURCES += proto_proxy.c proto_proxy.h vendor/mcmc/mcmc.h \
 					 proxy_luafgen.c \
 					 proxy_config.c proxy_ring_hash.c \
 					 proxy_internal.c \
+					 proxy_tls.c proxy_tls.h \
 					 md5.c md5.h
 endif
 
--- a/configure.ac
+++ b/configure.ac
@@ -130,6 +130,9 @@ AC_ARG_ENABLE(proxy,
 AC_ARG_ENABLE(proxy-uring,
   [AS_HELP_STRING([--enable-proxy-uring], [Enable proxy io_uring code EXPERIMENTAL])])
 
+AC_ARG_ENABLE(proxy-tls,
+  [AS_HELP_STRING([--enable-proxy-tls], [Enable proxy io_uring code EXPERIMENTAL])])
+
 AC_ARG_ENABLE(werror,
   [AS_HELP_STRING([--enable-werror], [Enable -Werror])])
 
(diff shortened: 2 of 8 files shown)


dormando:
Debating... looks like I could really cut down on the dedicated code by using a few function pointers, but:
- would require deeper change to existing code. (ie; making functions for the main recv/writev calls)
- they have to be stored/fetched from somewhere.

Not sure it's worth it vs trying to slim down the existing functions slightly and just accepting some small code duplication with the benefit of not potentially breaking anything.

Should remove the duplication from the `_set_event` stuff though.

dormando:
might leave the error handling with correct structure but not very communicative. until we do certificates and peer checks there aren't many reasons outside of network for things to fail... that would significantly cut down the time to merge this initial pass. it already has more correct error handling than the main mc code :(

dormando:
last things I'm considering doing is some missing init code... then just run tests and merge as-is with experimental tags.

dormando:
this is as far as I care to take this right now; will leave it experimental and merge. it may or may not work.

PR #1141, opened 2024-05-25
memcached-tool: add -u flag to unescape special chars in keys names

piwai:
The "lru_crawler metadump all" command called by "keys" option escapes unsafe special chars with corresponding %xx codes. For example, a key named "keywith+" will be shown in "memcached-tool keys" as:

key=keywith%2B exp=... la=... cas=...

But in some cases in can be useful to have the original key name printed as-is. This commit adds a "-u" flag to do the unescaping of lru_crawler output before displaying it.

### DIFF
--- a/scripts/memcached-tool
+++ b/scripts/memcached-tool
@@ -16,11 +16,13 @@
 
 use strict;
 use IO::Socket::INET;
+use URI::Escape;
 
 my $addr = shift;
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! Merged up.

dormando:
released in 1.6.28

PR #1144, opened 2024-06-01
proxy: str = rctx:tls_peer_cn()

dormando:
Returns a string copy of the TLS peer CN entry. Returns nil if none exists or unable to parse the certificate.

---

Untested but compiles.

TODO:

- [ ] Test it a bit.


dormando:
rebased

PR #1146, opened 2024-06-11
proxy: match, token = req:match_res(res)

dormando:
If req has 'k' and/or 'O' flags, match against the result object.

TODO:
- [x] unit tests. demonstrate match + failures
- [x] test sets and such
- [x] need to decide how to handle requests without k/O.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -803,6 +803,7 @@ int mcplib_request_flag_set(lua_State *L);
 int mcplib_request_flag_replace(lua_State *L);
 int mcplib_request_flag_del(lua_State *L);
 int mcplib_request_gc(lua_State *L);
+int mcplib_request_match_res(lua_State *L);
 void mcp_request_cleanup(LIBEVENT_THREAD *t, mcp_request_t *rq);
 
 // response interface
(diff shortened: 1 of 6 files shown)


dormando:
Did some manual testing in a debugger with:
```
function mcp_config_pools()
    local b1 = mcp.backend('b1', '127.0.0.1', 11215)
    return mcp.pool({b1})
end

function mcp_config_routes(p)
    local fgen = mcp.funcgen_new()
    local near = fgen:new_handle(p)
    fgen:ready({ n = "match", f = function(rctx)
        return function(r)
            local done = rctx:enqueue_and_wait(r, near)
            local res = rctx:res_any(near)
            print("checking match")
            local matched, token = r:match_res(res)
            if matched then
                print("request and response match")
            else
                print("mismatch: " .. token)
            end
            return res
        end
    end})
    mcp.attach(mcp.CMD_ANY_STORAGE, fgen)
end
```

dormando:
unit tests added. fixed a bug and changed behavior slightly

dormando:
check failure seems to be something wrong with github actions... just says "internal error"

PR #1147, opened 2024-06-17
add Dockerfile example for building with proxy support. Needs openssl…

sevrinrobstad:
Add Dockerfile example for building with proxy-support. Needs openssl-dev to build.


dormando:
Thanks!

What failure were you getting with openssl missing? It's supposed to build without it unless you add some extra ssl options.

dormando:
Just pushed a fix for that; if you don't use the proxy TLS features it won't require openssl to build.

dormando:
think the underlying issue was resolved

PR #1148, opened 2024-06-19
proxy: router refactor and cmap fallback

dormando:
Allows specifying a commandmap on a router as well as a map and default route. Lookup order is:
- Check map, if entry in map, check that
- If none matched, check fallback command map
- If nothing matched, return default route handler if exists

Also removes the router struct from the funcgen object, saving 32 bytes and letting me extend the size of the router struct without bloating every allocated fgen. This will later allow inlining the lookup map into the router struct for a lookup speedup.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -656,31 +656,6 @@ enum mcp_rqueue_e {
     QWAIT_SLEEP,
 };
 
-enum mcp_funcgen_router_e {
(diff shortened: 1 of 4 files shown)


dormando:
Fixes some missing API functionality from mcrouter, fixes an annoyance with memory allocation, and prepares for a later CPU optimization.

PR #1149, opened 2024-06-19
proxy: ustats internal refactor

dormando:
If a ustat changes name for the same index, its value is reset to 0.

Cleans up some confusing code around how ustats are handled; the global ctx and worker threads now use different structures with different intent, instead of different sides of the same struct.

Also reorders the entries in the global ctx so the most frequently accessed ones are clustered toward the front of the struct.

### DIFF
--- a/proto_proxy.c
+++ b/proto_proxy.c
@@ -184,8 +184,9 @@ void process_proxy_stats(void *arg, ADD_STAT add_stats, void *c) {
     buffer_memory_limit = ctx->buffer_memory_limit;
 
     // prepare aggregated counters.
-    struct proxy_user_stats *us = &ctx->user_stats;
-    uint64_t counters[us->num_stats];
+    struct proxy_user_stats_entry *us = ctx->user_stats;
+    int stats_num = ctx->user_stats_num;
+    uint64_t counters[stats_num];
     memset(counters, 0, sizeof(counters));
 
     // TODO (v3): more globals to remove and/or change API method.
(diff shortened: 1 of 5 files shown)


dormando:
this change prepares for a potential performance fix (at least benchmarking the effect of) of inlining the ustats names in the global struct vs using a pointer for each. If the `stats proxy` function has to stop and fetch lots of random areas of memory it can have significant performance drop.

However this needs to be proved and the code prepared for it since that change would use a little excess memory.

dormando:
eh. maybe it could "compile" the names down into a compact buffer during the reload stage. now that I have a separate structure for the config side vs the worker side that's pretty easy actually.

1024 stats would be over 130k of just name memory, which sucks if the names are much shorter than the limit (128 bytes).

debating just doing this now... if I also unroll the snprintf it'll get a significant CPU boost. then people using billions of counters wouldn't be punching themselves in the face as hard.

proto_proxy.c
```diff
@@ -212,10 +213,31 @@ void process_proxy_stats(void *arg, ADD_STAT add_stats, void *c) {
...
+        char *e = NULL; // ptr into vbuf
+        const char *pfx = "user_";
+        const size_t pfxlen = strlen(pfx);
+        for (int x = 0; x < stats_num; x++) {
+            if (us[x].cname) {
+                char *name = ctx->user_stats_namebuf + us[x].cname;
+                size_t nlen = strlen(name);
+                if (nlen > STAT_KEY_LEN-6) {
+                    // impossible, but for paranoia.
+                    nlen = STAT_KEY_LEN;
```
dormando, line 227: -6

proxy.h
```diff
@@ -191,11 +191,19 @@ struct proxy_int_stats {
 };
 
 struct proxy_user_stats {
-    size_t num_stats; // number of stats, for sizing various arrays
-    char **names; // not needed for worker threads
+    int num_stats; // number of stats, for sizing various arrays
     uint64_t *counters; // array of counters.
 };
 
+// TODO: performance test inlining the name[128] vs having an external string
```
dormando, line 198: remove

dormando:
merged to staging. fixes a lot of longstanding annoyances.

PR #1151, opened 2024-06-24
OpenSSL API usage cleanup

dormando:
After reviewing the frontend TLS code it became obvious that we weren't using the API right: the return and error codes from OpenSSL API calls must be handled by more OpenSSL API calls, and can't be simply returned upward like read/write commands are normally.

Except that does normally work, because it was working. Except maybe when it wasn't, but we wouldn't know because the errors weren't being handled or printed anywhere.

This series fixes:
- Isolates OpenSSL headers to the `tls.c` file, as a general code cleanup step.
- Updates the minimum OpenSSL version to the already-EOL'ed 1.1.1, which allows us to rely on some API calls not in 1.1.0
- Properly break out the error handling for SSL accept/read/write, and ship errors to the `watch connevents` log stream.

TODO:
- [x] Counters for SSL specific errors (so a user would know to look at the log stream for a specific error)
- [ ] Still figuring out if the `SSL_CTX` needs to be re-created during certificate refresh, or not. Their documentation is a little hard to parse here. It does appear to be safe because it's been in heavy use. The worry is if OpenSSL changes something internally in the future which makes it not safe.
- [ ] Do more research into how WANT_WRITE during SSL_read/etc can affect the state machine. It's unclear if this can even happen unless the socket is full, which it never will be.


dormando:
https://wiki.openssl.org/index.php/TLS1.3
```
TLSv1.3 sends more non-application data records after the handshake is finished. At least the session ticket and possibly a key update is send after the finished message. With TLSv1.2 it happened in case of renegotiation. [SSL_read()](https://www.openssl.org/docs/manmaster/man3/SSL_read.html) has always documented that it can return SSL_ERROR_WANT_READ after processing non-application data, even when there is still data that can be read. When SSL_MODE_AUTO_RETRY is set using [SSL_CTX_set_mode()](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_mode.html) OpenSSL will try to process the next record, and so not return SSL_ERROR_WANT_READ while it still has data available. Because many applications did not handle this properly, SSL_MODE_AUTO_RETRY has been made the default. If the application is using blocking sockets and SSL_MODE_AUTO_RETRY is enabled, and select() is used to check if a socket is readable this results in SSL_read() processing the non-application data records, but then try to read an application data record which might not be available and hang.
```

... howtf are you supposed to know that WANT_READ means retry vs waiting for more data that will never arrive?

https://www.openssl.org/docs/man3.3/man3/SSL_pending.html
```
If the SSL object's read_ahead flag is set (see [SSL_CTX_set_read_ahead(3)](https://www.openssl.org/docs/man3.3/man3/SSL_CTX_set_read_ahead.html)), additional protocol bytes (beyond the current record) may have been read containing more TLS/SSL records. This also applies to DTLS and pipelining (see [SSL_CTX_set_split_send_fragment(3)](https://www.openssl.org/docs/man3.3/man3/SSL_CTX_set_split_send_fragment.html)). These additional bytes will be buffered by OpenSSL but will remain unprocessed until they are needed. As these bytes are still in an unprocessed state SSL_pending() will ignore them. Therefore, it is possible for no more bytes to be readable from the underlying BIO (because OpenSSL has already read them) and for SSL_pending() to return 0, even though readable application data bytes are available (because the data is in unprocessed buffered records).

SSL_has_pending() returns 1 if s has buffered data (whether processed or unprocessed) and 0 otherwise. Note that it is possible for SSL_has_pending() to return 1, and then a subsequent call to SSL_read_ex() or SSL_read() to return no data because the unprocessed buffered data when processed yielded no application data (for example this can happen during renegotiation). It is also possible in this scenario for SSL_has_pending() to continue to return 1 even after an SSL_read_ex() or SSL_read() call because the buffered and unprocessed data is not yet processable (e.g. because OpenSSL has only received a partial record so far).
```

Can't even think of the use of these functions. You have to use both? I got a WANT_READ, do I have pending readable bytes? Try the read again? Oh still WANT_READ? Switch to the other pending?

I wonder if there're any complete examples of how to actually use OpenSSL. No application source I've run across has managed to catch these edge cases I see all over the manpages.

dormando:
think I'm going to add some extra counters for read/write failures and leave the branch for a while. The OpenSSL documentation says a lot about this or that happening but without any kind of complete example I can't _produce those situations_, so I can't figure out what's necessary or test the damn handler code. Based on the documentation a lot of the potential situations are simply impossible, and only cover theoretical scenarios that vary by TLS version or OpenSSL version.

I'll switch to working om mcshredder for a while. Need to refactor a few things then see if I can get TLS support across the finish line there. Which is also an incredible pain because uring and OpenSSL hate each other.

dormando:
Don't want to do the SSL_CTX change with this PR, unless I can get something definitive on it being necessary.

dormando:
Been working on getting TLS into mcshredder (which sucks, F). So I Can attempt to test this branch.

The _only_ potential weirdness I can find in TLS 1.3 are key refreshes. Once I get a benchmark up I can force those to happen and see how that affects the state machine.

dormando:
Small update: looks like you can force a TLS 1.3 key update. By default it happens every N million packets. I've got mcshredder mostly working with TLS now, so I can hack it to send key updates or just run tons of traffic and see what happens.

dormando:
still waiting to hear back from the user.

going to schedule time to self-test with mcshredder in august.

dormando:
Tested key updates by adding a line to `ssl_callback`:
`fprintf(stderr, "WHERE: %d RET: %d STR: %s\n", where, ret, SSL_state_string(s));` (also checked `SSL_state_string_long()`, though this didn't seem to match the short version!)

Ran half a billion requests through the same connection via mcshredder without issues. This would require some key updates, likely.

connected via `openssl s_client -connect localhost:11211` and checked the output . ran 'k'/'K' commands a few times to force TLS 1.3 key updates and ensure the server was handling that properly. Works fine; I see the callback updates as well.

next is to try harder to deliberately cause errors, though the test suite does this already.

dormando:
Doesn't seem like the test suite tries very hard to run TLS error conditions... might be forced to get this into the shredders suite.

Caused a few errors/disconnects/etc via `openssl s_client`. With `-verify_return_error` I can see the proto errors counter ticking properly.

dormando:
PITA. I don't really have a setup for testing this properly. Logically I think it's fine:

1) most of what I did was move code into `tls.c` for cleanliness reasons. The code didn't change.
2) where it did change, I added handling where there was previously none. Otherwise the logic is the same.
3) I opted out of refactoring the `SSL_CTX` reload until a later time.

Seems like it might be impossible to get a canary test from a user and I don't want this branch to rot much longer. I can add better testing via shredders but it's going to be many hours of work setting it up.

memcached.h
```diff
@@ -523,12 +521,12 @@ struct settings {
     double ext_max_frag; /* ideal maximum page fragmentation */
     double slab_automove_freeratio; /* % of memory to hold free as buffer */
     bool ext_drop_unread; /* skip unread items during compaction */
+    bool ssl_enabled; /* indicates whether SSL is enabled */
```
dormando, line 524: FIX: this got moved into the extstore ifdef

dormando:
some issues snuck in with tls not compiled in. all tests pass now.

PR #1162, opened 2024-08-04
Proxy API: req/res accessors and inspector object

dormando:
Adds interfaces for inspecting request and response objects without or with minimal allocations and argument checking overhead.

Simple extensions:
```lua
-- directly converts token at offset n into a lua integer. avoids string -> num round tripping
-- used for standard ascii commands (ie; set)
int = req:token_int(n)
```

```lua
-- returns "exists", then an int conversion of a token directly.
-- if the flag exists but the token could not be converted to a number, returns 'true, nil'
bool, int = req:flag_token_int("f")
```

```lua
-- Replaces the flag and associated token with blank spaces. This removes the flag without fully
-- rewriting the result.
res:flag_blank("O")
```

### req and res inspector objects

```lua
-- inspectors are pre-configured objects that can be applied to a request or result to make
-- one or more queries against said object.
-- This example splits a key by the "/" character and checks the second (pos = 2) token against the
-- supplied 'map' table. It then returns the result of that map.
-- If map is not supplied, it returns the token as a string
ins = mcp.req_inspector_new(
  { t = "sepkey", sep = "/", pos = 2, map = {
    foo = 1, bar = 2
  }}
)

-- later...
result = ins(req)

-- an inspector with multiple steps:
ins = mcp.res_inspector_new(
  { t = "hasflag", flag = "N" },
  { t = "hasflag", flag = "c" }
)

-- later...
has_N, has_c = ins(result)
```

### More complete example

```lua
local fgen = mcp.funcgen_new()
-- The inspector object pre-validates arguments from lua, so we can avoid some overhead at request time.
-- EX: we don't have to check that 'flag' exists, has a string attached to it, is a correct "flag" string, etc.
ins = mcp.res_inspector_new(
    { t = "hasflag", flag = "f" },
    { t = "hasflag", flag = "t" }
)

-- the inspector object is created _once_ for the entire function generator, and shared among all slots.
-- this saves some memory and improves CPU cache hit rate.
fgen:ready({
  f = function(rctx)
    return function(r)
        local res = rctx:etc()
        -- Function calls in lua have some fixed overhead. By having multi-step inspectors we gain speed by
        -- turning two function calls (and any runtime arg validation) into one.
        -- This style also lets us avoid more kinds of memory allocation.
        local has_f, has_t = ins(res)
        -- do logic
       return res
    end
  end
}
```

### Inspector features

```lua
-- split and tokenize request key. ie: `foo/bar/baz`, position 2. check against map
-- if no map supplied, returns the token as a string
{ t = "sepkey", sep = "/", pos = 2, map = {
    foo = 1, bar = 2, 
}},

-- check if key begins with `string`
{ t = "keybegin", str = "sepkey" }

-- check if key is `string`
-- in this example you can check if a key is one of three strings quickly in one call.
ins = mcp.req_inspector_new(
    { t = "keyis", str = "reqkey/one" },
    { t = "keyis", str = "reqkey/two" },
    { t = "keyis", str = "reqkey/three" }
)
is_one, is_two, is_three = ins(req)

-- check if meta flag exists (req or res)
{ t = "hasflag", flag = "f" }

-- get meta flag token as a string (if exists) (req or res)
{ t = "flagtoken", flag = "O" }

-- get meta flag as an integer (64bit signed, as per lua restrictions) (req or res)
{ t = "flagint", flag = "t" }

-- check if flag token matches `string`
{ t = "flagis", flag = "O", str = "baz" }
```

Planned:

- check if flag token starts or ends with `string`
- split tokenizer for flag tokens, same as request key above
- more...

### Future `mcp.(req|res)_mutator`

A similar multi-step pre-configured object is planned for mutating request and response objects.

Except instead of mutating the existing object they will be primarily aimed at making new objects from old objects. IE: existing request and response objects should be immutable if they already contain data.

API is planned but not implemented yet.

### TODO

- I have a planned mode of: `ins(req|res, output_table)` where instead of directly returning the results, they are put into the `output_table` as an array. I might delay this until later so I can get to the mutator sooner.

This would allow you to inspect something from a function and pass the results around a little easier. Users _should_ pre-allocate and re-use the output tables to avoid allocations where possible.

### DIFF
--- a/proxy_inspector.c
+++ b/proxy_inspector.c
@@ -1,0 +1,674 @@
+/* -*- Mode: C; tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+
+#include "proxy.h"
+
+enum mcp_ins_type {
+    INS_REQ = 1,
+    INS_RES,
+};
+
+enum mcp_ins_steptype {
+    mcp_ins_step_none = 0,
+    mcp_ins_step_sepkey,
+    mcp_ins_step_keybegin,
+    mcp_ins_step_keyis,
+    mcp_ins_step_hasflag,
+    mcp_ins_step_flagtoken,
+    mcp_ins_step_flagint,
+    mcp_ins_step_flagis,
+    mcp_ins_step_final, // not used.
+};
+
+// START STEP STRUCTS
+
+struct mcp_ins_sepkey {
+    char sep;
+    int pos;
+    int mapref;
+};
+
+struct mcp_ins_string {
+    unsigned int str; // arena offset for match string.
+    unsigned int len;
+};
+
+struct mcp_ins_flag {
+    uint64_t bit; // flag converted for bitmask test
+    char f;
+};
+
+// TODO: it might make more sense to flatten the structs into the ins_step
+// struct. It wouldn't take much more space if we can be careful with
+// alignment.
+struct mcp_ins_flagstr {
+    unsigned int str;
+    unsigned int len;
+    uint64_t bit; // flag bit
+    char f;
+};
+
+struct mcp_ins_step {
+    enum mcp_ins_steptype type;
+    union {
+        struct mcp_ins_sepkey sepkey;
+        struct mcp_ins_string string;
+        struct mcp_ins_flag flag;
+        struct mcp_ins_flagstr flagstr;
+    } c;
+};
+
+// END STEP STRUCTS
+
+struct mcp_inspector {
+    enum mcp_ins_type type;
+    int scount;
+    unsigned int aused; // arena memory used
+    unsigned int rcount; // number of results to expect
+    char *arena; // string/data storage for steps
+    struct mcp_ins_step steps[];
+};
+
+// PRIVATE INTERFACE
+
+// COMMON ARG HANDLERS
+
+// multiple step types only take 'flag' as an argument.
+static int mcp_inspector_flag_c_g(lua_State *L, int tidx) {
+    if (lua_getfield(L, tidx, "flag") != LUA_TNIL) {
+        size_t len = 0;
+        const char *flag = lua_tolstring(L, -1, &len);
+        if (len != 1) {
+            proxy_lua_ferror(L, "inspector step %d: 'flag' must be a single character", tidx);
+        }
+        if (mcp_is_flag_invalid(flag[0])) {
+            proxy_lua_ferror(L, "inspect step %d: 'flag' must be alphanumeric", tidx);
+        }
+    } else {
+        proxy_lua_ferror(L, "inspector step %d: must provide 'flag' argument", tidx);
+    }
+    lua_pop(L, 1); // val or nil
+    return 0;
+}
+
+static int mcp_inspector_flag_i_g(lua_State *L, int tidx, int sc, struct mcp_inspector *ins) {
+    struct mcp_ins_step *s = &ins->steps[sc];
+    struct mcp_ins_flag *c = &s->c.flag;
+
+    if (lua_getfield(L, tidx, "flag") != LUA_TNIL) {
+        const char *flag = lua_tostring(L, -1);
+        c->f = flag[0];
+        c->bit = (uint64_t)1 << (c->f - 65);
+    }
+    lua_pop(L, 1); // val or nil
+
+    return 0;
+}
+
+static int mcp_inspector_string_c_g(lua_State *L, int tidx) {
+    size_t len = 0;
+
+    if (lua_getfield(L, tidx, "str") != LUA_TNIL) {
+        lua_tolstring(L, -1, &len);
+        if (len < 1) {
+            proxy_lua_ferror(L, "inspector step %d: 'str' must have nonzero length", tidx);
+        }
+    } else {
+        proxy_lua_ferror(L, "inspector step %d: must provide 'str' argument", tidx);
+    }
+    lua_pop(L, 1); // val or nil
+
+    return len;
+}
+
+static int mcp_inspector_string_i_g(lua_State *L, int tidx, int sc, struct mcp_inspector *ins) {
+    struct mcp_ins_step *s = &ins->steps[sc];
+    struct mcp_ins_string *c = &s->c.string;
+    size_t len = 0;
+
+    // store our match string in the arena space that we reserved before.
+    if (lua_getfield(L, tidx, "str") != LUA_TNIL) {
+        const char *str = lua_tolstring(L, -1, &len);
+        c->str = ins->aused;
+        c->len = len;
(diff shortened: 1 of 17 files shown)


dormando:
great.... `res:line()` is still screwed up.. and looks like r->key is also screwed up, but thankfully we don't seem to reference it anywhere in the code... also since `res:line()` was broken until this patch series nobody should've been using it either.

these should be offsets, not pointers. I parse the response structure out of the receive buffer, then we copy the response line into a dedicated buffer for the response object... thus the pointers now point back to the wrong place.

dormando:
spent a while auditing things...

think I'm going to take a conservative approach so I can fix this upstream with a minimal change:

**the fault:**
- We read data from a backend socket into a buffer. We parse the result directly out of that buffer. The bug is that some `char*` pointers in the response struct point directly into the backend socket read buffer... but we then memcpy the response into its own buffer associated with the object. So now the `char*` pointers are _probably_ correct but not always.
- This has been fine because we haven't used those pointers until recently (`mcp.match` and `res:line()`, which was broken for other reasons)

**approach**
- If any of the busted accessors are about to be used, re-parse the results off of `r->buf`, which will zero the struct and re-fill the pointers.
- Since we only use this in 1-2 places this is a minimally invasive change that I can immediately upstream.
- After this patch series, I will change the mcmc resp struct to be offset based + macro accessors instead of direct pointers. This should have a benefit of shrinking the size of the struct anyway...

dormando:
addressed a bunch of cleanups for the proxy code... need to fill out some more tests too.

also looks like I do need to revisit the mcmc branch and clean up/finish some stuff. some flag int accessors are missing.

going to have to squash the branch and then split it into the mcmc update and proxy update separately.

dormando:
Mostly cleaned up... going to audit the mcmc tests now.

Then decide on how much more work is necessary vs continuing it when I do the mutator.

At this point adding more types of steps has a pattern so I can add more as-needed in individual PR's/commits. I have a bunch of ideas but there's no reason they have to be done right this second.

dormando:
think it's time to squash this booger and split it into intentional commits.

need to decide if I care for history/bisection sake to separate the mcmc update from the ins stuff.

dormando:
squashed and separated.

will let this bake for a day or two before merging.. maybe see if I can do an easy shredders burn-in test with it.

dormando:
final self-review and then merge-down I think.

proxy_lua.c
```diff
@@ -1795,6 +1684,12 @@ int proxy_register_libs(void *ctx, LIBEVENT_THREAD *t, void *state) {
         {NULL, NULL}
     };
 
+    const struct luaL_Reg mcplib_inspector_m[] = {
```
dormando, line 1687: maybe make a reqinspector/resinspector as different meta tables? I don't think this helps much though.

proxy_request.c
```diff
@@ -1015,14 +1104,14 @@ int mcplib_request_match_res(lua_State *L) {
...
     // requests all have keys. check for an opaque.
     mcp_request_find_flag_token(rq, 'O', &opaque_token, &opaque_len);
-    mcmc_bare_parse_buf(rs->buf, rs->blen, &reresp);
+    mcmc_parse_buf(rs->buf, rs->blen, &reresp);
 
     // scan the response line for tokens, since we don't have a reciprocal API
     // yet. When we do this code will be replaced with a function call like
     // the above.
     const char *p = reresp.rline;
     // TODO: Think this is an off-by-one in mcmc.
```
dormando, line 1113: delete commenet

t/proxyins.lua
```diff
@@ -0,0 +1,141 @@
...
+                    local has_O, O, has_t, t = mgresflaga_ins(res)
+                    return string.format("SERVER_ERROR O[%q]: %s t[%q]: %q\r\n",
+                        has_O, O, has_t, t)
+                elseif key == "reshasf/reqhasf" then
+                    local has_O, O, has_T, T = mgreqflaga_ins(r)
+                    return string.format("SERVER_ERROR O[%q]: %s T[%q]: %q\r\n",
+                        has_O, O, has_T, T)
+                else
+                    local f, t = mgreshasf_ins(res)
+                    return "SERVER_ERROR f: " .. tostring(f) .. " t: " .. tostring(t) .. "\r\n"
```
dormando, line 110: change to string.format

vendor/mcmc/mcmc.c
```diff
@@ -358,32 +459,272 @@ static int _mcmc_parse_response(mcmc_ctx_t *ctx, mcmc_resp_t *r) {
     }
 }
 
-// EXTERNAL API
+// TODO: possible to codegen the 32 vs 64 variants
```
dormando, line 462: need to at least add a note here: not defending against len == 0 in these functions.

the while loop won't do anything but the minus check can go past if the len is 0. There's no practical case where a token can exist and +1'ing would go past the end of the buffer though: we _must_ have a `\n` at the end of the string.

So I should probably add this as a comment and see about adjusting later.

vendor/mcmc/mcmc.c
```diff
@@ -358,32 +459,272 @@ static int _mcmc_parse_response(mcmc_ctx_t *ctx, mcmc_resp_t *r) {
...
+    }
+
+X(mcmc_token_get_u32, uint32_t, mcmc_toktou32)
+X(mcmc_token_get_u64, uint64_t, mcmc_toktou64)
+X(mcmc_token_get_32, int32_t, mcmc_tokto32)
+X(mcmc_token_get_64, int64_t, mcmc_tokto64)
+
+#undef X
+
+int mcmc_token_has_flag(const char *l, mcmc_tokenizer_t *t, char flag) {
```
dormando, line 636: I've forgotten the weird rules around signed vs unsigned chars and what platforms have what. might be worth doing this differently, but probably not now.

proxy_inspector.c
```diff
@@ -0,0 +1,674 @@
...
+    mcp_ins_step_keybegin,
+    mcp_ins_step_keyis,
+    mcp_ins_step_hasflag,
+    mcp_ins_step_flagtoken,
+    mcp_ins_step_flagint,
+    mcp_ins_step_flagis,
+    mcp_ins_step_final, // not used.
+};
+
+// START  STEP STRUCTS
```
dormando, line 22: two spaces

proxy_lua.c
dormando: meh maybe later. should do an actual asm comparison + branch miss count to see if it does anything. odds are pretty good it wouldn't.

dormando:
confident this is the correct approach, so merging and moving on.

any further steps will be separate PR's

dormando:
Also: need to move info to the wiki docs once released.

PR #1165, opened 2024-08-13
Proxy req/res mutator object

dormando:
Usage is similar to inspector object in #1162

`status = mut(destination, src, src, src, src)`
IE: based on configured steps, takes arguments for assembling the new request or response string
`status = mut(new_res, "HD", 33, another_result)`

On the "config" side an argument is referred to by an `idx` parameter. IE: in the above example "HD" would be in `idx = 2`, as the first user parameter. 33 would be 3, etc. Multiple steps can refer to the same idx. ex: pulling a key and some flags from the same request object in idx 2.

Some steps can also refer to multiple input arguments. IE: take a range of arguments and combine them.

Complete basic example:
```lua
    -- basic; no flags or outside info
    local mut_mgreq = mcp.req_mutator_new(
        { t = "cmdset", cmd = "mg" },
        { t = "keyset", str = "override" }
    )

    mgfg:ready({
        n = "mgtest", f = function(rctx)
            -- make blank request objects for handing to mutator

            -- these objects must be made per slot (rctx)
            -- they're made via the rctx so it can release memory inbetween
            -- requests.
            local nreq = rctx:request_new()
            local nres = rctx:response_new()

            return function(r)
                    -- creates a static request of "mg override\r\n" to use instead of 'r'
                    local ret = mut_mgreq(nreq)
                    -- if ret == false etc
                    return rctx:enqueue_and_wait(nreq, mgfgh)
            end
        end
    })
```

### Implemented steps

Note: Some of these are partially implemented as of this writing. ie; not all types of arguments accepted.

```lua
  -- manually set the command
  { t = "cmdset", cmd = "ma" },

  -- inherit the command
  { t = "cmdcopy", idx = 1 },

  -- copy the key wholesale from req or res
  -- if res, must be meta and have 'k' flag
  { t = "keycopy", idx = 1 },

  -- set key to exact string
  { t = "keyset", str = "foo" },

  -- set result to code. maybe int instead?
  { t = "rescodeset", str = "code" },

  -- copy res code from another res
  { t = "rescodecopy", idx = 1 },

  -- set "SERVER_ERROR|CLIENT_ERROR|ERROR msg\r\n"
  { t = "reserr", mode = "server|client|error", msg = "blah blah" },\

  -- set flag. sets token if val supplied
  { t = "flagset", flag = "f", val = 1234},

  -- copy flag from idx
  { t = "flagcopy", flag = "f", idx = 1},
```

### Steps TODO

Note: not all steps will be completed in this PR. 'o' in front of the `{` means optional but planned

```lua
  -- format string version of reserr
  -- runs once against the test array table when making the object
  { t = "resfmterr", mode = "server|client|error", msg = "blah %d blah %s blah",
    test = { 5, "foo" } },

  -- copy key but trim until char is found
  -- also add "count" to trim specifically n chars?
  o { t = "keycopytrim", idx = 1, until = "/" },

  -- assemble the key from multiple args with optional sep
  o { t = "keycombine", sidx = 1, eidx = 2, sep = "/" },

  -- copy flag/token from one flag to another
  { t = "flagcopyfrom", dflag = "F", sflag = "f", idx = 1},

  -- copy all flags + tokens from idx to dst
  -- unless specified in 'except'
  -- ie: copy everything but O, replace with our own O
  o { t = "flagcopyall", idx = 1, except = "abcd"},

  -- for copying ascii command arguments
  { t = "tokencopy", tok = 2, idx = 1},

  -- for converting commands?
  { t = "tokentoflag", tok = 2, flag = "f", idx = 1},

  -- for ascii commands.
  -- add shortcut strings for set-ttl or something?
  -- name -> tok
  { t = "tokenset", tok = 2, val = 1234},

  -- inherit the value from idx 1
  { t = "valcopy", idx = 1},

  -- directly set the value here
  { t = "valset", val = num|string },

  -- set the value from strings or objs in args 1 through 2
  -- with optional separator
  o { t = "valcombine", sidx = 1, eidx = 2, sep = "|" },

  -- walk an array table and combine into the value
  -- with optional separator
  o { t = "valtcombine", tidx = 1, sep = "|"},

  -- TODO: val splitter? maybe that's something else
  -- for taking a req and splitting it into N things
```

### Further TODO

There are lots of internal TODO/FIXME's for fleshing this out. Will copy this to a new PR later.

- [ ] inherit starttime/elapsed time when creating requests/responses, else logging is busted. Maybe allow overriding via a step?
- [ ] track req/res objects created by an rctx slot so temporary buffers can be cleaned up when the request completes, instead of when the next request starts.
- [ ] flagcopy needs an 'arg' to take string/numbers/req/res as input instead of just "same as the dst" as now.
- [ ] some better consistency between when steps take 'str' vs 'val' vs 'token' vs etc.
- [ ] `idx` args should probably start at 1 instead of 2 to prevent user confusion? "first argument" is technically arg 3 internally to lua: `mut(dst, arg)` -> arg 1 is the mut object (self), 2 is dst, 3 is first user arg.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -62,6 +62,7 @@ memcached_SOURCES += proto_proxy.c proto_proxy.h vendor/mcmc/mcmc.h \
 					 proxy_ratelim.c \
 					 proxy_jump_hash.c proxy_request.c \
 					 proxy_result.c proxy_inspector.c \
+					 proxy_mutator.c \
 					 proxy_network.c proxy_lua.c \
 					 proxy_luafgen.c \
 					 proxy_config.c proxy_ring_hash.c \
--- a/proxy.h
+++ b/proxy.h
@@ -790,6 +790,8 @@ int mcplib_rcontext_res_ok(lua_State *L);
 int mcplib_rcontext_result(lua_State *L);
 int mcplib_rcontext_cfd(lua_State *L);
 int mcplib_rcontext_tls_peer_cn(lua_State *L);
+int mcplib_rcontext_request_new(lua_State *L);
+int mcplib_rcontext_response_new(lua_State *L);
 int mcplib_rcontext_sleep(lua_State *L);
 int mcplib_funcgenbare_new(lua_State *L);
 int mcplib_funcgen_new(lua_State *L);
@@ -841,6 +843,12 @@ int mcplib_req_inspector_new(lua_State *L);
 int mcplib_res_inspector_new(lua_State *L);
 int mcplib_inspector_gc(lua_State *L);
 int mcplib_inspector_call(lua_State *L);
+
+// mutator interface
+int mcplib_req_mutator_new(lua_State *L);
+int mcplib_res_mutator_new(lua_State *L);
+int mcplib_mutator_gc(lua_State *L);
+int mcplib_mutator_call(lua_State *L);
 
 void mcp_response_cleanup(LIBEVENT_THREAD *t, mcp_resp_t *r);
 void mcp_set_resobj(mcp_resp_t *r, mcp_request_t *rq, mcp_backend_t *be, LIBEVENT_THREAD *t);
--- a/proxy_lua.c
+++ b/proxy_lua.c
@@ -1674,6 +1674,8 @@ int proxy_register_libs(void *ctx, LIBEVENT_THREAD *t, void *state) {
         {"result", mcplib_rcontext_result},
         {"cfd", mcplib_rcontext_cfd},
         {"tls_peer_cn", mcplib_rcontext_tls_peer_cn},
+        {"request_new", mcplib_rcontext_request_new},
+        {"response_new", mcplib_rcontext_response_new},
         //{"sleep", mcplib_rcontext_sleep}, see comments on function
         {NULL, NULL}
     };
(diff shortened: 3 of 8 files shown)


dormando:
Working through how to handle values now. It's a little tricky:

- Have to be careful of security issues when specifying a value that doesn't match the header length (ensure we send what's in the value, or enforce syncing of the ms/set size number)
- result objects have been a single buffer since they're read sequentially off of the network. now we probably need it to have a separate value buffer.

After this it's mostly simple cleanup stuff.

dormando:
miiiight have to limit the "set a value to a result" scope to a fixed buffer size or limit for now the scope of this PR, then do a separate PR to focus refactoring of the result object.

It's not looking super complex but I would feel better having it be separate.

dormando:
Just lost the better part of a week to being sick. Going to cut the scope for this PR:

- Only support creation of new request objects.
- Follow up next month with support for result objects, possibly after a data structure refactor PR.

dormando:
Changed my mind again:

- Result objects should work (just finished a big refactor here)
- Going to do some basic testing and some cleanups.
- Will merge _without_ core changes necessary for efficient memory usage, marking the mutator as experimental/flawed.

Will do extended testing/integration and patch it in a future PR, along with the necessary core rctx changes.

What this means:

- mutator code will be 100% inert if not executed.
- request/result objects used by the mutator will not have their memory freed until the rctx slot itself is freed, instead of when the request is done processing.

To mark this as "not experimental" that optimization _must_ be done, but it requires changing some core code to ensure req/res objects created by an `rctx` are cleaned up inbetween requests. Doing this adds some risk which isn't worth it while the interface is experimental.

dormando:
I'm super low on time for the month so trying to kick this across the line.

I _could_ and probably normally _should_ keep the PR open, but it's going to have a bad merge conflict with the V1 removal.

So I'm taking this as a special case and merging it since it's safe (inert code) partially works (can at least test things), and there's a special situation (I need to reserve time to clear out v1 this month)

dormando:
As noted above I've taken the (rare) step of merging this before it's complete, as it can't break anything and will save time later by not having to deal with merge conflicts.

I'm estimating another 1-3 days at most for completing the work, which will likely be next month.

PR #1169, opened 2024-09-05
Fetch CRC32c of request value

dormando:
For debugging/informational purposes, maybe.


dormando:
guess I'll close this out. nobody ended up wanting it; if I do it I might implement it directly into the protocol.

PR #1176, opened 2024-10-04
Create Perl tests of inspector behavior when mg returns a value

FinnitoProductions:
The response inspector flag parsing logic behaves incorrectly when a meta-get response includes a value. 

Introduce tests of this behavior for both standard requests and `mcp.internal` requests. These tests currently fail, but they are expected to pass once the bug is patched.

### DIFF
--- a/t/proxyins.t
+++ b/t/proxyins.t
@@ -31,8 +31,8 @@ is(<$w>, "OK\r\n");
 
 {
     test_mgintres();
-    #test_mgreq();
-    #test_mgres();
+    test_mgreq();
+    test_mgres();
 }
 
 sub test_mgintres {
@@ -44,6 +44,12 @@ sub test_mgintres {
 
     subtest 'flagtoken and flagint' => sub {
         $t->c_send("mg intres/tokenint f t s Omoo\r\n");
+        $t->c_recv("SERVER_ERROR O[true]: moo t[true]: -1\r\n");
+        $t->clear();
+    };
+
+    subtest 'flagtoken and flagint with value returned' => sub {
+        $t->c_send("mg intres/tokenint v f t s Omoo\r\n");
         $t->c_recv("SERVER_ERROR O[true]: moo t[true]: -1\r\n");
         $t->clear();
     };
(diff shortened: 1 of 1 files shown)


dormando:
Noting for myself:
- mcmc tokenizer is being passed an incorrect mstart parameter for VA responses (is 1, should be 2), so it ends up looking at the second argument as though it's a flag and fails validation.
- from the code this looks like a pretty deep mistake, since we have to look at the result a bit before knowing where mstart is. it's unit tested but I did the mstart correctly since that was the lowest layer... sigh.

not immediately obvious how to fix simply so I'll come back to this after a break. forget where I put the HD/VA/etc code

dormando:
More notes for me:

- can we call `mcmc_tokenize_res` against `r->resp.rline` instead since that will always point to the right place?
- if not, will have to have the parser remember the metastart offset instead of passing it into `mcmc_tokenize_res`

object vlen of 0 still has an internal vlen of 2, so I can just check for vlen != 0 to see where mstart is.

dormando:
It's not a complicated fix but I need to make some API decisions. will get back to you.

dormando:
Simplest method here might be:
- change mstart arg from offset to boolean
- if flag parsing fails, check if character is numeric and skip if so

that should be it. no need to check for HD vs VA anywhere up the chain.

dormando:
The biggest issue here is:

- We tend to parse things directly out of input buffers.
- We then sometimes memcpy that actual request/response data into other buffers
- Now there are pointers in the `mcmc_resp_t` struct that point into the original buffer
- Thus we need to re-parse things sometimes. Which I'd prefer to not do.
- This matters because I can't just parse the meta line by looking at the `rline` offset without first re-parsing the thing.

I need to change the various pointers in `mcmc_resp_t` into offsets... been wanting to do this for a long time but it's a little more involved so I keep pushing it off.

TL;DR: I'm going to do the workaround of _not_ throwing an error if we run into numbers, and see about raising the priority of doing a large rework of mcmc.

dormando:
fix should be merged now. Your tests + some additional ones I added all pass. Thanks!

PR #1177, opened 2024-10-08
Adding unit tests for watch deletions when calling mcp.internal()

YufeiHu:
The `watch deletions` command is not printing the deletion logs correctly when sending deletion requests through `mcp.internal()`. The attached unit test file illustrates this issue.


dormando:
Thanks!

dormando:
@YufeiHu 

Thanks! Any chance you could integrate this test into `t/proxyinternal.t` instead of having it be its own file? If not I'll do it; but just offering in case you feel up to contributing.

That test is listening on CMD_ANY so it should be pretty trivial to add the watcher bits as a subtest

YufeiHu:
@dormando 

I've addressed the suggestion! The test now passes after I applied the same fix as in your branch (https://github.com/memcached/memcached/pull/1178).

If there are any further adjustments needed, I’ll take care of them promptly. Thanks!

dormando:
Looks rad, thanks!

I need to port more of these test files to the object based one though, sorry!

dormando:
merged

PR #1182, opened 2024-10-23
proxy: embed routelib

dormando:
`-o proxy_config=routelib` will use the binary-distributed routelib library

Not thrilled about including a mystery data file in the vendor directory.
Don't want to make `xxd` part of the build requirements for proxy either?

Maybe the plan is:
- Start with this so we have something.
- Include the routelib.lua file in here too.
- Soon write my own minimal bin2c that lives in the tree somewhere
- Remove the distributed .h and somehow get the build system to build and run the bin2c first
- Have the .h as a build artifact.

You can just ld files but it's not portable either.

TODO:
- [x] Minimal perl test.
- [x] Test reloads?
- [x] help output?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -68,7 +68,8 @@ memcached_SOURCES += proto_proxy.c proto_proxy.h vendor/mcmc/mcmc.h \
 					 proxy_config.c proxy_ring_hash.c \
 					 proxy_internal.c \
 					 proxy_tls.c proxy_tls.h \
-					 md5.c md5.h
+					 md5.c md5.h \
+					 vendor/routelib/routelib.h
 endif
 
 if ENABLE_EXTSTORE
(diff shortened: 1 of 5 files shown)


dormando:
This is probably shippable... soon...

Maybe one more change to put the above aspirations into the README file?

dormando:
Maybe also include the rlib git sha in there?

dormando:
Another annoying thing: if you use the embedded routelib but want to load a non-embedded one later, you're SOL without a full restart? Maybe we don't care and/or solve this later?

dormando:
If I complete/merge/release this before the talk tomorrow I'll make it in time... lol...

dormando:
mmmmmmmmmmmmmmmmmmmm think it'll be worth using the external vendor tarball for this. the routelib is pretty huge and prone to churn.

once I get the bin2c stuff settled and the pre-built .h file removed I'd be less sad about putting routelib.lua directly into here.

dormando:
moved routelib files to the external vendor repo... which I need to write a maintenance script for. sigh. first update since I made the repo in 2021!

dormando:
think I'm happy with this for now.

PR #1183, opened 2024-10-29
Proxy mutator object continuation

dormando:
The goal of this PR is to stabilize the mutator object for production use. Additional steps will most likely be in future PR's.

Tasks:

- [x] Fix error handling for totalling. Don't allow errors in copy phase.
- [x] Replace assert()'s with error handling
- [x] Fix dynamic argument handling for valcopy/flagcopy/etc (take a string or a req or a res or etc)
- [x] Track req/res objects and free their data when the rctx is freed, instead of at the next use
- [x] user idx of args should start at 1 insead of 2.
- [x] check the naming consistency of arguments: str vs arg vs token vs etc
- [x] support value handling for result objects (might be later)
- [ ] ~~inherit starttime/elapsed from copied objects (is this an explicit step?)~~
- [x] address/evaluate remaining TODO/FIXME's.
- [x] other keycopy/rescodecopy/cmdcopy need code from valcopy/flagcopy

Result handling might have to wait for another PR, pending fixes to mcmc to use relative offsets vs full fixed pointers. This is causing us to need to re-parse results every time we look at them.

Delayed:
- [ ] Support copying values off of source result objects (requires mcmc fix or hack)
- [ ] don't allow specifying steps out of order (cmd -> key -> flags -> etc)

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -653,7 +653,7 @@ struct mcp_funcgen_s {
     int self_ref; // self-reference if we're attached anywhere
     int argument_ref; // reference to an argument to pass to generator
     int max_queues; // how many queue slots rctx's have
-    int extra_queues; // how many extra queue slots for object storage we have
+    int uobj_queues; // how many extra queue slots for object storage we have
     unsigned int refcount; // reference counter
     unsigned int total; // total contexts managed
     unsigned int free; // free contexts
@@ -748,7 +748,7 @@ struct mcp_rcontext_s {
     int conn_fd; // fd of the originating client, as *c can become invalid
     enum mcp_rqueue_e wait_mode;
     uint8_t lua_narg; // number of responses to push when yield resuming.
-    uint8_t obj_count; // number of extra tracked req/res objects.
+    uint8_t uobj_count; // number of extra tracked req/res objects.
     bool first_queue; // HACK
     lua_State *Lc; // coroutine thread pointer.
     mcp_request_t *request; // ptr to the above reference.
--- a/proxy.h
+++ b/proxy.h
@@ -653,6 +653,7 @@ struct mcp_funcgen_s {
     int self_ref; // self-reference if we're attached anywhere
     int argument_ref; // reference to an argument to pass to generator
     int max_queues; // how many queue slots rctx's have
+    int extra_queues; // how many extra queue slots for object storage we have
     unsigned int refcount; // reference counter
     unsigned int total; // total contexts managed
     unsigned int free; // free contexts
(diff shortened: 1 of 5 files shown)


dormando:
Been chipping at methods to do the req/res object tracking with minimal code impact. I try for a while, then go to another TODO then etc.

It's easy to do but it adds overhead to every request if done the easy way.

dormando:
Took way too long for me to come to that conclusion for tracking the req/res objects :/ really don't like adding more fixed overhead to the system. When we do loop the objects to run their cleanup routines the memory is at least sequential so it shouldn't impact much.

dormando:
Doing a TODO/FIXME scrub and maybe delaying the rest of the TODO items.

Once delayed I should:
- Write more unit tests for an hour or two.
- Merge with the "experimental" tag softened a bit.
- Open an issue to publicly track the mutator's API supported steps + wishlist steps.

dormando:
Notes for timing inheritance:
- The result's timing is started when we _enqueue_ a request to a backend
- There's no timing on the _request_ or _rcontext_ objects. otherwise sequential fetches would have longer timings
- "elapsed time" is only stamped when we read an object off the network

So if we create a result object from the mutator we don't really have a start or elapsed time. I think in a lot of cases that's going to be fine: if we're making a res from a res, we probably already logged when the original result was fetched from a pool. If not the elapsed time will be zero.

We can add an explicit step to inherit elapsed time into a new result object.

proxy.h
```diff
@@ -653,6 +653,7 @@ struct mcp_funcgen_s {
     int self_ref; // self-reference if we're attached anywhere
     int argument_ref; // reference to an argument to pass to generator
     int max_queues; // how many queue slots rctx's have
+    int extra_queues; // how many extra queue slots for object storage we have
```
dormando, line 656: rename to uobj_queues for consistency?

proxy.h
```diff
@@ -744,6 +748,7 @@ struct mcp_rcontext_s {
     int conn_fd; // fd of the originating client, as *c can become invalid
     enum mcp_rqueue_e wait_mode;
     uint8_t lua_narg; // number of responses to push when yield resuming.
+    uint8_t obj_count; // number of extra tracked req/res objects.
```
dormando, line 751: uobj_count

proxy_luafgen.c
```diff
@@ -1447,22 +1489,49 @@ int mcplib_rcontext_tls_peer_cn(lua_State *L) {
...
+        proxy_lua_error(L, "rctx request new: object count limit reached");
+        return 0;
+    }
+
     mcp_resp_t *r = lua_newuserdatauv(L, sizeof(mcp_resp_t), 0);
     memset(r, 0, sizeof(mcp_resp_t));
     luaL_getmetatable(L, "mcp.response");
     lua_setmetatable(L, -2);
+
+    lua_pushvalue(L, -1); // dupe rq for the rqueue slot
```
dormando, line 1528: function call

PR #1185, opened 2024-11-08
Do conn_release_items when a TRANSMIT_HARD_ERROR occurs in the UDP protocol

ing-eoking:
In the UDP protocol, when an error other than EAGAIN or EWOULDBLOCK occurs during a transmit operation, it is switched to conn_read without any additional handling. I think this could lead to a situation where IO wraps and any half-uploaded items are not freed.


dormando:
Thanks! I'll audit this and merge but it could take a while. If you can somehow add tests to prove the before/after that would speed up the process.

PR #1187, opened 2024-11-16
Slab mover and automover improvements

dormando:
Addresses numerous issues with page moving:
- will sometimes evict random memory while moving pages when it doesn't strictly need to
- evicts mostly random memory when it does need to, instead of evicting the LRU tail
- page mover code is confusing and embedded inside main slabs code.
- automover algorithm runs from LRU maintainer thread, uncoordinated with page mover thread
- automover algorithm can be overly aggressive

Improvements so far:
- automover algo and page mover are now their own .c file.
- slabber code no longer deeply intertwined with item/extstore code
- fewer branches and lock calls for small performance improvements
- new thread does not rely on global variables
- automover algo coordinates directly with page mover thread, to avoid spamming move calls or attempting to make decisions while the page mover is still busy
- `evictions_nomem` case is completely removed
- if it needs memory to rescue an item from a page being moved, it is pulled from free memory.
- if there is no free memory, at all (global page pool is 0), it will evict from the LRU tail.
- locking code around rebalancer thread has been isolated to just its thread and cleaned up
- automover algorithm re-instantiates by a global version counter, allowing easier addition of runtime tunables
- extstore's global freeratio is now runtime tunable
- extstore's automover algorithm is now less likely to attempt to move pages when there isn't enough free memory in the class.
- page mover "busy backoff" sleep was never being reset, could get stuck at 1ms after encountering busy items, so all future busy items cause it to sleep for 1ms.

Does not fix:
- if a chunked item (> 512kb on default configurations) is "Busy" (referenced, being sent to network, etc) for several seconds we need to delete the item to make progress. This used to be true for any type of busy item, but now we can at least rescue non-chunked items.

Also:
- This again removes `slabs automove =2` mode where random memory is thrown into a class receiving an eviction. I _might_ put it back again before release but I really don't like it.
___

### PROBLEM:
- sometimes we _want_ to evict memory to move a page: when memory is full
  everywhere.
- sometimes we _don't_ want to evict memory to move a page: pages are available in global pool
### QUESTIONS:
- is the only true metric for evict-to-move "is the global page pool 0"?
  - if not zero, but below threshold and high slab class, storage thread _should_ be flushing
    - thus busy-wait and keep retrying until memory is available.
  - if zero and pulling from low class (to either global or another class)
    evictions are necessary
  - if zero and NOT extstore all memory is used, so we should be evicting
    tail to make space.
  - If user manually asked to move memory: always evict if necessary.

___

TODO for this PR:
- [x] more `sysevents` logging
- [x] re-add page mover busy loop limiter.
- [x] add ability to rescue "hot" items by swapping them out in the hash table as if a `set` happened
- [x] if a page move was manually requested, allow evicting from LRU tail even if global pages != 0
- [x] limit per-slab-class free memory targets from extstore flush thread: avoids accidentally emptying some slab classes
- [x] complete updates to stats counters (no more nomem, some new busy/rescue counters) + doc/protocol.txt
- [x] live tuning of automove sliding window
- [x] "extstore memory pressure" stat via automove module (note to self: use stats-state struct)
- [ ] live-tuning min-pages-for-reclaim from algo; ~~set default to 1meg per 1-2 gigs of total ram~~ (need other fixes first)
- [x] maybe: combine the three different "min pages" const's in the extstore automove algo into the one tunable from above
- [ ] maybe: attempt C unit tests (requires further work on slabber code)
- [x] re-evaluate / rewrite perl integration/unit tests. add special command to intentionally leak refs so we can test busy scenarios.
- [ ] maybe: more perl tests specifically for automove scenarios (though this is difficult; might just do bench burn-ins)
- [ ] more benchmark burn-in tests. already started (and fixed some slowness). load/reload/change of size/etc

TODO for after this PR:
- [ ] support rescuing busy chunked items by swapping out just the chunk (details to come later)
- [ ] refactor slabber's per-slab "page list" to be a single block of memory split into a linked list, instead of one array per class which needs to be constantly checked and resized. saves a couple megabytes on big servers and removes error handling code paths in slabber and page mover.
- [ ] remove storage flush thread by enabling other "tail pulling" methods (normal alloc, lru maintainer, lru crawler) to flush directly to extstore. full details later. will fix issues from lag between needing to flush and flushing, and needing to over-flush in order to move page memory around.
- [ ] refactor slabber to remove use of globals (enables unit testing, future features)
- [ ] ability to abort a page move if busy for too long (ie; locked or uploading items). requires a refactor in the slab page list management so we can reorder the list of pages and try again.


dormando:
tests all pass now, whoo.

dormando:
Rescuing busy chunked items will require too much churn for this PR so I'll do it afterward. For this one we will support rescuing busy normal non-chunked items.

dormando:
May have fixed a bug with double-freeing a chunked item that was found as expired.

dormando:
Thinking of adding a debug protocol command to intentionally change the refcount of an item. Then the perl unit tests could test the "high refcount chunk move" paths where it currently just doesn't seem to.

dormando:
nearing feature complete... then need to re-evaluate perl tests and get some mcshredder burn-in tests going. important to test chunked with non-chunked since there appears to have been some chunked bugs.

dormando:
Need to think about more/maybe model:
- I made it less likely to move memory around within 30s window of a class getting an eviction or adding memory.
- really bursty classes may screw this up. ie; it could be better to toss memory back quickly after a big burst in case another burst happens somewhere else a few seconds later.

with the page mover busy-waiting maybe it can be slightly more aggressive. maybe I can test this with perl tests?

dormando:
added more TODO items after back-burnering for a few days.

it's difficult to express how much "free memory" an extstore instance actually has. I may have some luck expressing it as a "pressure %" where 100% means "oops we're close to evicting" and 0% means "instance is empty". VS expressing it as a literal number this would let me swap the internals of the calculation when things change in a few months.

dormando:
note for extstore memory pressure:

might want to remove the "free_chunks" worth of pages from the calculation... but "Excess free memory" will end up back in global? it might be fine?

dormando:
okay now I just need to do some focused testing for a while. unit testing and some benchmark scenarios.

dormando:
note mostly to self: take a break from the other file and start the extstore automove test file. might be a crash in there.

dormando:
maybe: for automover instead of selecting large class by oldest-age, sort by # of free bytes?

would be nice to _attempt_ to rotate through the classes as well. that's the main problem I was trying to solve by flushing extra anyway.

dormando:
Should probably just table any further tuning of the flush thread and put that effort into removing the flush thread...

As noted above, instead of trying to coordinate page moves with a flush thread that's randomly sleeping the page mover could just flush them directly as needed. Also the LRU maintainer could begin flushing from COLD if global pool is low and LRU crawler could potentially flush items.

dormando:
think all I want to do pending release (at least critically) is ensure my burn-in test suite has baked it for a while.

PR #1188, opened 2024-11-25
Changing the proxy GC default ratio from 1.9 to 2.0

YufeiHu:
Updating the proxy GC default ratio from `1.9` to `2.0` to reduce its aggressiveness. This adjustment aligns with the GC algorithm fix introduced in [commit d164eb1](https://github.com/memcached/memcached/commit/d164eb151af58f3835b63a83bb38e8d42892e669) and has been verified to perform well during tests.


dormando:
merged to next, thanks!

PR #1192, opened 2024-12-27
proxy: internal backend for V2 API

dormando:
Replaces `res = mcp.internal(req)` with the standard V2 API flow. Create a handle for an fgen using `mcp.internal_backend` as an argument, then call wait against it like a normal pool.

- [x] rebase on `io_refactor` branch and complete code
- [x] write tests for memory and extstore calls via a subrctx
- [x] complete missing error handling code
- [x] performance audit. seems to still be causing GC runs? why?
- [x] code audit? what else is missing?
- [x] burn-in suite run (started)
- [x] triple check that refcounts don't leak for results that weren't returned to the user.
- [ ] rename `mcp.internal_backend` to `mcp.internal_pool` (update in shredders suite too)

Terrible thought: am I naming this wrong?

`mcp.internal_backend` -> `mcp.internal_pool`

because you can't actually add it as a pool backend. it _could_ be nice if that worked but I think it would get people into more trouble than it's worth.

### DIFF
--- a/proxy.h
+++ b/proxy.h
@@ -521,6 +521,7 @@ struct _io_pending_proxy_t {
     // original struct ends here
 
     mcp_rcontext_t *rctx; // pointer to request context.
+    mcp_resp_t *client_resp; // reference (currently pointing to a lua object)
     int queue_handle; // queue slot to return this result to
     bool ascii_multiget; // passed on from mcp_r_t
     union {
@@ -541,7 +542,6 @@ struct _io_pending_proxy_t {
             struct iovec iov[2]; // request string + tail buffer
             int iovcnt; // 1 or 2...
             unsigned int iovbytes; // total bytes in the iovec
-            mcp_resp_t *client_resp; // reference (currently pointing to a lua object)
             bool flushed; // whether we've fully written this request to a backend.
             bool background; // dummy IO for backgrounded awaits
         };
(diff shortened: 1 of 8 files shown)


dormando:
I'm impressed this started working already, though it probably leaks memory and has tons of unhandled error cases.

dormando:
think doing this with extstore is blocked on refactoring the IO queue system, unfortunately.

dormando:
went off into another branch to refactor the pending IO subsystem. already a good way through it.

dormando:
blocked on #1193 - will need to rebase or rewrite on top of that once it's merged.

dormando:
unblocked. passes unit tests and a quick run through the stability burn-in suite... need to carefully audit everything and add more tests now.

looking into why it's still causing GC runs... we're getting the stack grow/shrink on every request which sucks... and it's freeing thread CallInfo's when the GC does run.

dormando:
note to self: `thread_libevent_ionotify` is high because in V2 for an in-memory internal call, we're roundtripping through the IO notify subsystem before resuming the request context. flamegraph shows a few percent in read/writes for the ionotify, plus we will have to roundtrip through epoll before continuing rcontext.

I was doing that to avoid recursion when running queue's but I'll need to give this a serious stare since it's apparently worse than the lua GC overhead we were just fixing.

dormando:
think I need to rip the bandaid off and avoid the roundtripping through epoll for locally executed background IO's and locally returning background IO's.

Been mapping it out. requires some care to avoid problematic recursion.

dormando:
(ignore the second commit for a bit here)

dormando:
burn-in running now on the underlying IO optimization + this branch.

audit seems okay... tested the error path by manually breaking it then hammering it. no leaks/etc.

just need to write more tests for sub-rctx internal fetches.

dormando:
needs to be used/referenced in some place where config reload happens. burn-in suite busted that.

dormando:
I _thought_ it had crashed but it didn't reproduce. re-running the full suite several times just in case.

dormando:
burn-in passed several times.

going to audit the code one more time and _probably_ rename the reference to `mcp.backend_pool`, then merge and move on.

PR #1193, opened 2024-12-30
Refactoring of background IO subsystem

dormando:
The IO subsystem was written as a semi-abstraction specifically for extstore. The proxy piggybacked on this system to save time. Sadly the proxy has weird semantics... like turning one IO into many IO's, or completing the original request while background IO's are still outstanding... which doesn't line up with how the IO system works.

Which lead to some hacks to ensure the proxy code can always make progress, plus convoluted completion detection code.

This PR changes the system from a set of conn object oriented queues to per-thread work queues. This lets us submit more work (such as backgrounded proxy coroutines) without access to a connection object. Further, response objects are now individually suspended (and counted on a connection), instead of constantly checking every queue during request handling.

Current status: code complete. Needs burn-in testing and extra unit tests.

---
TODO:
- [x] t/error-extstore.t is the only non-passing existing test. The abort code is incomplete so I'm guessing it's that.
- [ ] Need additional tests on subrctx scenarios: all down backends, re-running waits with everything completed, sub-sub-sub-rctx, etc.
- [x] Re-evalute all FIXME/TODO's and expand this TODO.
- [x] See if there's anything else we can now simplify in the proxy code (after mcp.internal is retired, yes)
- [x] Try to get rctx:sleep() to function in the proxy code, since this is supposed to be the required fix for that.
- [x] Add some extra tests for rctx:sleep()? IE: call multiple times, call after sending req/res, call after enqueue but before wait.
- [x] split backend IO queue into to-write and to-read, remove the cursed flushed position tracker code
- [ ] ~~See if we can get a cleaner method of passing IO's into extstore now (see below)~~
- [x] Attempt to re-create the scenarios which caused protocol desync in the past (difficult, not easy/possible to do via integration level tests) (apparently the stability tests automatically do this by reducing the BE_IOV_MAX in debug binaries)
- [x] performance audit... see if there're any branches that can be cut back.


dormando:
Maybe I'll just give up and make the pending object slightly less dynamic (which might let me shrink the size anyway).

IE:
- combined bit flag section for all of the random bool's
- shared iovec offset section
- maybe an extra void ctx ptr (for rctx in proxy)
- payload section now always starts in the same offset.
- remove payload offsetof code.
- specify the size of iop's as the max() of a couple sizeof's if possible to save some RAM.

since iop objects are fixed size the extra fields aren't a waste anyway.

would be nice to get a define/something going for the repeated first section of the struct too. can't remember what I got stuck on when initially doing that.

dormando:
Thinking it makes more sense to do major reworking on the `io_pending` type in a different PR.

instead I'll spend the brain points on giving backends split write/read queues and rctx:sleep

dormando:
Think this is actually code complete now? Only had a couple hours to work on it today even. Happily the rctx:sleep code JFW'd. glad I left it merged with tests but disabled!

Next step is to put it through burn-in testing while expanding the unit tests. Then do a more careful audit.

dormando:
Paying extra attention to the backend request list read/write split since that's where our worst issues have come from. Just did some digging:

jan 22 2023 - ac55ac888e6252836ca4a233daf79253934c5728
optimization was added dec 5 2022 - fixed major slowdown when large IO queues happen: was looping over every pending request repeatedly. Fix was to add a marker into the queue to avoid looping over requests that were already sent to the backend.
the bug fixed above was memory corruption caused by the marker pointer being handled incorrectly.

may 3rd 2024 - 8a188a09916029abbc15526bbcd4b02c5b94ccc2
This change fixed an issue caused by the prior optimization: extra latency was caused when backend depth was > 500

may 28th 2024 - 33c00b49f20a092a33ffb5d48ff3a0037088fc7c
The prior commit caused a new bug which corrupted memory (giving bad responses) now for the second time: the offset pointer was being set wrong, causing memory corruption in scenarios caused by network lag.

The changes in this PR should make the original bugs impossible. I've pulled notes from the testing I did for the previous remediations and will re-run them.

dormando:
running into an odd issue with the stability benchmarks.

the fastgood multiget tests (rapidly create background requests) are getting behind with this branch and not in `next`... mystery to me so far. everything's writing properly, everything's reading properly, and when read/write code fires didn't change here.

it's almost like it's running a bit faster (which can end up making too many background requests even in the old code) but the rps seems the same.

dormando:
symptoms are bizarre, so I'm moving on with the performance A/B test to see what stands out.

- The initial test to "leak" was a standard multiget test, which splits into single gets on the proxy backend.
- this was creating async requests too fast and running out of memory
- I did a ton of digging and saw the exact same network traffic in both versions. Nothing leaks, everything catches up once the load is off, etc.
- The standard get requests only process one key per packet though. stacked pipelining only works with metaget (I might fix this now that I've seen it....)
- so I switched to running a deeply pipelined metaget test instead, which is showing a perf boost from 600krps to 1mrps (or higher, depending), which causes the same problem with FASTGOOD making async requests too quickly.

Used some new gdb level utility functions to examine the backend states mid-test: everything is flushed to the backend and they're just waiting for results.
I also turned the recv() handler into a loop to try to read faster but without any luck.
the `next` branch gets more data in each recv for reasons I don't understand: if I add tiny occasional sleeps to the read code it looks more similar to the `next` branch, which makes me think it could be the event loop running faster. if I totally pause the proxy for a second it'll get huge read loops and get all data back quickly.

I also tried running backend nodes A/B without any difference. GDB'ed them and tcpdump'ed and evrything's correct.

Maybe there's some socket option I screwed up?

Running the full performance suite A/B now to see if it gives anymore useful information.

dormando:
should still force myself to audit the tests for subrctx tests that were missed. otherwise we're just doing burn-in at this point; I can't think of anything else that I want in this particular PR.

dormando:
okay perf tests came back and there's no massive difference in performance (or any failures) in any test.

It seems to be consistently a few dozen or a hundred microseconds worse in some situations, but often tighter on outliers (fewer > 1ms compared to `next`). capacity is the same for all of the perf tests. need to compare perf tests with the stability tests and see if there's anything I need to add to perf. (very deep pipelines?)

re-running stability tests with 120s runs...

dormando:
performance mystery mostly solved over a few more iterations of code, I think:

- Added an optimization for batched writes and now the perf results for the `io_refactor` branch are equal to or better than `next`
- watched the get rate directly on the backend nodes while running A/B tests with the stability tests that failured before and the total rate is just a bit higher, and the backends were just a bit behind which caused the active request bloat.

so I think it's all solved now. Should probably still audit unit tests and add some extra subrctx tests but otherwise I think this is complete and just needs baking.

dormando:
think I'm going to go ahead and merge this into `next`. I'll keep developing on it and running burn-in tests for the week before cutting a release.

PR #1195, opened 2025-01-14
proxy: speed up per-worker backend IO

dormando:
Uses a shorter return path for completed IO objects, resulting in 5%+ performance bump.

Also cleans up some of the remaining hack code from before the IO subsystem fix.

---

TODO:
- [x] performance suite pass to confirm deltas in more scenarios.
- [x] stability suite burn-in pass.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -547,8 +547,9 @@ void conn_worker_readd(conn *c) {
             // Explicit fall-through.
         case conn_io_queue:
             conn_set_state(c, conn_io_resume);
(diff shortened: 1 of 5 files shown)


dormando:
Performance looks good. Slight boost in p99 vs pre io-refactor but solid overall boost when comparing histograms. Need to log p50/75/90 instead of just p99 :)

Now needs to run a burn-in test.

dormando:
doing one more pass...

Issue #1196, opened 2025-01-15
Remove "Override the size of each slab page" from the man page for the -I, --max-item-size option


thatsafunnyname:
**Describe the bug**

Misleading/incorrect ```"Override the default size of each slab page"``` in the man page for the  ```-I, --max-item-size``` option.

**To Reproduce**

```
> man memcached | grep -A2 max-item
       -I, --max-item-size=<size>
              Override the default size of each slab page. The default size is 1mb. Default value for this parameter is 1m, minimum is 1k, max is 128m.  Adjusting this value changes
              the item size limit.
```

I will file a PR against the ```next``` branch.

----

```"Override the size of each slab page"``` was removed from the ```--help``` in https://github.com/memcached/memcached/commit/3f3e1379368753e540a332f90f996a3eb7ab8e9e
 
As per https://github.com/memcached/memcached/issues/1094#issuecomment-1826234316
 
-   _Pages are fixed at 1mb in size._

I think since version 1.4.29 , released on 2016 July 13th

https://github.com/memcached/memcached/wiki/ReleaseNotes1429#new-features

  _"The maximum slab chunk size is no longer tied to the maximum item size."_

Chunked (large) item support - https://github.com/memcached/memcached/pull/181/files#diff-e471fd56f09dcb25f96a32a2645c4511b65a12fb4cda7fced63d379789050f76R229

dormando:
Ah this did get fixed. closing.

PR #1197, opened 2025-01-15
Remove "Override the size of each slab page" from the man page for the -I, --max-item-size option

thatsafunnyname:
https://github.com/memcached/memcached/issues/1196
 
I think since version 1.4.29 , released on 2016 July 13th, pages are fixed at 1mb in size and the maximum slab chunk size is no longer tied to the maximum item size.
 
"Override the size of each slab page" was removed from the --help in https://github.com/memcached/memcached/commit/3f3e1379368753e540a332f90f996a3eb7ab8e9e


dormando:
Thanks! I do neglect that man page.

dormando:
merged, thanks!

PR #1198, opened 2025-01-16
proxy: per-backend automatic logging

dormando:
Options similar to mcp.log_reqsample() are applied to responses as they're read off the network.

Can be a lot faster than relying on callbacks to do logging, if you're not doing logging inline with a route handler. Also simpler in some cases (ie routelib).

This code also leverages the removal of V1 and cuts 250 bytes from every res object, resulting in less memory used per "request slot", and an extra minor speedup on _all_ requests from less time in memset and not copying the backend host/port onto every response.

TODO:
- [x] Fix for `internal` response logging since we've removed the response-inline label.
- [x] Additional tests. (with/without detail, error/no error, rate check, etc)
- [x] Double check the tag/detail logic for HIT logs. It might print HIT's too often.

### DIFF
--- a/logger.c
+++ b/logger.c
@@ -6,7 +6,6 @@
 #include <string.h>
 #include <errno.h>
 #include <poll.h>
-#include <ctype.h>
 #include <stdarg.h>
 
 #if defined(__sun)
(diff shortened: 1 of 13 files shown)


dormando:
Not happy with the way tags are being handled, after sleeping on it:
- need a "tag = " that is the only tag in case user doesn't intend to differentiate
- `tag_hit` should probably be called good to match up with the rest of the nomenclature.
- worried people might want tags based on command executed, but the command is in the log so it should be fine?

dormando:
check:

GOOD vs OK vs ERR is based off of rstatus and rcode, which are already in the log line (code and status). The detail/tag is only varying based on rstatus and rcode.

Is it reasonable to have a single tag and make GOOD/OK/ERR more obvious?

`ts=%lld.%d gid=%llu type=proxy_req elapsed=%lu type=%d code=%d status=%d cfd=%d be=%.*s:%.*s detail=%.*s req=%.*s\n`

dormando:
just need some additional tests now.

PR #1203, opened 2025-02-08
FIX: fix 2 memory leak bugs in file mamcached.c

mugitya03:
Dear Developers:

We found and fixed 2 potential memory leak bugs in the file `mamcached.c`. Here are the details:

**1**
Function `strdup` can return a memory object. The memory is stored in [`subopts_orig`](https://github.com/memcached/memcached/blob/1b981b040d51d3cedb7e4a001ab642c46a685271/memcached.c#L5179).
```c
subopts_orig = subopts = strdup(optarg);
``` 
The memory is not freed when the function exits in exception processing branches.

**2**
A memory is allocated and stored in pointer [`temp_portnumber_filename`](https://github.com/memcached/memcached/blob/1b981b040d51d3cedb7e4a001ab642c46a685271/memcached.c#L6089). 
```c
temp_portnumber_filename = malloc(len);
``` 
It is not freed when the function exits [here](https://github.com/memcached/memcached/blob/1b981b040d51d3cedb7e4a001ab642c46a685271/memcached.c#L6132)
```c
    if (stats_state.curr_conns + stats_state.reserved_fds >= settings.maxconns - 1) {
        fprintf(stderr, "Maxconns setting is too low, use -c to increase.\n");
        exit(EXIT_FAILURE);
    }
```

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -6163,6 +6163,7 @@ int main (int argc, char **argv) {
             } else {
                 vperror("failed to listen on one of interface(s) %s", settings.inter);
             }
+            free(temp_portnumber_filename);
             exit(EX_OSERR);
         }
 
--- a/memcached.c
+++ b/memcached.c
@@ -5192,17 +5192,20 @@ int main (int argc, char **argv) {
             case HASHPOWER_INIT:
                 if (subopts_value == NULL) {
                     fprintf(stderr, "Missing numeric argument for hashpower\n");
+                    free(subopts_orig);
                     return 1;
                 }
                 settings.hashpower_init = atoi(subopts_value);
                 if (settings.hashpower_init < 12) {
                     fprintf(stderr, "Initial hashtable multiplier of %d is too low\n",
                         settings.hashpower_init);
+                    free(subopts_orig);
                     return 1;
                 } else if (settings.hashpower_init > 32) {
                     fprintf(stderr, "Initial hashtable multiplier of %d is too high\n"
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! I'm not sure these really count though. In both cases the binary is exiting.

mugitya03:
Thank you for your response. We acknowledge that this case is somewhat special compared to other typical memory leak scenarios. The allocated memory, though not manually freed, will indeed be reclaimed by the system when the program terminates.

Our analysis tool detects memory leaks based on an abstract model, where it checks whether the memory returned by malloc is freed on every possible program path. Under this abstraction, this case is classified as a memory leak.

We appreciate your feedback and understand your perspective. However, from the standpoint of code quality and consistency, we still encourage you to consider our PR. Ensuring explicit memory management not only improves the overall code style but also makes the codebase more robust and maintainable. We believe this would be a valuable enhancement for the project.

dormando:
@mugitya03 for the first commit, a `goto error:` pattern would be better. otherwise this change is making the code worse since any new arg will need the extra free line, which would be prone to human error.

If that's too much churn I'd prefer to leave it and deal with it if we do a bigger refactor later. As I noted it doesn't actually fix anything.

dormando:
@mugitya03 I don't have any problem with the second commit. If you move it to a separate PR I could merge that one.

mugitya03:
Thanks for your further response. I'll split these two commits into two separate PRs and modify the first commit.

dormando:
dealt with via other PR's.

PR #1206, opened 2025-02-19
proxy: deduplicate protocol processing code

dormando:
This commit is step one of... a few... for standardizing the code paths for protocol handling. We started with three:
- mcmc response parsing
- proxy_request.c request parsing
- proto_text.c (original tokenizer)

Both requests and responses can now be tokenized by the same code from mcmc. This commit series does this in stages, which will result in removal of duplicated code and fixing many long standing issues.

In this particular commit the duplicated tokenizer is removed from proxy_request.c, and mcmc functions are now used in various places.

All of the mcmc related functions have actual C unit tests behind it which should help them be more trustworthy.

TODO:

This work is being done in stages because there's too much code to hold in my brain at once. Not necessarily in order, but:

- [ ] convert strtol to mcmc number parsing functions that don't rely on C strings.
- [ ] use more convenience functions in proxy_internal.c instead of directly accessing tokens all the time
- [ ] can even more of the protocol parsing in proxy_request.c be removed?
- [ ] can we use an mcmc request struct instead of rq->pr?
- [ ] remove duplicated parsing code entirely from proxy_internal.c
- [ ] (big task) evaluate requirements for converting proxy_text.c to new tokenizer

Got another month of work on this including extra testing that should be done.


dormando:
I added a lot of generic parsing code to mcmc while writing the inspector and mutator last year, which I've been able to use mostly without modification for this first big change.

If I can get `mcp_parser_t`  moved from proxy.h to mcmc that would be a huge win: I could move the entire request parser into mcmc and thoroughly test it, then reuse _that_ code for proto_text.

dormando:
still want to revise the mcmc_ API's... but the plan is to probably keep using them by converting code and then update and re-test everything.

dormando:
Trying to get the request/response struct's merged and fix some awkwardness with how mcmc's types/codes are handled. Taking more time than I hoped.

dormando:
trying to de-scope "single struct for req/res" from this PR. going to stick with:
- audit/simplify random bools
- see about de-scoping lifting mcp_parser_s into mcmc (might just temporarily add a parser code unit)
- big protocol conversion stage (put storage stuff into the parser code unit?)

then in separate PR's do more mcmc reworking...

dormando:
cleaning up the ascii_multiget hack was time consuming but helps a shitton in reducing code fragility...

dormando:
aight got a few hacks cleaned up and did planning work for the big merge:
- pull all the protocol handling out into its own code file (including the request parser code, for now)
- convert proto_text.c to use this new file (ow)
- commit all of that once working again...

... then will re-evaluate if it's a good time to continue doing mcmc work or not.

I kiiiinda feel like I could get something big out of finishing the mcmc work... since I could write some wrapper code and get an "instant mc-speaking server" out of it. which would allow simpler experiments with funny backends by just having them be proxy backends.

dormando:
got things wired up so I could convert one command at a time. still need another round of parser work after this stage though. All of the `proxy_internal.c` code is converted and done now, so that file is nice and small!

down to the last meta command for `proto_proxy.c`: meta set. which is complicated due to the network read stage splitting the processing. I'm going to try to get it on the common code and un-splitting the command handler so it's more straightforward.

dormando:
figuring out mset without duplicating code was a pain but I think it's close now.

all of the base meta commands are ported.

next is to port the base text protocol... get commands are going to be the annoying one since I'll need to write a wrapper.

dormando:
migrating get is moderately painful because of the multikey syntax. attempting to lift the recent hack cleanup code from proxy's version of this.

dormando:
all the storage commands are done. need to hype myself up for converting the maintenance commands to use the new tokenizer and an array/hash table instead of an if/else tree.

then... delete the old tokenizer code and re-evaluate for what's next.

dormando:
been having trouble getting time to sit down on this lately.

just finished the asciiauth and the old tokenizer is now completely gone. input command buffers are now `const`! (or can be made `const` everywhere...)

the next big step is a lot of small cleanups. then extended testing can begin.

dormando:
sigh. wtf was I doing....

dormando:
fixing the command length issue took way too long to unwind... looks great now though.

dormando:
updated the TODO list.

dormando:
maybe one more day to complete this? these are short days right now though :(

dormando:
oh right.. need to think about testing again..

dormando:
can't find any regressions in the noreply code... but I do need to do a separate change to lift the noreply handling up to the top level request context code.

dormando:
performance testing showed a regression in the ~10us range. pinned to the mcmc tokenizer and eventually refactored it into being more performant in 9/10 cases.

doubt any users would be able to measure the difference but at least it's better and not worse.

I'm giving it one more big regression scrub then will merge. there's still a big list of stuff I want to do but they need to be individual PR's. This is already a big improvement; the various code trees are merged and a lot of workarounds/hacks were removed.

The big delayed items are around moving the rest of the server parsing code to mcmc so it can be C tested and reused, and ideas around making the parser more pluggable. A bunch of strto's still exist but are harmless now.

dormando:
have a noreply problem still...  meta works but ascii doesn't. Despite cleaning it up twice in here it's still overly complex. might have to take a stab at fixing it more.

have to map the whole stupid thing out and look again.

dormando:
think I'm calling it for this one... I can keep fiddling forever but it needs to get merged.

going to hold this to sneak in some smaller fixes for mutator/inspector first, cut a release, then merge immediately afterward so it can stage for the following tarball.

If I get a sudden burst of motivation, the memory perf tests I did could use incr/decr and non-meta tests. though that code is all unit tested as well as tested by other shredders suites.

dormando:
rebased over the mutator changes in `next`.

should be clear to merge... blegh..

dormando:
note to any passers-by: I've merged this just after a release so it can get extra testing while I continue to develop on top of it for the next month. so be careful if you deploy the `next` branch instead of tagged releases.

Issue #679, opened 2020-05-13
'current_time' vs live time checks and edge cases


dormando:
the `current_time` variable is a monotonic counter that's updated once per second via a libevent timed event. It was added to avoid overhead of fetching the clock, especially on systems where that function is slow (not typically linux).

It's probably worth re-evaluating this, at least subtly.

1. Benchmark under high thread load with `current_time`'s most frequent usages replaced with a direct clock call. Is `current_time` a bottleneck on some systems due to false sharing/etc? Is that the same with the system clock?
2. What systems still exist which have a slow clock? Any of the BSD's? Mac?
3. For cases where SET'ing a 1s TTL, the expiration time can be anywhere from immediate to nearly two seconds. Is the overhead of a clock call low enough that we can check the "time since the last time `current_time` was updated" during a SET and increment the TTL where it makes sense? Ensuring an item will last for 1-2 seconds rather than 0-2 seconds (or 1-3 seconds if blindly incrementing).

Might be a fun task for a student. Make sure you use mc-crusher to really get the request rate high.

dormando:
https://github.com/memcached/memcached/issues/307 - was reported here years ago. users do actually run into this.

PR #1207, opened 2025-02-25
stats: Update time_since_server_cert_refresh to prevent showing underflow result

YueTang-Vanessa:
#### Issue
* Memcached version 1.6.17

We received a case that the stats `time_since_server_cert_refresh` shows unreasonable number

```
STAT uptime 21323273
STAT time 1734085745
STAT rusage_user 941.709452
STAT rusage_system 1270.867859
STAT time_since_server_cert_refresh 1864631599
``` 
`APPEND_STAT("time_since_server_cert_refresh", "%u", now - settings.ssl_last_cert_refresh_time);`

This seems like an edge case lead to underflow related to known current_time issue: https://github.com/memcached/memcached/issues/679.

Since we have customers monitoring this metric, and considering this known issue has no signs of change in the upcoming plans. Raise this PR to prevent showing edge case number in stats response.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1885,7 +1885,7 @@ void server_stats(ADD_STAT add_stats, void *c) {
         }
         APPEND_STAT("ssl_handshake_errors", "%llu", (unsigned long long)stats.ssl_handshake_errors);
         APPEND_STAT("ssl_proto_errors", "%llu", (unsigned long long)stats.ssl_proto_errors);
-        APPEND_STAT("time_since_server_cert_refresh", "%u", now - settings.ssl_last_cert_refresh_time);
+        APPEND_STAT("time_since_server_cert_refresh", "%u", (now >= settings.ssl_last_cert_refresh_time) ? (now - settings.ssl_last_cert_refresh_time) : 0);
     }
 #endif
     APPEND_STAT("unexpected_napi_ids", "%llu", (unsigned long long)stats.unexpected_napi_ids);


dormando:
Thanks!

I don't think this is actually related to #679 - `current_time` cannot go backwards, `current_time` can start weird and maybe refresh_time is getting initialized to `current_time` before `current_time` gets adjusted during startup, leading to a weird initial value?

Might be better to directly call the clock for this use case, else double check the start sequence.

dormando:
I'm pointing this out because I think with your change and a weird initial value it could print `0` for a long time, which is also not useful to the user.

dormando:
Do you have a reproduction case? Or any other information as to when/where this happens? It doesn't seem to be what I expected

YueTang-Vanessa:
> Do you have a reproduction case? Or any other information as to when/where this happens? It doesn't seem to be what I expected

It is hard to reproduce this issue. I found some metrics in the investigation, operator manually updated the cert, but this metric still off at the time.
```
before:
STAT time_since_server_cert_refresh 1864634526

Manually triggering refresh certs:
refresh_certs
OK

After:
STAT time_since_server_cert_refresh 1864631599

time_since_server_cert_refresh is being changed. We can see it was decreased by 2927
```

YueTang-Vanessa:
I tried several times and unfortunately cannot reproduce this issue. We just observed this case on one server, so maybe this is an one-off event.

I tried to send `stats` command to server several times, I observed that, while the `uptime` and `time` is increasing, the `time_since_server_cert_refresh` haven't changed which is weird to me. Shouldn't the `settings.ssl_last_cert_refresh_time` stay the same unless there has a successful cert refresh?
```
stats
STAT uptime 28371917
STAT time 1741134929
STAT time_since_server_cert_refresh 1858019468
...
stats
STAT uptime 28371929
STAT time 1741134941
STAT time_since_server_cert_refresh 1858019468
...
stats
STAT uptime 28371964
STAT time 1741134976
STAT time_since_server_cert_refresh 1858019468
```

dormando:
`time_since_server_cert_refresh` is literally using the same variable (`now`) as uptime/time so if those two change it should also change.

Are you sure this server hasn't been modified in some way? I can't at all see how this is possible.

QuChen88:
What happens if there is a short clock drift in between the time of the cert update to now? Would that cause any such issues?

dormando:
We're using the monotonic clock, and the cert update timestamp is based off of the same clock variable as the rest of the code. So if the clock ever drifts, the next time the certs reload it should fix itself. It should again be relative to `current_time`

- but it's not doing that in the example above? updating the certs is changing the delta but not fixing it... and the delta doesn't move with the normal clock.

Which makes me suspicious that this isn't my code?

YueTang-Vanessa:
This issue is more likely an one-off event. Close this PR as it will not help with the condition.

dormando:
Please let me know if you see it again. Thanks!

YueTang-Vanessa:
Hi @dormando, we do see this issue again when we tried to update the ssl certificate on one 6 nodes cluster, 5 out of 6 nodes are fine, just one node showing large value.

Healthy node:
```
STAT uptime 36814207
STAT time 1761086308
STAT version 1.6.22
STAT libevent 2.1.11-stable
STAT pointer_size 64
......
STAT ssl_handshake_errors 5377
STAT ssl_handshake_success 51140837
...
STAT time_since_server_cert_refresh 10853162
...
```

Problematic node:
```
STAT uptime 36814208
STAT time 1761086309
STAT version 1.6.22
STAT libevent 2.1.11-stable
STAT pointer_size 64
...
STAT ssl_handshake_errors 14433
STAT ssl_handshake_success 54657987
...
STAT time_since_server_cert_refresh 4269006191
...
```

Could it be a u32 overflow/underflow problem?

dormando:
@YueTang-Vanessa I've asked a couple times if the code's been modified, and I see you have some counters that I don't, so I suspect it has been :)

This bug is impossible to cause in the code that I have. There is nothing complex around the timestamp handling. Referring to version 1.6.22:

```
memcached.h:537:    rel_time_t ssl_last_cert_refresh_time; /* time of the last server certificate refresh */
memcached.h:906:extern volatile rel_time_t current_time;
tls.c:163:        settings.ssl_last_cert_refresh_time = current_time;
memcached.c:1785:    rel_time_t now = current_time;
memcached.c:1923:        APPEND_STAT("time_since_server_cert_refresh", "%u", now - settings.ssl_last_cert_refresh_time);
// rel_time_t is defined as:
typedef unsigned int rel_time_t;
```

There are several constants:
- `current_time` is a volatile int.
- `current_time` is based off of `CLOCK_MONOTONIC`, it cannot go backwards.
- All of the relevant types are `rel_time_t` (which is unsigned int)
- `now` in the print _must_ be equal to or greater than last_cert_refresh_time, which is only ever set from the same source as `now`
- Since we're always checking `current_time` vs the refresh time during the stats call, the stats output _must_ change once per second.

Some possibilities come to mind:

- You're not using `CLOCK_MONOTONIC` and we get negative values sometimes (how? is refresh_cert being called multiple times in one second?) Unlikely since refreshing again would fix it, but you said that doesn't work?
- You've modified `rel_time_t` but aren't telling me (which could allow %u to give a bunk print? maybe?)
- You're on exotic hardware where a 4 byte volatile int isn't always atomic and we're assigning `current_time` to another variable halfway through an update and getting a weird number?
- You've modified `rel_time_t` to be 8 bytes on a platform with 4 byte atomics and causing the above issue.
- The code has been modified in some other way?
- Is `refresh_certs` getting called in parallel a bunch of times in parallel (might be missing a config lock?)? Why doesn't it fix itself on a further refresh?

I'm fairly sure the first and third aren't possible either, since the number should still move on successive `stats` calls. If `current_time` doesn't work properly the item expiration times would be super corrupt.

Is your source code available anywhere? Or am I missing something?

dormando:
@YueTang-Vanessa Hi, any update? Anything else I can look at or any new details?

YueTang-Vanessa:
Thanks for your quick response, IIRC, we only introduced some informative metrics in the source code for better monitoring purpose, I will try to check if we have more data point when I have some bandwidth.

PR #1211, opened 2025-03-07
assoc.c: fix leak in assoc_get_iterator when trylock fails

kanak:


kanak:
@dormando : i saw this while reading the assoc code. 

I have two other minor improvement ideas in assoc, wanted to run them by you first:
* `assoc_insert` returns an int (always 1) but none of the current callers care. I want to change it to `void`.
* `assoc_delete` takes `key, keylen, hash` but the item must already exist in table, and all current callers have an item already. maybe we should change this to take `item* and hash_value` like the `assoc_insert` function?

This one would increase the size of item struct and I'm guessing it's not worth it, but mentioning it just in case:
* since `h_next` is singly linked, when deleting items, we have to re-traverse the bucket. If it was doubly linked, we could do constant time deletes (since already have item with us when performing delete).

Thanks, and looking forward to your feedback.

dormando:
Swear I responded to this a few weeks ago, sorry. This is a legitimate bug, thanks!

kanak:
@dormando: thank you so much. What do you think about the other assoc improvement ideas in that comment. No worries if you just want to leave working code alone :).

dormando:
Yeah in the comment that I was pretty sure I left I'd said that I don't deliberately change working code unless it gets cleaned up while being changed for some other reason.

The `h_next` thing isn't measured as a major performance issue so it's fine. We need every byte of memory.

PR #1219, opened 2025-03-24
crawler: speedup and extstore info

dormando:
Unrolls the sprintf used for `lru_crawler metadump` since it is a hot path. Increases dump speed with no crawler sleep by 30%.

Also adds the disk page and disk page offset via `ext_page` and `ext_offset` if the item is on disk.

Credit to @hemal-shah for the idea on dumping extstore information.

### DIFF
--- a/crawler.c
+++ b/crawler.c
@@ -15,7 +15,6 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <stdio.h>
-#include <signal.h>
 #include <string.h>
 #include <time.h>
 #include <assert.h>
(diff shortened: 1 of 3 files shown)


crawler.c
```diff
@@ -278,28 +285,81 @@ static void crawler_metadump_eval(crawler_module_t *cm, item *it, uint32_t hv, i
...
+#endif
+
+    KADD(p, "\n", 1);
+
+    refcount_decr(it);
+    assert(p - start < LRU_CRAWLER_MINBUFSPACE-1);
+    cm->c.bufused += p - start;
 }
 
+#undef KVADD
```
hemal-shah, line 360:
```suggestion
#undef KADD
```

crawler.c
```diff
@@ -278,28 +285,81 @@ static void crawler_metadump_eval(crawler_module_t *cm, item *it, uint32_t hv, i
...
-        // Failed to write, don't push it.
-        return;
+    assert(it->nkey * 3 < LRU_CRAWLER_MINBUFSPACE*2);
+    // unrolled snprintf for ~30% time improvement on fullspeed dump
+    // key=%s exp=%ld la=%llu cas=%llu fetch=%s cls=%u size=%lu flags=%llu\n
+    // + optional ext_page=%u and ext_offset=%u
+
+    char *p = cm->c.buf + cm->c.bufused;
+    char *start = p;
+    KADD(p, "key=", 4);
```
hemal-shah, line 295: nit: wouldn't it be better to have static strings instead of magic strings.

dormando: would be inconsistent with rest of code + it ends up being a lot more verbose. not really, no.

dormando:
took a second to ensure sizeof("static string") works the way I expect via the macro and seems fine, so that removes all of the manual strings lengths.

Issue #1220, opened 2025-03-24
`t/proxyunits.t` failing on 32bit x86 Debian


tianon:
**Describe the bug**
Running `make test` on a build with `--enable-proxy` fails in `t/proxyunits.t` on i386 Debian.

**To Reproduce**
See https://github.com/docker-library/memcached/pull/101, but it's a pretty vanilla `./configure --enable-proxy ... && make && make test` in a 32bit Debian (12) environment.

```console
$ docker run -it --rm --pull=always --platform linux/386 debian:bookworm-slim
bookworm-slim: Pulling from library/debian
bdeb081d427d: Already exists 
Digest: sha256:1209d8fd77def86ceb6663deef7956481cc6c14a25e1e64daec12c0ceffcc19d
Status: Downloaded newer image for debian:bookworm-slim
root@f7a29a944b57:/# apt update -qq
All packages are up to date.
root@f7a29a944b57:/# apt install -yqq wget make gcc libevent-dev perl
The following additional packages will be installed:
... (output removed for brevity)
root@f7a29a944b57:/# cd /tmp/
root@f7a29a944b57:/tmp# wget https://www.memcached.org/files/memcached-1.6.38.tar.gz
--2025-03-24 23:07:31--  https://www.memcached.org/files/memcached-1.6.38.tar.gz
Resolving www.memcached.org (www.memcached.org)... 107.170.231.145
Connecting to www.memcached.org (www.memcached.org)|107.170.231.145|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1282099 (1.2M) [application/octet-stream]
Saving to: 'memcached-1.6.38.tar.gz'

memcached-1.6.38.tar.gz                        100%[=================================================================================================>]   1.22M  --.-KB/s    in 0.1s    

2025-03-24 23:07:31 (10.3 MB/s) - 'memcached-1.6.38.tar.gz' saved [1282099/1282099]

root@f7a29a944b57:/tmp# tar -xf memcached-1.6.38.tar.gz 
root@f7a29a944b57:/tmp# cd memcached-1.6.38
root@f7a29a944b57:/tmp/memcached-1.6.38# ./configure --enable-proxy
checking build system type... x86_64-pc-linux-gnux32
checking host system type... x86_64-pc-linux-gnux32
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... no
[...]

PR #1221, opened 2025-03-25
Fix `mcplib_time_*_millis` overflow on 32bit systems

tianon:
This fixes a time overflow by casting `tv_sec` to `lua_Integer` (which is `long long` unless explicitly configured otherwise) before multiplying it by 1000, which ensures the result doesn't become negative before being assigned to a `lua_Integer` value and returned.

Concretely, this fixes the test failure in `t/proxyunits.t` on systems that have not yet transitioned to `time64_t` (notably, Debian 12 / Bookworm; that transition should be part of 13 / Trixie, which will likely be released later this year).

Fixes https://github.com/memcached/memcached/issues/1220

### DIFF
--- a/proxy_lua.c
+++ b/proxy_lua.c
@@ -223,7 +223,7 @@ static int mcplib_schedule_config_reload(lua_State *L) {
 static int mcplib_time_real_millis(lua_State *L) {
     struct timespec now;
     clock_gettime(CLOCK_REALTIME, &now);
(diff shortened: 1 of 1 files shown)


dormando:
merged to `next`, thanks!

tianon:
Oops, sorry, I forgot I should target `next` instead of `master`!

dormando:
no worries, was a trivial rebase.

PR #1225, opened 2025-04-10
Removing `proxy_req_await` from protocol.txt

YufeiHu:
Removing `proxy_req_await` from protocol.txt as `proxy_req_await` is already deprecated.


dormando:
nice catch, thanks!

dormando:
merged to `next`

PR #1227, opened 2025-04-18
Implement tests of best_result function comparing mutator errors with nil 

ffrankis-stripe:
One of these tests currently fails, indicating a possible bug to be patched.

This is not quite merge-ready -- I want to clean up some of the variable/function names and also add equivalent tests for `worst_result`.


dormando:
think this got resolved upstream with different tests but I forgot to close it, and it didn't get cleaned up either.

PR #1228, opened 2025-04-21
Adds info for listeners to help string

ailangovan:
This listener feature was super helpful but non-obvious from the help messaging. Adding a little more context in the help messaging so it's clearer for folks in the future. 

Thanks for the help identifying this for us @dormando !


ailangovan:
I elevated this out of draft. Thanks for the 👀

dormando:
merged to `next`, thanks!

PR #1232, opened 2025-06-29
ci: Enable proxy in CI

fsaintjacques:
Proxy is currently disabled in CI, this is a step towards enabling it.

PR #1233, opened 2025-06-29
proxy: Add farmhash hash function and modulo distribution

fsaintjacques:
This PR is a followup of the CI [PR](https://github.com/memcached/memcached/pull/1232). See the commit message for the description of the changes. You might wonder why I'm adding such changes. I'm currently prototyping an architecture where datasets (in OLAP storage) are partitioned by a primary key and then loaded into sharded memcached instances. The dataset size will drive the number of memcached instances (and not the reverse).

The loading itself (writing to memcached servers) is *not* going through the proxy but directly on each backend via sidecars in kubernetes. On top of that, I have the following restrictions:

- The hashing scheme is restricted to what the OLAP system provides me, in this case [FARM_FINGERPRINT](https://cloud.google.com/bigquery/docs/reference/standard-sql/hash_functions#farm_fingerprint)
- The distribution mechanism is also dictated by the OLAP system, hence the modulo.
- Bigquery only supports signed INT64, hence why the optional masking is added to the modulo. (You can see how this fits in the integration test).

This PR significantly help me reducing the number of moving pieces required to make it "work".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,13 @@ jobs:
     - name: Install deps
       run: |
         sudo apt-get update -y
-        sudo apt-get install -y libevent-dev libseccomp-dev git libsasl2-dev libio-socket-ssl-perl
+        sudo apt-get install -y libevent-dev libseccomp-dev git libsasl2-dev libio-socket-ssl-perl wget
+        cd vendor && ./fetch.sh
     - name: Build
       run: |
         gcc --version
         ./autogen.sh
-        ./configure --enable-seccomp --enable-tls --enable-sasl --enable-sasl-pwdb
+        ./configure --enable-seccomp --enable-tls --enable-sasl --enable-sasl-pwdb --enable-proxy
         make -j
     - name: Test
       run: PARALLEL=5 make test
--- a/Makefile.am
+++ b/Makefile.am
@@ -60,6 +60,8 @@ if ENABLE_PROXY
 memcached_SOURCES += proto_proxy.c proto_proxy.h vendor/mcmc/mcmc.h \
 					 vendor/mcmc/mcmc.c \
 					 proxy_xxhash.c proxy.h \
+					 proxy_farmhash.c \
+					 proxy_modulo_dist.c \
 					 proxy_ustats.c \
 					 proxy_ratelim.c \
 					 proxy_jump_hash.c proxy_request.c \
@@ -71,6 +73,7 @@ memcached_SOURCES += proto_proxy.c proto_proxy.h vendor/mcmc/mcmc.h \
 					 proxy_internal.c \
 					 proxy_tls.c proxy_tls.h \
 					 md5.c md5.h \
+					 farmhash.c farmhash.h \
 					 vendor/routelib/routelib.h
 endif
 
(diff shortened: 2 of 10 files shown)


fsaintjacques:
You can see the successful CI [here](https://github.com/fsaintjacques/memcached/actions/runs/15958415954/job/45007705423#step:5:364).

dormando:
Thanks!

Did you see this, by chance: https://github.com/memcached/memcached-proxylibs/tree/main/src/ketama - the hash and distribution functions are properly modular. I can readily accept new modules for hash/dist on that repository. I'm aiming to keep a minimal number of internal ones because that has lead to an explosion of options in the past and confuses users. Ones that get a lot of use can be promoted to internal over time.

fsaintjacques:
> Did you see this, by chance: https://github.com/memcached/memcached-proxylibs/tree/main/src/ketama - the hash and distribution functions are properly modular. I can readily accept new modules for hash/dist on that repository. I'm aiming to keep a minimal number of internal ones because that has lead to an explosion of options in the past and confuses users. Ones that get a lot of use can be promoted to internal over time.

I did look at the proxylibs repository (including the ketama facility). The problem with it is the lack of trivial integration with the build system and tests. If I have to setup a meta build in order to glue custom shared library on top of losing any integration test? I'll be blunt, it's simply ain't going to cut it. I'm not going to waste time, I'll fork and add it to the my tree.

What would change my opinion is if the proxylibs were integrated in the main build (git-submodule + ./configure flag?), then this would be a no brainer. 

By the way, the proxy facility is fantastic, it is extensible, clean and helps me avoid mcrouter. But in order to use it, I have to customize it and it requires touching C for my specific use case.

dormando:
what's your intent here, long term?

fsaintjacques:
My intent is to ship this ASAP while avoiding a custom fork of memcached that will drift over time.

dormando:
You're the first person with a non-proprietary hash/dist module, so the process for handling those modules hasn't been fleshed out yet. I'm happy to work on this process and excited to onboard more users, but the way you've stated your problem makes me anxious. If you'll fork vs any kind of compromise it's a likely loss for me to work on this.

We're very heavily vendored which made me stay away from git submodules; I don't want to inflict that on downstream distributions and companies. The compromise I made was a vendoring system that fetches an extra blob to unravel and that seems to be fine with people so far. The vendoring lets me revise the "semi external" libraries, like routelib, then periodically resync to the vendored blob, without blowing up the git history of the main tree. It's also trivial for people to make modifications to these semi-included modules without having to then fork the main repo with them. There's a huge long tail of companies who've made one-off modifications to hash/dist systems, which is why that particular system ended up modular.

There's also a lua-based test suite that's been brewing in the proxylibs repository, which is used for routelib (and some users have forked it for internal usage). Long term this may be used to replace the perl test suite for various reasons. I would likely move this around in the proxylibs repo and flesh it out a bit more for CI usage.

If you're up for letting me take a whack at narrowing the gap here some I can try to make some time for it.

dormando:
@fsaintjacques Hi! I'm wondering where you landed with this project. I have some time this week to revisit the hash/dist plugin setup and I didn't hear back about what you needed here.

Thanks!

PR #1234, opened 2025-07-10
core: fix pipelined set failures on pending reads

dormando:
- IO objects are enqueued per-worker-thread since a refactor in ec1fb560
- If a set payload is being read off the network but the read gets EAGAIN, _and_ there are already pending IO objects on the connection, the worker thread will execute the pending IOs
- The worker resume code did not check for conn_nread state and would resume the connection's state machine if those IO's completed before the conn_nread state completes.
- The connection state is now corrupt and will both leak memory and throw parsing errors.

This is a one-line fix to avoid resuming the state machine if we were in conn_nread state, as well as avoiding finalizing a connection from the conn_closing state if there are pending suspended responses.

TODO:
- [x] Complete and add unit tests to the PR.
- [x] Double check the state machine graph for any other error handling cases coming from mixed workloads (gets + sets)

Have some minimized unit tests that all fail without fix and run with fix. Still mapping the state machine for any other potential issues but not seeing anything yet. Nearly all intermediate states lead to `conn_closing` on error which is covered by this patch.

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -551,6 +551,8 @@ void conn_worker_readd(conn *c) {
             // any recursion here.
             event_active(&c->event, 0, 0);
             break;
+        case conn_nread:
+            // ran IO queue while waiting for set payload.
         case conn_write:
         case conn_mwrite:
         case conn_read:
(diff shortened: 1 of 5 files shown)


dormando:
didn't find any other code paths (all roads lead to `conn_closing`) but I'll check one more time before merging this.

Issue #799, opened 2021-06-27
arm32v6 always (and arm32v7 sometimes(?)) fails unit test "low TTL item should be gone" / t/restart.t


grooverdan:
**Describe the bug**

Unit test failures on https://doi-janky.infosiftr.net/job/multiarch/job/arm32v6/job/memcached/ are constant for the emcached:1.6.9-alpine build preventing the creation of the docker image.

Arm32v7 can fail the same way
https://doi-janky.infosiftr.net/job/multiarch/job/arm32v7/job/memcached/432/execution/node/78/log/
```
./testapp
....

[91m# Load a couple chunked items
[0mt/metaget.t ................. ok
[91m# Data that should expire while stopped.
[0m[91m# killed, waiting
[0m[91mGracefully stopping
[0m[91m# reconnected
[0mt/lru-maintainer.t .......... ok
[91m# low TTL item should be gone

#   Failed test 'low2 == 'mo''
#   at /usr/src/memcached/t/lib/MemcachedTest.pm line 97.
#          got: 'END
# '
#     expected: 'VALUE low2 0 2
# mo
# END
# '
[0m[91m# Looks like you failed 1 test of 93.
[0mt/restart.t ................. 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/93 subtests 
t/issue_68.t ................ ok
[91m# Some chunked item tests
[0mt/chunked-extstore.t ........ ok
t/binary.t .................. ok
t/getset.t .................. ok
t/chunked-items.t ........... ok
t/lru-crawler.t ............. ok
t/binary-extstore.t ......... ok

Test Summary Report
-------------------
t/restart.t               (Wstat: 256 Tests: 93 Failed: 1)
  Failed test:  54
  Non-zero exit status: 1
Files=80, Tests=175266, 91 wallclock secs (36.82 usr  1.90 sys + 95.62 cusr 28.70 csys = 163.04 CPU)
Result: FAIL
```

**To Reproduce**

Per build logs linked.

**System Information**
 - OS/Distro: Linux/alpine
 - Version of OS/distro: 3.14
 - Version of memcached: 1.6.9
 - Hardware detail: arm32v6, arm32v7

**Detail (please include!)**

I hope you see something in the logs in the links. I don't have the hardware to test or extract more detail from this test failure unfortunately.

dormando:
[...]

PR #1237, opened 2025-07-23
Fix typecast on setting stop_time during restart

tianon:
Fixes https://github.com/memcached/memcached/issues/799
Closes https://github.com/memcached/memcached/issues/1220

See https://github.com/memcached/memcached/issues/1220#issuecomment-3109594827 and following comments for testing that led to this discovery :partying_face:

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -4455,7 +4455,7 @@ static int _mc_meta_save_cb(const char *tag, void *ctx, void *data) {
     {
         struct timeval tv;
         gettimeofday(&tv, NULL);
-        restart_set_kv(ctx, "stop_time", "%lu", tv.tv_sec);
+        restart_set_kv(ctx, "stop_time", "%llu", (unsigned long long) tv.tv_sec);
     }
 
     // Might as well just fetch the next CAS value to use than tightly


memcached.c
```diff
@@ -4455,7 +4455,7 @@ static int _mc_meta_save_cb(const char *tag, void *ctx, void *data) {
     {
         struct timeval tv;
         gettimeofday(&tv, NULL);
-        restart_set_kv(ctx, "stop_time", "%lu", tv.tv_sec);
+        restart_set_kv(ctx, "stop_time", "%llu", (unsigned long long) tv.tv_sec);
```
tianon, line 4458: For reference, I've updated this to match the similar typecast just a few lines up (one line outside the default GitHub context):

```c
    restart_set_kv(ctx, "process_started", "%llu", (unsigned long long) process_started);
```

dormando:
Thanks! I'm going to evaluate the surrounding code but hopefully this is all that's needed.

dormando:
There's some type jank around stop time but I'm not sure I would change it right now. We're serializing it via uint64, but then parsing it back in as int64, so we can check for underflow.

It might be more appropriate to dump as int64 so the types are consistent all the way through... but honestly I'd prefer to leave it as is for now and maybe look at things again holistically later.

The tests should also be explicitly checking for if the system reused memory or not, since in your failure cases the resume actually failed but the test is whining about a TTL'ed item missing... which lead me down the wrong path of thinking it was a slow system problem.

These are notes for me though; this PR seems fine!

PR #1238, opened 2025-07-25
proxy mutator minor improvements

dormando:
- lets mutator use results from mcp.internal calls.
- if flagcopy's index is a nil, does not set the flag at all.
- adds very limited flagcopyall that can copy the result line from a result object. can't do requests or selectively exempt flags.


dormando:
added a few more improvements for flagcopy. wasn't handling nil arg right (was adding token). also does boolean now (true for add flag, false/nil for skip)

PR #1243, opened 2025-09-29
Reload TLS certificates on SIGHUP

amatveev-cf:
SIGHUP is the standard way of forcing daemons to reload their configuration, but memcached only reloads TLS certificates in response to the `refresh_certs` command, which requires a lot more work to issue. When implementing automatic certificate updates, it's a lot more straightforward to follow the standard signal-based approach.


dormando:
Thanks!

I'll need to check on something; the user who wrote/commissioned the TLS subsystem didn't want to refresh the certificates on a SIGHUP for some reason.

amatveev-cf:
Hey!
Any updates? I'm ok with making this feature optional, either at runtime or at compile time, if there are any doubts

dormando:
Thanks for the ping! I never heard back from the user. I'll take another look soon, but I've been overloaded.

dormando:
merged for the next release. thanks and sorry for the long delay.

PR #1244, opened 2025-10-01
meta: add request W flag for CAS LWW semantics

dormando:
If W flag is present in ms, change C[cas] behavior: If MISS, allow write
If HIT, allow write if C[cas] is HIGHER than current item CAS.

Allows using timestamp or version for CAS value.

Notes for storing extra data in the CAS value while using this method:
- `E` flag also exists for "use this value as the new CAS entry"
- If you want to reserve some bits for an HLC or similar:
  - Take your timestamp, mask off N bits from the end.
  - OR in your new data
  - Send this value as your `E` timestamp.
- When sending the CAS with the `C` flag for comparsion:
  - Ensure the final N bits are all 1's
  - Send this as the `C` value instead of the original CAS so we are comparing the timestamp portion and not the extra data
  - (I'm forgetting something here and will fill in later once I find the note; one of the values needs to be +/- 1 I think)

TODO:
- [ ] doc/protocol update
- [ ] audit ma and md use cases (specifically md for tombstone creation)
- [ ] internal code cleanup (`store_item` is too confusing to use now)
- [ ] tests
- [ ] document in this thread (and maybe also doc/protocol) HLC/etc use cases.
- [ ] also add docs to wiki. (note pending release)

### DIFF
--- a/memcached.c
+++ b/memcached.c
@@ -1544,7 +1544,7 @@ static int _store_item_copy_data(int comm, item *old_it, item *new_it, item *add
  *
  * Returns the state of storage.
  */
-enum store_item_type do_store_item(item *it, int comm, LIBEVENT_THREAD *t, const uint32_t hv, int *nbytes, uint64_t *cas, uint64_t cas_in, bool cas_stale) {
+enum store_item_type do_store_item(item *it, int comm, LIBEVENT_THREAD *t, const uint32_t hv, int *nbytes, uint64_t *cas, uint64_t cas_in, bool cas_stale, bool cas_lww) {
     char *key = ITEM_key(it);
     item *old_it = do_item_get(key, it->nkey, hv, t, DONT_UPDATE);
     enum store_item_type stored = NOT_STORED;
@@ -1568,6 +1568,8 @@ enum store_item_type do_store_item(item *it, int comm, LIBEVENT_THREAD *t, const
             cas_res = CAS_MATCH;
         } else if (cas_stale && it_cas < old_cas) {
             cas_res = CAS_STALE;
+        } else if (cas_lww && it_cas > old_cas) {
+            cas_res = CAS_MATCH;
         } else {
             cas_res = CAS_BADVAL;
         }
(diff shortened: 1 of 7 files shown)


memcached.c
```diff
@@ -1684,11 +1686,16 @@ enum store_item_type do_store_item(item *it, int comm, LIBEVENT_THREAD *t, const
                 do_store = true;
                 break;
             case NREAD_CAS:
-                // LRU expired
-                stored = NOT_FOUND;
-                pthread_mutex_lock(&t->stats.mutex);
-                t->stats.cas_misses++;
-                pthread_mutex_unlock(&t->stats.mutex);
+                // Item missing.
```
dormando, line 1689: The `cas_res = CAS_MISS` above can be removed. Doesn't look like cas_res is used inside or outside of that block.

dormando:
I didn't mean to sit on this; I think it's relatively complete feature wise and have been focusing on some design work in the background. I'll get another round on this next week.

dormando:
never heard back from anyone. going to do a self-review.

dormando:
going to hold this for the next release... should dedicate more time than I have right now to writing comprehensive tests, and the code is entirely missing for ma/md.

dormando:
Note to self:
- md with `x` and `W1234` works as expected against an existing item
- do we need to autovivify a tombstone if delete comes in against a miss?

For `ma`:
- Update `do_add_delta` for the lww semantic
- Looks like `E` may not be plumbed into `do_add_delta`; two fixes to do here. the `E` override is applied outside of the delta function. maybe could do as different commit for cleanup.

dormando:
Tossed around the idea for a few hours yesterday:
 - md with `x` and `W` against a miss should autovivify
 - check the rules for CAS inheritence in that case so either C or E is used.
 - ma shouldn't support `W` at all.

I have future plans for ma and `W` has no real use case. So avoiding it will avoid being boxed in.

One open question on having ma with `W` causing a hard-fail. Typically we ignore unused flags but since `W` does something in some commands and not others it might make more sense to error out there.

dormando:
fixed a few things with `md`... should `md x` always autovivify? That would be new behavior for an existing combination which is usually a non-starter.

dormando:
might have to punt this to next week. running on fumes and worried I'll make a bad decision.

In general the problem is narrowed (no `ma`, `md` can create tombstones, etc) and the next step is to sit here and write tests that pass/fail and ask the right questions.

dormando:
Bleh:
- Rewire all callers to `store_item`: remove `NREAD_CAS` case, instead adopting that behavior when an input CAS is supplied. This fixes an issue where the external command semantics (ie; replace) are lost if CAS is supplied.
- Optionally: stop currying input cas from the input item (?!?!?!?)
- md `W` implies `x` or not? Leaning toward no to preserve flexibility but there are probably few/no cases requireing it.
- `W` stamps value of `C` as new CAS, unless `E` supplied. I can't come up with a reason to still use E in this case, but screw it.
- Some random bugs to fix.

PR #1246, opened 2025-10-09
proto: add a mult command to multiply stored values.

simek-m:


simek-m:
Sorry, I made a mistake and didn't mean to open here.

PR #1249, opened 2026-01-16
Use SSLv23 method when TLSv1.3 is unsupported (e.g., macOS)

ibujna:
New macOS Tahoe versions ship with OpenSSL 3.6.0 builds that disable TLSv1.3. Update the SSL context creation to use SSL_version => 'SSLv23' after test fail with TLSv1.3.


dormando:
@ibujna It'll probably take me a while to work through this big batch of PR's, so hang tight please. I'll leave comments as to style changes and hopefully you'll be receptive to them.

Since this is a lot at once, may I ask what this is coming from? Looks like you might work for Oracle?

ibujna:
Yes, I working at Oracle.

dormando:
merged, thanks!

PR #1250, opened 2026-01-16
Fix test compatibility on IPv6-only systems.

ibujna:
Check IPv6-bound port when IPv4 is not in use.


dormando:
looks fine. just need to run it.

dormando:
moving through CI, but I don't expect problems. thanks!

PR #1251, opened 2026-01-16
Improve parallel test execution, RPM build, and test configuration

ibujna:
- Enable running tests in parallel using the Makefile variable PARALLEL. Example usage: PARALLEL=5, PARALLEL=-j5, PARALLEL=-jobs=5.

- Add environment variable TEST_CMD to print the memcached command executed during tests.

- Configure unit tests to skip starting memcached with SSL or UDP when not required.

- Fix parallel test execution:
    1. Each test allocates its own port to avoid conflicts.
    2. Disable SSL or UDP in tests when not used to prevent parallel test failures.


dormando:
Parts of this are easy to take but I'm not sure about half of it... If you feel up to splitting it up into a few more PR's that would help. Or else I guess eventually I'll do it.

PR #1252, opened 2026-01-16
Refactor TLS implementation

ibujna:
Unify TLS functionality between memcached and proxy Extend proxy with full mTLS support
Fix memcached -h help text and complete description of the listen interface parameter (-l <...>) Extend proxy Lua configuration to allow mTLS certificate setup per connection Add tests for all possible TLS connection variants with a memcached server Extend TLS tests to cover invalid (expired) certificates Add tests for all possible TLS connection variants between memcached server and proxy Add memtier-based load tests for all TLS connection types between memcached server and proxy


dormando:
Thanks for this! It's a huge change and I have to be very careful with the proxy code for my customers' sake. I'll review this carefully as I get time. It would be great to get the rest of TLS options supported for the proxy code.

However I took a quick read and I can't tell if you intend this to get merged as is, or are looking for some kind of feedback first? There are some "smells" here:
- Code style doesn't match internal C/perl code (spacing, for vs foreach, etc)
- Relies on external memtier app for some of the testing, which is neither the tooling we use (mcshredder) nor something we rely on or ship anywhere else in the code.
- Some code is `ifdef 0`'ed.
- Similar small things throughout.

Again I didn't examine it deeply but it has enough red flags that I will stop and first ask if you can clarify your intentions here.

Thanks!

dormando:
@ibujna attempting to ping again. Haven't heard back in a week.

ibujna:
Sorry I was on vacation. 

I implemented full mTLS on the proxy so that the proxy as memcached client can verifies the certificate.
During development, I discovered several issues in the code and tests.
I created separate pull requests for these changes.

Yes, I work at Oracle.

The proxy configuration is able to communicate over all types: **notls**, **btls**, and **mtls**. These TLS versions come from the original code, with memcached and the proxy using the same functionality. The new **mtls2** version is based on **mtls** and additionally includes client-side certificate verification.

To maintain consistent code style, I would recommend using the Tidy package for Perl and gnu-indent for C.

I used **Memtier** only as a load test to compare connection speeds. Without the environment variable that sets the path to the Memtier binary, this test is skipped. I will look into **mcshredder** and try to adapt this test for it.

I also still want to add comments and some descriptions of the functionality in t/lib/MemcachedTest.pm.

Could you clarify what you mean by this suggestion: **It would be great to get the rest of TLS options supported for the proxy code?**

dormando:
> To maintain consistent code style, I would recommend using the Tidy package for Perl and gnu-indent for C.

As a submitter to an upstream project you need to conform to their code style. We have an informal style, but please "make it look like the surrounding code" - I have to maintain this, mostly by myself. I'm not going to tidy/lint the entire codebase to make it compatible with your PR.

> I used **Memtier** only as a load test to compare connection speeds. Without the environment variable that sets the path to the Memtier binary, this test is skipped. I will look into **mcshredder** and try to adapt this test for it.

Thanks! I probably still wouldn't include that in the main repo, but I do have a separate repo with benchmark tests that I run. I have reorganized the code a few times but it is still probably difficult to work with, so I apologize: https://github.com/memcached/shredders

> I also still want to add comments and some descriptions of the functionality in t/lib/MemcachedTest.pm.

Ok.

> Could you clarify what you mean by this suggestion: **It would be great to get the rest of TLS options supported for the proxy code?**

Sorry that was not a suggestion. It was an affirmation that these are desireable features and I would like to take this code.

However I should caveat that I need to be extremely careful with anything touching TLS; for performance and stability reasons. I just spent a month last year cleaning up the bugs in the original TLS code and users are very sensitive to problems with it.

dormando:
@ibujna Any thoughts on cleaning this up at all? It's been over a month since my last comment.

PR #1254, opened 2026-01-16
Account for absent 'ssl_proto_errors' in stats during SSL tests

ibujna:


dormando:
Merged, thanks!

I have a CI that's supposed to be running all of the SSL tests but I guess it's missing the `SSL_TEST=1` to force the full suite... I'll fix that next time I do a round of updates there.

PR #1255, opened 2026-01-16
SSL tests run over TCP using IO::Socket::SSL, which typically inherits

ibujna:
from IO::Socket::IP.

When IO::Socket::IP is not available, IO::Socket::SSL falls back to IO::Socket::INET. The latter returns undef (instead of an empty string) upon a server disconnect.


dormando:
Little bizarre. I'm delaying this until the next release though.

I feel like it'll be better to adjust the tests to allow '' or undef instead of trying to duck type by what modules are available, in case this changes again.

PR #1257, opened 2026-01-16
Exclude doc/xml2rfc/rfc2629.dtd, queue.h and t/authfile2

ibujna:
from modifications by devtools/clean-whitespace.pl


dormando:
Thanks! No idea why I didn't do this. always found it while in the middle of some branch and just didn't fix, heh.

ibujna:
It’s frustrating that changing the code also changes files that should remain untouched. 
So I try to fix it this call.

dormando:
Merged up finally, thanks!

PR #1258, opened 2026-01-16
Fix failing proxy*.t tests on some systems like OL8

ibujna:
Prevent crashes when memcached reloads and the authfile is not set up.


dormando:
Merged, thanks!

PR #1259, opened 2026-01-16
tests: add sleep to catch reload timeout in clock_handler

ibujna:
The reload timeout in clock_handler() is set to 1 second. Add a sleep to reliably trigger the event timeout during tests.


dormando:
Think you created this PR twice? Looks the same as #1256 ?

PR #1270, opened 2026-02-27
fix: storage_compact_readback checks wrong refcount

Sunjeet:
### Access pattern
The key is being read 10-13K times per second, not being written at this time, meanwhile compaction is high due to high disk util and max_frag at 0.99 (but also on another node at similar access pattern and disk util but max_frag is 0.97). item_age at 0. 

### Root cause
When refcount is 3+, the else branch (line 1009) creates a new HDR via item_replace. The old HDR (held by the pending IO) gets unlinked by item_replace. When the  reader's IO later fails and the miss handler calls item_unlink(p->hdr_it), ITEM_LINKED is already cleared → no-op. The new HDR in the hash table survives.  the HDR points to the new valid page. But then the racing reader's miss handler does: item_unlink(p->hdr_it)

Here's the exact sequence that kills this key:                                                                                                                     
   
  1. Worker: do_item_get(key) → finds HDR, refcount 1→2                                                                                                              
  2. Worker: storage_get_item() → copies page_id=P_old, page_version=V_old into io_pending. p->hdr_it = hdr_it
  3. Worker: Releases item_lock. IO not yet dispatched.
  4. Compactor: Acquires item_lock for this key
  5. Compactor: refcount_incr(hdr_it) → refcount now 3
  6. Compactor: Checks it->refcount == 2 (disk copy) → TRUE (always 2 for cold LRU items)
  7. Compactor: Updates HDR in-place: hdr->page_id=P_new, hdr->page_version=V_new
  8. Compactor: Releases item_lock. Continues processing other items on page P_old.
  9. Compactor: Finishes page → extstore_close_page(P_old) → page P_old is now closed
  10. IO thread: Tries pread(page=P_old, version=V_old) → page closed → ret = -2
  11. IO callback: _storage_get_item_cb: ret < 1 → p->miss = true
  12. Worker: recache_or_free: item_unlink(p->hdr_it) → removes the LIVE HDR (which now has valid P_new/V_new data) from the hash table
  13. Key permanently gone. miss_from_extstore++

Screenshot 2026-02-27 at 4 47 26 AM

### DIFF
--- a/storage.c
+++ b/storage.c
@@ -1000,7 +1000,7 @@ static void storage_compact_readback(void *storage, logger *l,
 
                 if (do_update) {
                     bool rescued = false;
-                    if (it->refcount == 2) {
+                    if (hdr_it->refcount == 2) {
                         hdr->page_version = io.page_version;
                         hdr->page_id = io.page_id;
                         hdr->offset = io.offset;


dormando:
Thanks!

This looks right, but I need to run this through some tests before merging, as now I'm not confident the realloc rescue code is well tested.

dormando:
Notes to self for tests:

- start from extstore compact tests
- debugitem ref [key] to leak a ref on some items
- run compaction

There's no specific counter for "realloc during compaction" - so either need to add one or add some debug logging so it can be verified.

Think I just eyeballed this compaction header swap as a `NEED_ALIGN` fault as well.

edit: I'll do the basic tests and step through it in a debugger so I can hand check the refs.

dormando:
Just checked git history and it's been wrong since the initial commit in 2017... Was curious if I broke it while adding the rescue code.

dormando:
The 2023 commit where I add the rescue code has a funny commit message though, detailing:
```
    - Be running with item_age=0 or a very low number
    - Constantly causing defragmentation
    - Get very unlucky during defragmentation.
```
yup... if only I did the fix right :P

dormando:
I'm going to think about the test plan for a bit and come back to this. will hopefully merge today.

dormando:
just doing a maintenance run this week... I think for this one the important bits are:
- The bugfix, obviously. a one-liner.
- Actually testing that the other code path works is of course way more work than pointing at the bad line, and I get to do that now...
- `miss_from_extstore` has a high cause cardinality but I don't want ten counters...

My plan is similar to what I did for the page mover. There'll be the miss counter and then a "last miss reason" stat with words in it. On top of that a log line sent to `watch evictions` stream.

dormando:
Got a consistent test and stepped through the realloc phase in a debugger... Trying to flesh out tests for all the possible issues with realloc.

Also figuring out how to add extra detail to "rescues" (some rewrite header, some realloc) and "lost" (some because we can't write to extstore, some for OOM). So far I'm adding info the `compact_end` log line but I'm not convinced that's all I should do.

Splitting up miss_from_extstore is even more involved. I'll plan it out but might split that into its own change so I can get this basic fix out sooner.

dormando:
Okay. Gone as far as I'm going to for this release. I merged this commit manually after rewriting the commit message to be more informational.

My testing work is the commit after that in `next`, which is now off to the CI cluster.

I have a few more small changes I want to merge and I think just one more regression to look at, then it has to run through the load tests before releasing (probably running overnight).

dormando:
eugh. got a bunch more prio PR's/issues actually... hopefully they don't take that long to verify/fix unlike these last two.

dormando:
Revisiting compact_lost:
- For each compacting item, we ask extstore if we can write
- We loop, waiting up to 10ms for free bufferspace to write to _per item_
- I bumped this to 100ms with my test change.

We don't have dual write buffers per bucket, so while a write buffer is out to write to disk extstore basically fails to write internally and is supposed to back off and wait... which isn't great in this compaction portion but was certainly the right idea at the time it was written.

Trying to come up with a minimal change to lower validation effort... So far my best idea is to limit the total waiting per readback and then set that to 1s+, IE:
- for each item, we wait up to 1s
- Or if across all items in this readback chunk we've waited longer than 1s, be less patient (minimum 1 or 10ms).
- It's a lot more work... but a starttime/runtime option for the total patience is probably a good idea?

Generally once we get a write buffer back all the writes become instantaneous and won't block until the write buffer fills again... so this long wait should be sufficient. We can't block forever for various reasons.

I'll see what I can do here without delaying the release too long. Still chewing through PR/issue backlog too.

dormando:
Sidenote: We did initially have up to N write buffers but I cut them as there was a hard to find bug, and haven't revisited since then. So I'm not fundamentally against A/B bucket write buffers for extstore I just couldn't justify the complexity and nobody's pushed on it since 2017.

dormando:
Went ahead with a 5s total wait per readback change. That should dramatically lower the amount of compact lost errors while avoiding hanging the system, and avoiding a longer devcycle for the time being.

dormando:
Released in 1.6.41 after finally passing CI

PR #1274, opened 2026-03-21
Add a warning message about a missing Lua application and include instructions on how to download and install it.

ibujna:
Add comment explaining what to do if attempting to compile Lua when Lua is not installed.


dormando:
Released in 1.6.42, thanks!

PR #1277, opened 2026-03-21
Add environment variable TEST_CMD to print the memcached command executed during tests

ibujna:


dormando:
not using `T_MEMD_EXTERNAL="1"`? it's arguably more useful.

PR #1278, opened 2026-03-21
Fix timing side-channel in SASL password database authentication

kingroryg:
- `sasl_server_userdb_checkpass()` breaks out of the password file loop early when a valid username is found, creating a measurable timing difference (~2.7ms) between valid and invalid usernames, enabling remote username enumeration against deployments compiled with -`-enable-sasl-pwdb`.
  - The password comparison uses `memcmp()`, which is not constant-time and could leak password bytes via timing analysis.
  - **Fix**: remove the early break so the entire file is always scanned, and replace `memcmp()` with the project's existing `safe_memcmp()`.

Test plan
  - Build with `./configure --enable-sasl --enable-sasl-pwdb && make`
  - Create a password file (`echo "admin:secret" > /tmp/pwfile`) and start memcached with `MEMCACHED_SASL_PWDB=/tmp/pwfile ./memcached -S`
  - Verify valid credentials authenticate successfully
  - Verify invalid credentials are rejected
  - Measure response times for valid vs invalid usernames and confirm the timing difference is eliminated

### DIFF
--- a/sasl_defs.c
+++ b/sasl_defs.c
@@ -71,19 +71,18 @@ static int sasl_server_userdb_checkpass(sasl_conn_t *conn,
     char buffer[MAX_ENTRY_LEN];
     bool ok = false;
 
-    while ((fgets(buffer, sizeof(buffer), pwfile)) != NULL) {
-        if (memcmp(user, buffer, unmlen) == 0 && buffer[unmlen] == ':') {
-            /* This is the correct user */
-            ++unmlen;
-            if (memcmp(pass, buffer + unmlen, passlen) == 0 &&
-                (buffer[unmlen + passlen] == ':' || /* Additional tokens */
-                 buffer[unmlen + passlen] == '\n' || /* end of line */
-                 buffer[unmlen + passlen] == '\r'|| /* dos format? */
-                 buffer[unmlen + passlen] == '\0')) { /* line truncated */
+    while (1) {
+        memset(buffer, 0, sizeof(buffer));
+        if (fgets(buffer, sizeof(buffer), pwfile) == NULL)
+            break;
+        if (safe_memcmp(user, buffer, unmlen) && buffer[unmlen] == ':') {
+            if (safe_memcmp(pass, buffer + unmlen + 1, passlen) &&
+                (buffer[unmlen + 1 + passlen] == ':' ||
+                 buffer[unmlen + 1 + passlen] == '\n' ||
+                 buffer[unmlen + 1 + passlen] == '\r' ||
+                 buffer[unmlen + 1 + passlen] == '\0')) {
                 ok = true;
             }
-
-            break;
         }
     }
     (void)fclose(pwfile);


sasl_defs.c
```diff
@@ -70,20 +70,38 @@ static int sasl_server_userdb_checkpass(sasl_conn_t *conn,
...
+                while (stored_pass[stored_passlen] != ':' &&
+                       stored_pass[stored_passlen] != '\n' &&
+                       stored_pass[stored_passlen] != '\r' &&
+                       stored_pass[stored_passlen] != '\0') {
+                    stored_passlen++;
+                }
+
+                /* Constant-time password comparison using safe_memcmp
+                 * to prevent leaking password bytes via timing. */
+                if (passlen == stored_passlen &&
```
dormando: Doesn't this still leak the timing by being slightly faster if the password length matches? or is that not measurable? :P 

Looks like the authfile code that calls `safe_memcmp` does this as well... checks that the len's all match first.

sasl_defs.c
```diff
@@ -70,20 +70,38 @@ static int sasl_server_userdb_checkpass(sasl_conn_t *conn,
 
     char buffer[MAX_ENTRY_LEN];
     bool ok = false;
+    bool found = false;
+
+    /* Zero the buffer so that password comparisons against short lines
+     * never read uninitialized stack data. */
+    memset(buffer, 0, sizeof(buffer));
```
dormando: Feels like the move is to memset this buffer back to 0 before every test, then don't calculate the length of the stored password. Then post-safe_memcmp test that the buffer just past the supplied password is correct, like the original code did?

unless I'm wrong this should be a three line change:

- clear buffer to 0 just under while
- memcmp -> safe_memcmp on the passlen
- remove the break

Doesn't seem to fix timing attacks around discovering users or anything but I'm not convinced we need to be that careful.

dormando (changes requested):
Left some comments. If you're still around, some thoughts would be nice :P Else I'll just make my own PR with these changes.

sasl_defs.c
dormando: I guess four lines, adding safe_memcmp to the user check.

dormando:
@kingroryg any thoughts? My comment still seems like the better fix here. so I'll probably do that shortly.

kingroryg:
@dormando Ack, sending in the fixes shortly.

dormando:
Shorter, cleaner, and actually fixes the problem :) thanks!

kingroryg:
One behavioral note: since we no longer break on the first username match, if a password file has duplicate usernames the last entry wins (rather than the first). I think this is fine, duplicate entries in the `pwfile` is an admin misconfiguration but wanted to call it out explicitly. Let me know if you'd prefer to keep `ok = true` sticky (i.e., never reset it back to false once set).

dormando:
If the entries are duplicate they'd need to have the same password too. In that case I can't imagine it matters.

dormando:
Released in 1.6.42, thanks!

PR #1280, opened 2026-03-30
Add an option to link against system Lua installation

Fothsid:
We'd like to have this for the memcached package shipped in Fedora. Fedora Packaging Guidelines restrict usage of bundled libraries unless it is absolutely necessary.

Currently the changes add an additional configure argument `--disable-proxy-builtin-lua` that stops `vendor/lua` from being built and relies on `-llua` being in the search paths.

Thanks & let me know if any changes are in order. 🙂

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -119,10 +119,12 @@ CLEANFILES += memcached_dtrace.o memcached_debug_dtrace.o
 endif
 
 if ENABLE_PROXY
+if ! DISABLE_PROXY_BUILTIN_LUA
 memcached_LDADD += vendor/lua/src/liblua.a
 memcached_debug_LDADD += vendor/lua/src/liblua.a
 memcached_LDFLAGS += -rdynamic
 memcached_debug_LDFLAGS += -rdynamic
+endif
 endif
 
 if ENABLE_PROXY_URING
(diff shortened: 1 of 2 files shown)


dormando:
This is interesting! I'm curious what you do for Redis and similar?

I ship the lua by default for a few main reasons:
- Lua's incompatibilities between versions (not just API but major internal changes that need to be validated)
- Similarly I'm evaluating upgrading from 5.4 to 5.5, but would like to avoid lots of internal ifdef's + having to CI and benchmark 5.4 and 5.5 for every release.
- I wanted to reserve the right to patch it (currently exploring some GC changes that I might have to carry as a forked patch)

I'd like to support this change but this messes with all of this; I can't validate that the lua you ship doesn't cause problems for end users and I can't patch it to work around limitations in the language.

Thoughts?

Fothsid:
Hi, those are valid concerns. After some discussion with colleagues, it would probably be okay for us to keep it bundled, as there is a bunch of other software bundling specific versions of Lua with them. Feel free to close the PR and sorry for the noise.

(As for Redis, it is no longer shipped in Fedora due to changed licensing, and is replaced with Valkey, which in itself has a bundled version of Lua 5.1.)

dormando:
Thanks for taking careful consideration about this. Lua's a tiny little weirdo that shows up in a lot of these cases I imagine.

PR #1289, opened 2026-05-04
update `meta set` protocol to remove obsolete flag

### LINKED ISSUE #1288: `meta set` protocol doc still references now obsolete flag
The latest protocol doc still calls for an `S` flag (which [doesn't exist](https://github.com/memcached/memcached/blob/3824603965ecbd38e552a725dbfc4ff38ab40cb2/doc/protocol.txt#L740-L752)) to supply the size of the `<data block>` as seen here:
https://github.com/memcached/memcached/blob/3824603965ecbd38e552a725dbfc4ff38ab40cb2/doc/protocol.txt#L716-L718

However, latest version of `meta set` commands specifies that data length will be provided after `<key>` without a flag:
https://github.com/memcached/memcached/blob/3824603965ecbd38e552a725dbfc4ff38ab40cb2/doc/protocol.txt#L703

This appears to be an artifact from pre-`1.6.10` where this was true and the lines can be omitted.

awfulstew:
fixes https://github.com/memcached/memcached/issues/1288

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -713,9 +713,8 @@ After this line, the client sends the data block:
 
 <data block>\r\n
 
-- <data block> is a chunk of arbitrary 8-bit data of length supplied by an 'S'
-  flag and token from the request line. If no 'S' flag is supplied the data
-  is assumed to be 0 length.
+- <data block> is a chunk of arbitrary 8-bit data of length <datalen>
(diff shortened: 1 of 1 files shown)


dormando:
Thanks!

dormando:
pushed to the `next` branch pending release.

PR #1291, opened 2026-05-21
SPARC traps on several memory accesses that are tolerated on x86.

psumbera:
MurmurHash3_x86_32() may be called with unaligned key pointers, so load 32-bit blocks byte-by-byte while preserving the little-endian layout expected by the x86 MurmurHash3 variant.

Item client flags are stored after the item key and can also be unaligned. Use the existing FLAGS_CONV path, backed by memcpy(), instead of directly dereferencing ITEM_suffix() as client_flags_t.

Logger payloads are overlaid with typed records, including records with 64-bit fields. Align the flexible payload array to uint64_t so those overlays are safe on strict-alignment platforms.

### DIFF
--- a/logger.h
+++ b/logger.h
@@ -164,6 +164,9 @@ struct _logentry {
     struct timeval tv; /* not monotonic! */
     int size;
     union {
+#ifdef NEED_ALIGN
+        uint64_t align;
+#endif
         char end;
     } data[];
 };
--- a/memcached.c
+++ b/memcached.c
@@ -4071,7 +4071,12 @@ static void usage(void) {
            "                          forcefully killing LRU tail item.\n"
            "                          disabled by default; very dangerous option.\n"
            "   - hash_algorithm:      the hash table algorithm\n"
-           "                          default is murmur3 hash. options: jenkins, murmur3, xxh3\n"
+#ifdef NEED_ALIGN
+           "                          default is xxh3 hash on NEED_ALIGN builds.\n"
+#else
+           "                          default is murmur3 hash.\n"
+#endif
+           "                          options: jenkins, murmur3, xxh3\n"
            "   - no_lru_crawler:      disable LRU Crawler background thread.\n"
            "   - lru_crawler_sleep:   microseconds to sleep between items\n"
            "                          default is %d.\n"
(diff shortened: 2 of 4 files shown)


dormando:
Thanks! I think we'll have to re-arrange this a bit. I no longer have access to a sparc machine so testing is difficult.

Can you tell me if `NEED_ALIGN` is getting set in config.h?

Also, if you set the hash type to xxhash, is that still broken?

psumbera:
> Thanks! I think we'll have to re-arrange this a bit. I no longer have access to a sparc machine so testing is difficult.

There is SPARC machine (requires registration): https://portal.cfarm.net/machines/list/ (sparc: 216)
 
> Can you tell me if `NEED_ALIGN` is getting set in config.h?

Yes. On the SPARC build:
```
  #define NEED_ALIGN 1
```
> Also, if you set the hash type to xxhash, is that still broken?

For xxhash: the option is hash_algorithm=xxh3. That avoids the Murmur3-specific unaligned block load, but it does not fix the whole problem. I reversed the alignment patch in a SPARC test copy and ran:
```
  ./memcached-debug -p 19676 -U 0 -o hash_algorithm=xxh3
  set ab 123 0 1
  mg ab f v
```
That still dies with Bus Error, exit status 138. The key ab makes the stored client flags unaligned. So xxh3 avoids only the murmur3_hash.c part; the client_flags_t * dereference in ITEM_suffix() is  still broken independently of hash type.

With the patch applied, the same xxh3 smoke passes and reports hash_algorithm xxh3.

dormando:
Thanks! Some machines weren't getting `NEED_ALIGN` set somehow.

Can you gate the client_flags change behind `NEED_ALIGN`? As much as possible I gate behind that flag. Behind the scenes here I've been restructuring the code to not need alignment workarounds without losing performance.

I'd rather gate murmurhash instead of modifying it. I intended to swap xxhash to the default a long time ago but ... never did I guess. In simple terms that means swapping the default on `NEED_ALIGN` but being more thorough would test if murmur fails specifically I guess... Since I've never heard of it failing before now on any alignment-sensitive system? I wonder why.

dormando:
Also, the logger code should have a `NEED_ALIGN` gate on the union addition. There's already a `NEED_ALIGN` gate in the logger code to pad an entry so the next structure is aligned.

dormando:
Just noticed your push fix. I'll review this week. Thanks!

dormando:
released in 1.6.43

PR #1292, opened 2026-05-24
lru_crawler: add missing OK\r\n prefix to metadump and mgdump

### LINKED ISSUE #667: Memcached lru_crawler metadump doesn't returns END as response
We are currently working on getting all the keys in Memcached through:

lru_crawler metadump all
For this implementation, we are expecting END or OK at the end of result.

But Memcached doesn't return any response code after the end of data. It just simply returns the metadata.

lru_crawler metadump 1
key=key exp=-1 la=1588598718 cas=1 fetch=no cls=1 size=80
We are using Memcached version 1.4.33.

Is there any configuration that we need to set up for this to return a response code at the end (or) Is there any way that we can find the end of metadump result other than END or OK?

NikitaNavrotskiy:
The metadump and mgdump commands were supposed to return OK\r\n
before data, but init() was called before set_client(), so the
client buffer wasn't available yet.

### Changes

- Reordered set_client() before init() in lru_crawler_start() so
  modules with needs_client=true have the client buffer ready
  during initialization.
- Added OK\r\n output in crawler_metadump_init().
- Added OK\r\n output in crawler_mgdump_init().
- Updated metadump test to skip the OK line when counting results.

### Testing

- `make test` passes
- Manual verification

Fixes #667

### DIFF
--- a/t/lru-crawler.t
+++ b/t/lru-crawler.t
@@ -74,6 +74,7 @@ while (1) {
     my $count = 0;
     while (<$sock>) {
         last if /^(\.|END)/;
+        next if /^OK/;
         /^(key=) (\S+).*([^\r\n]+)/;
         $count++;
     }
@@ -113,6 +114,7 @@ for (1 .. 70000) {
     my $count = 0;
     while (<$sock>) {
         last if /^(\.|END)/;
+        next if /^OK/;
         if (/^key=bfoo(\S+)/) {
             ok(exists $bfoo{$1}, "found bfoo key $1 is still in test hash");
             delete $bfoo{$1};
--- a/crawler.c
+++ b/crawler.c
@@ -263,6 +263,10 @@ static void crawler_expired_eval(crawler_module_t *cm, item *search, uint32_t hv
 
 static int crawler_metadump_init(crawler_module_t *cm, void *data) {
     cm->status = 0;
+    if (cm->c.buf != NULL) {
+        memcpy(cm->c.buf, "OK\r\n", 4);
+        cm->c.bufused += 4;
+    }
     return 0;
 }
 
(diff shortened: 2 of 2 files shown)


dormando:
Thanks! I saw this last week but spaced out on commenting.

I was holding off on doing this because of uncertainty around breaking existing clients that use the feature... and still haven't done that research. I'll take a quick look this week.

@NikitaNavrotskiy One comment though: #667 is specifically about the missing END/EN response, but your PR doesn't address that so it doesn't fully fix the issue.

dormando:
@NikitaNavrotskiy reminder

NikitaNavrotskiy:
@dormando Looking at the original issue again:

- "There's no END\r\n or EN\r\n because I probably just forgot"
- "The OK\r\n that was supposed to print gets eaten"

The END/EN was already in crawler_metadump_finalize() and 
crawler_mgdump_finalize() when I started. The only missing part was the OK part, which my PR adds.

So this PR should fully address #667.

dormando:
Ok double checked. I added the `END` marker in 2017... then fixed it a bunch of times due to crashes, ouch. Their 1.4.33 version was possibly too old (2016) and I didn't catch that while first reviewing.

t/lru-crawler.t
```diff
@@ -74,6 +74,7 @@ while (1) {
     my $count = 0;
     while (<$sock>) {
         last if /^(\.|END)/;
+        next if /^OK/;
```
dormando, line 77: Since this is a test, we should actually validate that he OK is there as the first response line instead of ignoring it here :) I'll note this but just make the adjustment myself since it's been so long since you submitted this.

dormando:
released in 1.6.43

Issue #1279, opened 2026-03-25
[Bug] Multiple self-deadlocks in Memcached from re-locking non-recursive mutexes


hanqing2025:
## Summary

Dear Memcached maintainers,

During our analysis of Memcached, we found 5 self-deadlock cases that appear to share the same root cause.

In the observed cases, a thread acquires a non-recursive mutex and later re-enters a path that attempts to acquire the same mutex again before the first acquisition is released, resulting in self-deadlock.

All 5 cases are single-thread lock re-entry deadlocks rather than multi-thread lock-order cycles.

## Concrete example

One concrete case is `worker_hang_lock` in `stop_threads()`.

In the current upstream tree, `stop_threads()` acquires `worker_hang_lock` in `thread.c:217`:

- `thread.c:217` `pthread_mutex_lock(&worker_hang_lock);`

The same mutex is also acquired in the pause path:

- `thread.c:169` `pthread_mutex_lock(&worker_hang_lock);`

and worker registration also synchronizes through it:

- `thread.c:149` `pthread_mutex_lock(&worker_hang_lock);`

This lock is used to force worker threads into a blocked state during pause or shutdown coordination. If the shutdown path re-enters logic that tries to acquire `worker_hang_lock` again on the same thread before the first acquisition is released, the second acquisition blocks forever because the mutex is non-recursive.

We observed this as a self-loop deadlock associated with `stop_threads()`.

## Affected revision

Latest upstream revision analyzed:

- `3824603965ecbd38e552a725dbfc4ff38ab40cb2`

Earlier revisions may also be affected, but we have not yet determined the exact introduction point.

## Affected sites

1. `lru_crawler_lock`
   - `crawler.c:719`
   - `lru_crawler_start()` also acquires the same lock at `crawler.c:833`
   - observed trigger API: `lru_crawler_crawl`

2. `lru_locks[id]` / `lru_locks[it->slabs_clsid]`
   - `items.c:1089`
   - function: `lru_pull_tail()`

3. `slabs_lock`
   - `slabs.c:808`
   - function: `slabs_locked_callback()`

4. `t->lock`
   - `slabs_mover.c:749`
   - function: `slabs_reassign()`
   - the same lock is also taken again in nearby maintenance helpers such as `slab_maintenance_pause()` at `slabs_mover.c:761`

5. `worker_hang_lock`
   - `thread.c:217`
   - function: `stop_threads()`

## Impact

[...]

PR #1295, opened 2026-06-30
Avoid slab reassignment while holding LRU locks

hanqing2025:
## Summary

Move the `slabs_reassign()` request in `lru_pull_tail()` out of the LRU/item lock scope.

Previously, the eviction path could call `slabs_reassign()` while still operating under the LRU lock. This patch records that slab reassignment is needed, releases the LRU lock, finishes item cleanup, and then invokes `slabs_reassign()` afterward.

## Rationale

This avoids unnecessary lock nesting between the item LRU eviction path and slab reassignment logic.

The change does not alter when slab reassignment is requested:

- it still only happens when `settings.slab_automove == 2`
- it still follows the same eviction path
- eviction accounting is unchanged
- item removal behavior is unchanged

Related to #1279.

### DIFF
--- a/items.c
+++ b/items.c
@@ -1084,6 +1084,7 @@ int lru_pull_tail(const int orig_id, const int cur_lru,
     void *hold_lock = NULL;
     unsigned int move_to_lru = 0;
     uint64_t limit = 0;
+    bool do_slab_reassign = false;
 
     id |= cur_lru;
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! This might actually be a reasonable fix here.

dormando:
released in 1.6.43

PR #1297, opened 2026-06-30
seccomp: allow ppoll in worker sandbox for loongarch64

wszqkzqk:
poll() is already allowed, but on some architectures glibc's poll() wrapper is implemented via ppoll(). Without this rule the worker thread is killed by the seccomp filter as soon as it waits on I/O.

This is reproducible on LoongArch64 with --enable-seccomp, where t/lru-crawler.t fails because the worker syscall 73 ppoll is denied. ppoll is functionally equivalent to poll for memcached's purposes and does not increase the sandbox attack surface.

Tested on Arch Linux for Loong64.


dormando:
released in 1.6.43

PR #1298, opened 2026-06-30
tests: cap polling backoff in slabs-mover wait helpers

wszqkzqk:
wait_for_stat and wait_for_stat_incr used an uncapped linearly increasing sleep.

Because $to_sleep itself accumulates, the per-iteration sleep grows quadratically. In the worst case of all 500 iterations the test would sleep for roughly 58 hours before giving up.

On fast x86_64 hosts the condition is satisfied in the first few polls, so the bug is hidden. On slower hardware the loop easily exceeds CI timeouts.

Cap the sleep at 0.2s. This preserves the original intent of polling quickly at first and then backing off, while bounding the total wait to about 50 seconds.

### DIFF
--- a/t/slabs-mover.t
+++ b/t/slabs-mover.t
@@ -40,6 +40,7 @@ sub wait_for_stat {
         last if ($stats_a->{$stat} == $amt);
         sleep $to_sleep;
         $to_sleep += $cnt / 100;
+        $to_sleep = 0.2 if $to_sleep > 0.2;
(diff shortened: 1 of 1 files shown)


dormando:
Thanks! I _think_ this is okay here. The self accumulator was clearly the wrong idea there. The systems I've been de-tuning these tests for are like weird old mips dudes that take 30+ minutes to run the tests. Capping it is the right idea and 50s might be okay but I want to bump that to 0.2s just for extra buffer?

dormando:
released in 1.6.43

PR #1307, opened 2026-07-31
core: refactor do_item_get

YA2IR:
Felt like `do_item_get` was too central a function to contain its legacy residue, which makes it worthy of a gentle cleanup:

- There's a redundant early null check dedicated to `refcount_incr`. The first commit merges it into the block below.
- `was_found` is named and initialized like a boolean (initially assigned to 0, then 1 when it's found), but the reader is then confused by it being assigned to 2 (found but flushed) and 3 (found but expired). The second commit replaces it with a standard enum similar to `delta_result_type`. The `was_found` naming is now gone.
- The flushing and expiry branches ran identical unlink/delete/remove sequences, each followed by its own stats mutex block, which makes the function look more stuffed and complex than it actually is. The third commit unifies them.

No functional changes


dormando:
Hi,

I do agree that code is gross (and it is rewritten in something I'm working on). I don't _typically_ make pure cleanup changes like this in hot paths since I have to try really hard to ensure they are not functional changes, if that makes sense?

I'll give it some consideration though. Sorry for the long delay.

PR #1310, opened 2026-08-11
stats: validate append_stat format arguments

yqtian-se:
## Summary

append_stat is a printf-style wrapper that forwards its format string and variadic arguments to vsnprintf. Declare it as format(printf, 4, 5), then fix the format mismatches exposed by that contract:

- report lru_crawler_starts as a 64-bit value;
- pass uint64_t values to %llu as unsigned long long;
- print sizeof(client_flags_t) with %zu;
- use %s for the sizes_status string.

The compiler-detection change that enables __gcc_attribute__ under GCC and Clang has been split into #1315 at the maintainer request.

## Impact

These values are exposed through the server stats command. In particular, lru_crawler_starts is stored as uint64_t but was formatted with %u, so a sufficiently large counter could be reported incorrectly. The annotation also lets compatible compilers detect future format mismatches at build time once #1315 is applied.

## Testing

Rebased onto the latest next branch and built with GCC using -Wall -Wformat=2.

make test completed successfully:

Files=114, Tests=180089
Result: PASS

### DIFF
--- a/items.c
+++ b/items.c
@@ -964,7 +964,7 @@ void item_stats_sizes(ADD_STAT add_stats, void *c) {
             }
         }
     } else {
-        APPEND_STAT("sizes_status", "disabled", "");
+        APPEND_STAT("sizes_status", "%s", "disabled");
     }
 
     add_stats(NULL, 0, NULL, 0, c);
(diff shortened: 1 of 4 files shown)


dormando:
Shoot, sorry for the really long delay. This looks useful. I'll at least take the main fixes and evaluate the rest later.

yqtian-se:
Thanks! I’d be happy to split the two changes into separate pull requests if that would make review easier. Please let me know which one you prefer!

dormando:
Argh sorry! Yes, please split them up.

yqtian-se:
Done — I kept the append_stat annotation and format fixes in this PR and split the GCC/Clang compiler-detection change into #1315. Both branches are rebased onto the latest next branch, and the full test suite passes independently on each.

PR #1311, opened 2026-08-14
vendor/mcmc: SIMD-accelerate the meta tokenizer for long tokens

Samith2509:
### What

The meta-command tokenizer (`_mcmc_tokenize_meta`) scans a command byte by byte
to find spaces. This change adds a SIMD delimiter scan that jumps over a token
body 16 bytes (SSE2) or 32 bytes (AVX2) at a time. It is length-gated: only
lines longer than 64 bytes take the SIMD path; short commands keep the original
tight per-byte loop, isolated in its own function so its codegen is unchanged.

### Why

Commands carrying a long token (e.g. a large base64 key) spend most of their
time skipping over that token one byte at a time. SIMD skips it in big strides.
Short commands, the common case, are deliberately left on the original loop.

### Correctness

The new tokenizer was checked against a verbatim copy of the original on
2,000,000 random inputs plus targeted edge cases (leading/trailing/multiple
spaces, lines ending in \n with no \r, single-token lines, and lengths right
around the 64-byte crossover). It matched the original exactly every time:

    iters=2000000 mismatches=0
    EDGE TESTS: total=32 fails=0

### Performance

Old vs new tokenizer, same input, ns per call (representative of several runs):

    LONG  (one ~300-byte token):  old ~100 ns  ->  new ~20 ns   (~5x faster)
    SHORT (typical small command): ~18 ns, within ~5% either way (0.99-1.07x)

The short-command result is within code-layout noise for an ~18 ns operation,
measured in a non-inlined test build (the function is `static` and inlined in
the real build). I did not find a reliable short-command regression.

### Portability

Single portable binary. The AVX2 path is compiled with a per-function target
attribute and selected at load time via a CPUID check, so it never runs on a
CPU without AVX2. The fallback is SSE2, always present on x86-64. Non-x86 builds
use `memchr` (SIMD-optimized in glibc). This mirrors the runtime dispatch
memcached already uses for CRC32C.

### Build / tests

`make` builds clean (no new warnings) and `make test` passes.

### Note

`vendor/mcmc/` is vendored from the mcmc project (https://github.com/dormando/mcmc).
Happy to send this to that repo instead if you'd prefer the change land there first.


dormando:
Hi,

The upstream repo is the correct place to file a PR against this vendored file.

I'll add some caveats for you to consider though:

- This is probably the most sensitive loop in the whole codebase, so I need to understand it.
- Meaning I might only be able to learn from your commit until I have time to rewrite it myself from scratch to make sure I understand it (since I have to support it)
- It's not clear if you (if you're a person) understand it since I'm pretty sure this is AI code/post, so I'm going to be careful with my time in giving feedback.
- While testing claims "5x faster", that's a microbenchmark of the function. For consideration of inclusion in the main codebase you need to benchmark the daemon itself before/after. This is how I weigh the value of complexity vs performance since most changes are dwarfed by syscall overhead.

PR #1316, opened 2026-09-16
logger: calculate minimal NEED_ALIGN padding

yqtian-se:
The `NEED_ALIGN` logger path intends to align the next entry to eight bytes.
The current expression is:

```c
8 - (sizeof(logentry) + e->size % 8)
```

Because `%` binds before `+`, only `e->size` is reduced modulo eight. On the
tested 64-bit build, `sizeof(logentry)` is 40, so the unsigned result is
truncated to the `uint8_t` `pad` field as 217-224 bytes instead of the
intended 0-7 bytes.

Producer and consumer remain synchronized because both use the stored
padding, so this does not immediately corrupt the log stream. It does,
however, make each entry consume 216 or 224 extra bytes from the default
64 KiB per-worker logger buffer. For example, an item-get event with a
three-byte key has a 15-byte payload. The current code pushes 272 bytes for
that event; the
minimal aligned size is 56 bytes. Ignoring ring segmentation, this reduces the
buffer capacity from about 1,170 such entries to about 240. Once the buffer no
longer has enough space, the event is not retained; callers of `LOGGER_LOG`
discard the return value from `logger_log`.

This patch computes the remainder of the complete entry size and adds padding
only when the remainder is nonzero. `e->pad` is initialized to zero before
this block, so an already aligned entry retains zero padding.

| Payload-size residue | Current padding | Patched padding |
|---:|---:|---:|
| 0 | 224 | 0 |
| 1 | 223 | 7 |
| 2 | 222 | 6 |
| 3 | 221 | 5 |
| 4 | 220 | 4 |
| 5 | 219 | 3 |
| 6 | 218 | 2 |
| 7 | 217 | 1 |

Validation:

- a focused probe over all eight payload residues produces padding
  `0,7,6,5,4,3,2,1` after the change;
- every resulting next-entry offset is eight-byte aligned;
- after a normal configuration, rebuilding with `CPPFLAGS=-DNEED_ALIGN` and
  running the project suite passes 114 files and 180,089 tests; and
- `t/watcher.t` passes.


dormando:
This is a nice find, thanks!

